PR #24, opened 2026-09-02
Close the board-API threat-model findings: loopback boundary, landing-module gate, CI-log credential scope, coder env secret-scrub

avri-schneider:
# Close the board-API threat-model findings: loopback boundary, landing-module gate, CI-log credential scope, coder env secret-scrub

## What this is

A security pass over the orchestrator, driven by a threat model of the repo, that closes **10 confirmed findings** in one change. Most share a single root cause on the board API; the rest are three independent issues in the landing, CI-log, and coding-backend paths.

## How these were found — the method

This was not a scattershot bug hunt. The findings came out of a **threat model generated with Mipiti**, used as a *search plan* rather than as a document to file away:

1. **Model the system.** Mipiti generated a threat model of the repository — components, trust boundaries, attackers, and the security **controls** each threat implies — from the actual code and docs, not a one-line description.
2. **Correct the model against reality.** The first-pass model carries assumptions that don't match the system; those were fixed so every control maps to a real component and trust boundary. A control against a mis-modeled boundary produces noise, not findings.
3. **Turn each *unimplemented* control into a hypothesis.** Mipiti flags which controls are present and which are missing. Every **missing** control is a testable claim: *"if this control is absent, here is the attack it exists to stop."* That converts an abstract gap ("no origin validation on writes") into a concrete probe ("a cross-origin page can POST this route").
4. **Verify in source, then confirm live where it's decisive.** Each hypothesis was checked against the code, and the ones where a live check settled the question were exercised against a real instance — a real browser from a non-loopback origin, ASGI probes, a raw WebSocket client — all against a throwaway database, with no real repo pushed and the operator's `~/.no_human` untouched.
5. **Keep the distinction honest.** A **confirmed vulnerability** (reachable + verified) is separated throughout from **missing hardening** (an absent control that isn't directly exploitable, or is absent by design). Only the former drove code changes here.

The value of the model was the *coverage*: it enumerated the controls a system like this should have, so the review was guided by "what's missing" rather than by whatever the reviewer happened to think to look for.

## Root cause (one decision behind most findings)

The board API on `127.0.0.1:8420` is unauthenticated, serves `CORSMiddleware(allow_origins=["*"])`, validates no `Host` header, and had its origin check (`_require_local_origin`) on only **11 of 50** mutating routes — **39 mutating routes had no origin guard at all.** Most of the CORS/CSRF findings below are instances of that one gap, and a single boundary closes them together.

## Findings and how each is closed

| ID | Severity | Finding | Fix in this PR |
|----|----------|---------|----------------|
| F1 | Critical | CSRF → coder-prompt rule injection → likely RCE. A cross-origin page POSTs a rule with `confirmed=True`; it lands in the confirmed-injection set fed into every future coder session (which runs with unrestricted Bash), skipping the human confirm gate. | Loopback boundary on every non-GET route |
| F2 | High | CSRF → PR merge under the operator identity (`approve`/`approve-landed`/`finish-review`/`shipped`). | Loopback boundary |
| F3 | Med–High | Cross-origin WebSocket read of all task data — `ws_board` had no Origin check, and browsers don't apply CORS to WebSockets. | Origin + Host check on the `/ws` handshake |
| F4 | Medium | Cross-origin read of task list/details via `CORS *`. | `allow_origins` tightened off `*` |
| F6 | Medium | Cross-origin LLM credential/token burn — an unguarded route builds a backend and calls the model on the operator's credential, no rate limit. | Loopback boundary |
| F7 | Medium | No `Host`-header validation → DNS rebinding, defeating the Origin defense. | Host allowlist `{127.0.0.1, localhost, ::1}` on writes |
| F8 | Medium | Cross-origin config tamper — repoint a project's `repo_paths`/`primary_repo`. | Loopback boundary |
| F5 | High | SSO credentials sent over `verify=False` to a **forge-supplied** URL (any failing check's `targetUrl`), for any host, not just Jenkins. | `default_ci_log_excerpt` scopes the SSO Basic-auth credentials **and the request itself** to the configured `ci_gate.jenkins_controller`; any other host gets no request at all; cleartext `http` is refused; TLS is always verified — against `ci_gate.jenkins_ca_bundle` (new PEM path) when set, else the system trust store. |
| F9 | Medium | Gate check absent from the landing module itself — a marked agent driving `approve_merge` in-process lands with no act-level check. | `land_task` refuses in-process when the caller carries the agent-session mark, at the one site the product performs a real merge; a marked caller returns a failed `LandResult` (the function's never-raise contract); operator/server callers are unmarked and pass. |
| F10 | High | Coder subprocess inherits the launcher's full environment (credential-exfil amplifier). Startup scrubs only Anthropic-billing vars; nothing scrubs `GITHUB_TOKEN`/cloud keys/`SSH_AUTH_SOCK`/the OAuth token — turning F1 from "run commands" into "steal all credentials." | `ClaudeBackend._options` blanks every secret-shaped environment variable that is not Anthropic/Claude auth or the session mark before it reaches the coder subprocess (the Agent SDK inherits the whole parent environment), so a prompt injection in that child cannot read those credentials out of its env. Git authenticates through the `gh` credential helper + `GIT_ASKPASS`, not these variables, so the coder's own work is unaffected. |

The boundary fix (loopback Origin + Host on non-GET routes, plus the `/ws` Origin check) closes F1, F2, F3, F6, F7, F8 at once; tightening CORS closes F4. F5, F9, and F10 are independent and fixed at their own sites.

## What this PR deliberately does **not** change

Kept out of scope, and stated so the review isn't misread as covering them:

- **A per-install bearer token** for the whole API is the real fix for the missing-auth *root cause*; this PR raises the boundary but does not add authentication.
- **Board XSS** is effectively mitigated by the framework already (React auto-escaping; `react-markdown` without `rehype-raw`; the linkifier matches only `https?://`) — no live path found.
- **Agent sandbox / egress allowlist / fs denylist** are absent **by design** — `docs/security.md` §7 assigns that control to the host (a container, a network policy, an egress proxy).
- **Supply-chain signing / provenance / hash-pinned installs, pip-audit gate, push-hook secret scan** are real hardening gaps, but each needs a separate compromise to exploit — hardening, not live vulnerabilities.

## Tests

- `tests/test_local_boundary_guard.py` — the loopback Origin/Host boundary on writes, the `/ws` Origin+Host check (incl. the `TestClient` WebSocket `Host` behavior it has to work around), and CORS scoping.
- `tests/test_gate_at_the_act.py` — the landing-module act-level mark refusal (marked → failed `LandResult`; unmarked reaches the ordinary preconditions), and the coder-subprocess env scrub (secret-shaped vars blanked, Anthropic auth + mark + operational vars intact).
- `tests/test_pr_watcher.py` — the CI-log fetch sends nothing to a foreign or cleartext host and never emits an `Authorization` header off the configured controller; on a match it attaches auth and verifies TLS per `jenkins_ca_bundle`, never `verify=False`.

## Disclosure

Reported through the project's own SECURITY channel; this PR is the coordinated fix. Threat model and per-finding confirmation notes available on request.


eyalgolan:
Thanks for this — the findings are real and the fixes are the right shape. I reviewed the branch against current `main` (merged, full suite, live probes of every control) and pushed maintainer commits on top of yours rather than send it back and forth. Summary of what they change and why:

- **Rebased-by-merge onto current `main`.** Three test files added after your base still addressed the app as `http://test`; they get the same loopback normalisation you applied to the other 39.
- **Structural budget.** `tests/test_structural_budget.py` freezes `api/app.py` and `config.py` line counts and only ratchets down; your additions grew them. The boundary now lives in `api/local_boundary.py` and the env scrub in `agent/child_env.py`, behaviour unchanged, plus `require_local_origin` moved with them.
- **F10 for the Codex backend too.** `CodexBackend._child_env` builds a full copy of `os.environ` and was not scrubbed, so a Codex coder still inherited every ambient secret. `drop_foreign_secrets` handles the full-copy shape; the additive Claude path is your function with a keep-list parameter.
- **The configured `server.host` is an allowed Host.** The `nh` CLI addresses the board by `server.host` (`cli/shell.py:base_url_from_config`), so a `0.0.0.0` or LAN-name bind was turning the product's own CLI channel into a 400. Loopback plus the configured host is accepted; a rebinding page cannot make its domain equal it. Both checks now run in one middleware dispatch.
- **CI-log scoping by controller prefix.** `jenkins_controller` is a URL with a path; comparing hostnames alone still sent SSO credentials to any port or path on that host. It is now an https scheme+host+port+path prefix, a link with a dot segment or an encoded separator (`/..`, `%2e%2e`, `..%2f`) gets no request at all because httpx normalises those after the compare, `load_config` runs off the event loop, and an empty or non-https controller is named in a warning when SSO credentials exist.
- **Env scrub name rules.** Added credential-file pointers, `*_KEY`, `*WEBHOOK*`, the named connection URLs, plus a value rule: any variable whose value is `scheme://user:password@...` is a secret under whatever name. Exempted `AWS_REGION`, `AWS_DEFAULT_REGION`, `AWS_PAGER`, `AWS_CA_BUNDLE`, `AWS_ENDPOINT_URL*`, `TOKENIZERS_PARALLELISM`. Every `*_PROXY` variable is kept even with a password in it, because the child is the process that reaches the model through it. Docs now state exactly what is and is not removed, including that `HOME` and the keyring remain reachable.
- **`tests/test_coder_cache_burn.py`** asserted the child env exactly, which after the scrub passes only in a shell with no secret-shaped variable; it now states the exact env in any shell.
- **Docs, CHANGELOG, `RELEASE_MANIFEST.txt`** (CI checks the manifest with `--strict`; rows are re-pinned in each commit).

One thing only you can do: the `CLA ledger` CI job needs `contributors/avri-schneider.md` (template in `contributors/README.md`) in a commit authored by you. It is the signature, so I have not created it.

If any of the maintainer commits changes something you disagree with, say so here and I will adjust.

eyalgolan:
@avri-schneider To get the `CLA ledger` check green, please add your CLA ledger entry in a commit of your own. It is one file, once, and it is the signature itself, so I cannot add it for you.

1. Read [`CLA.md`](https://github.com/no-human-ai/no_human/blob/main/CLA.md), version 1.0. Agreeing to it is the point of the file.

2. Pull first, since maintainer commits were pushed to your branch:

```
git checkout fix/loopback-origin-host-guard
git pull
```

3. Create `contributors/avri-schneider.md` with exactly this content (fill in your name if you want it recorded; the handle is what CI matches on):

```markdown
# avri-schneider

I have read CLA.md version 1.0 and I agree to it.

- GitHub: @avri-schneider
- Name: Avri Schneider
- Date: 2026-09-03
- Signing as: myself
```

4. Commit and push it with the same author email your other commits use (it is already linked to your GitHub account, which is how the ledger job resolves your login):

```
git add contributors/avri-schneider.md
git commit -m "Agree to CLA.md version 1.0"
git push
```
Thanks!

eyalgolan (approved):

PR #26, opened 2026-09-03
Lock three security behaviors from the fix pass with regression tests

avri-schneider:
Follow-up to #24. Three adversarial tests pinning behaviors the merged code already has but did not directly test:

- **`test_cross_origin_gets_no_allow_credentials_grant`** (tests/test_local_boundary_guard.py) — `allow_credentials` is never enabled, so no page (cross- or same-origin) is granted `Access-Control-Allow-Credentials`, and an adversarial foreign-origin preflight is not granted credentials either.
- **`test_gh_token_and_dotenv_exported_credentials_are_scrubbed`** (tests/test_child_env.py) — `GH_TOKEN` and the `~/.no_human/.env`-exported integration credentials (an SSO password, a Jenkins/JIRA API token) are foreign secrets scrubbed from the coder child env in both the additive (Claude/SDK) and full-copy (Codex) shapes.
- **`web/src/markdownSafety.test.mjs`** — the board renders untrusted tracker/PR/CI text through react-markdown; this renders the REAL dependency and asserts the XSS-relevant defaults it relies on (raw HTML escaped — no rehype-raw; output-encoding; `javascript:`/`data:` link schemes neutralized; control characters inert; a positive control that a real https link still renders), plus a source guard that `Markdown.jsx` enables no `rehype-raw` and no `dangerouslySetInnerHTML`. It fails if a react-markdown/remark-gfm bump regresses that safety.

No source changes; tests + manifest re-pin only. All pass locally.


eyalgolan:
Thanks — locking these behaviours is the right follow-up, and the intent of all three tests is exactly what the code needs pinned. An independent review found that two of them did not yet lock what they describe, so I pushed one maintainer commit on top of yours rather than round-trip it:

- **The markdown suite rendered a bare `ReactMarkdown`, not the board's `Markdown.jsx`.** Adding `urlTransform={(u) => u}` to the real component, or a small local rehype plugin that revives raw HTML (no `rehype-raw` string, no `dangerouslySetInnerHTML`), left all eight tests green while the board rendered a live `data:text/html` link and a live `<iframe>`. The suite now compiles `Markdown.jsx` at test time with vite's own esbuild and renders it, holds every `<ReactMarkdown>` in the file to a props allowlist read by a brace-aware scan of the opening tag, and walks `web/` asserting `Markdown.jsx` is the only react-markdown consumer. Disabling the sanitiser now fails four tests; each of the regressions above fails the suite.
- **The control-character test could not fail**: U+2028/U+2029 break the CommonMark link destination, so the link never parsed and the href it matched came from GFM autolinking the truncated prefix. It now percent-encodes the controls in the destination, asserts exactly one parsed link, and checks the href by char code. A `data:` image test was added for the CSP's `img-src data:` allowance, with a payload that parses.
- **The env test's Codex half** passed the Claude keep-list to the full-copy scrub and asserted the Claude credential survives, the inverse of what `codex_backend` does; it now uses the default Codex keep-list and asserts `CLAUDE_CODE_OAUTH_TOKEN` is dropped while `OPENAI_API_KEY` survives.

The PR description still describes the earlier design; feel free to refresh it, or I will at landing. If you disagree with any of the changes, say so here.

eyalgolan (approved):

PR #27, opened 2026-09-03
Serve the board with a strict Content-Security-Policy

avri-schneider:
Serves the board HTML document with a strict Content-Security-Policy — defense-in-depth for the board UI. Board XSS is *same-origin*, so it would otherwise pass every Origin/Host check and drive the whole loopback API (approve merges, plant a coder rule); this closes the script-execution half even if markup ever slipped past react-markdown's escaping (locked by #26's `markdownSafety.test.mjs`).

Applied in the **`local_boundary` middleware** (not `app.py`, which is under a size ratchet) to `text/html` responses only — the board document. JSON/API responses are unaffected.

- `default-src 'self'; script-src 'self'` — the built `index.html` loads a single same-origin module and **no inline script**, so an injected `<script>` or inline handler cannot execute.
- `object-src 'none'`, `base-uri 'self'`, `frame-ancestors 'none'`, `form-action 'self'` — close plugin, base-tag, clickjacking and form-redirect vectors.
- Inline **style** is allowed (`style-src 'self' 'unsafe-inline'`) for React style props / bundler CSS — not a script-execution vector.

**Trusted Types is sent report-only, not enforced.** The board's own code uses no guarded DOM sink, but bundled dependencies (React/markdown internals) call `innerHTML`, so enforcing `require-trusted-types-for 'script'` could break rendering until each is confirmed TrustedHTML-safe in a real browser.

`tests/test_board_csp.py` pins the policy strict and applied to HTML responses only.


eyalgolan:
Thanks — the reasoning is right (board XSS is same-origin, so a CSP is the layer that stops script execution behind react-markdown's escaping). Two facts changed the shape of this one, so I pushed a maintainer commit on top rather than round-trip:

- **The board already serves a strict CSP.** `api/app.py` defines `_CSP`, computes it once per app start in `_build_csp` (opted-in telemetry adds exactly the PostHog hosts), and applies it on every response in the `_csp_header` middleware. That middleware runs inside `local_boundary`, so the `setdefault` in this branch found the header already set and never applied — measured on a real GET of the built board, the header equals the existing policy. The test passed because it built a bare FastAPI without `app.py`'s middleware. Had the branch's policy applied, its `connect-src 'self'` would also have dropped the explicit `ws:`/`wss:` the board's WebSocket relies on and ignored the telemetry hosts.
- **What is genuinely new is `form-action 'self'`**, which the existing policy lacked. It now sits in `_CSP` (the two exact-string pins in the telemetry tests updated), and `tests/test_board_csp.py` runs against the real app: the board document and a JSON response carry exactly the policy the app computed for its config, with `form-action` present; removing the directive turns both tests red.

The Trusted Types report-only header is not carried over: with no `report-to` endpoint it reports nowhere, and enforcing it is, as you noted, not safe for the bundled dependencies. The PR description now describes a superseded design; feel free to refresh it, or I will at landing.

eyalgolan (approved):

PR #30, opened 2026-09-03
MCP bridge resolves its base URL from server.host and server.port

### LINKED ISSUE #8: MCP bridge hard-codes 127.0.0.1:8420 instead of honouring `server.port`
\`src/no_human/intake/mcp_bridge.py\` line 29 fixes the API base URL:

\`\`\`python
BASE_URL = "http://127.0.0.1:8420"
\`\`\`

That constant is the only thing \`_client()\` and \`_ensure_api_reachable()\` use; the module never reads config. The server side does: \`nh start\` binds \`server.port\` from \`~/.no_human/config.yaml\` (\`DEFAULT_CONFIG['server'] == {'host': '127.0.0.1', 'port': 8420}\`, documented in \`docs/configuration.md\`). So an operator who moves the board to another port gets:

\`\`\`
$ nh mcp-serve
no_human API unreachable at http://127.0.0.1:8420
\`\`\`

with no flag or env var to point it elsewhere (\`nh mcp-serve --help\` has only \`--help\`).

**Fix:** resolve the base URL from \`server.host\` / \`server.port\` when the bridge starts, keeping the current literal as the fallback, plus a test that a non-default port is honoured. ~15 lines.

Sreekant13:
## What this changes

Fixes #8. `main()` resolves the bridge's base URL from `server.host` and `server.port` before the reachability check, so an operator who moves the board is no longer left dialling 8420.

`BASE_URL` stays as the literal fallback rather than being replaced. It is what `_client()` and `_ensure_api_reachable()` read, three existing tests name it, and keeping it means a default install and an unreadable config both behave exactly as before.

The two egress entries for this module are reworded the way `cli/pool_probe.py` already words the same situation, since that module resolves host and port from config too.

## Evidence

```
$ uv run pytest -q tests/test_mcp_bridge.py
.........                                                                [100%]
9 passed in 1.19s

$ git stash push -- src/no_human/intake/mcp_bridge.py
$ uv run pytest -q tests/test_mcp_bridge.py
FAILED tests/test_mcp_bridge.py::test_base_url_follows_server_port_from_config
FAILED tests/test_mcp_bridge.py::test_base_url_falls_back_when_config_unreadable
2 failed, 7 passed in 1.47s

$ uv run pytest -q tests/test_mcp_bridge.py tests/test_egress_disclosure.py tests/test_egress_allowlist.py tests/test_mcp_dependency_bound.py tests/test_mcp_registry_listing.py
...............................................                          [100%]
47 passed in 72.37s (0:01:12)

OK: 1478 file(s) match RELEASE_MANIFEST.txt
```

Test count went 7 to 9 in `tests/test_mcp_bridge.py`. Nothing was removed.

Two notes on my environment, so the numbers above are readable. I am on Windows, and I had to run everything under `PYTHONUTF8=1`. Without it, `tests/test_egress_disclosure.py`, `tests/test_egress_allowlist.py` and `tests/test_cla_nudge.py` fail on a `cp1252` decode error. I checked that against a clean checkout of `main` before touching anything: the same tests fail there in the same way, so it is my platform and not this change. `tests/test_cla_nudge.py` also shells out to `scripts/cla_nudge.sh`, which does not run here at all.

## Checklist

- [x] First PR only: I have read `CLA.md` and added my `contributors/sreekant13.md` entry to this PR.

`RELEASE_MANIFEST.txt` is repinned in this commit, and the diff there is five rows, one per changed file.

One thing worth flagging for anyone else developing on Windows: `scripts/check_release_manifest.py --write` writes the manifest in text mode, so on Windows it lands with CRLF and every one of the 1478 rows shows as changed. Combined with a `core.autocrlf=true` checkout, my first attempt produced a manifest that would have broken CI for everyone. I set the clone to LF and normalised the written file, which is why the diff here is five rows. Happy to open a separate issue for that if it is worth guarding.

Disclosure: I used AI assistance while investigating this and preparing the patch. I have reviewed every changed line and ran every command above myself.


src/no_human/intake/mcp_bridge.py
```diff
@@ -36,6 +37,22 @@
...
 
 
+def _resolve_base_url() -> str:
+    """Base URL of the local API, from ``server.host`` and ``server.port``.
+
+    Falls back to :data:`BASE_URL` when the config is missing or unreadable, so
+    a default install behaves exactly as before.
+    """
+    try:
+        from ..config import load_config
```
eyalgolan: Imports should always be at the top of the file. See [PEP8's imports section](https://peps.python.org/pep-0008/#imports).

One thing to do at the same time, or your two new tests go red: they currently patch `no_human.config.load_config`, which stops reaching this module once the name is bound at import time. Patch `mcp_bridge.load_config` instead (`monkeypatch.setattr(mcp_bridge, "load_config", ...)`). I tried the move on your branch without that change and `test_base_url_follows_server_port_from_config` failed with `'http://127.0.0.1:8420' == 'http://127.0.0.1:9999'`.

src/no_human/intake/mcp_bridge.py
```diff
@@ -36,6 +37,22 @@
...
+        server = load_config(create_if_missing=False).data.get("server") or {}
+    except Exception:  # noqa: BLE001 — an unreadable config must not stop the bridge
```
eyalgolan: Could we catch more specific exceptions here? What `load_config` can actually raise on a bad file is `OSError`, `yaml.YAMLError`, `ValueError`, and the module's own `AuthError` and `ConfigError` (all in `config.py`); a malformed shape below that raises `AttributeError`/`TypeError`. Name those, and log which one it was (see point 2 in the review body), so a typo in config.yaml is visible instead of silently turning into "unreachable at 8420".

src/no_human/intake/mcp_bridge.py
```diff
@@ -36,6 +37,22 @@
...
+        return BASE_URL
+    host = server.get("host") or "127.0.0.1"
+    port = server.get("port") or 8420
```
eyalgolan: We should probably have "127.0.0.1" and 8420 as consts at the top of the file, and derive the fallbacks here from them.

One constraint: keep `BASE_URL = "http://127.0.0.1:8420"` as a literal string. `docs/security.md` cites that line for the token `127.0.0.1:8420`, and `tests/test_readme_claims.py` checks the token is really there. I tried `BASE_URL = f"http://{_DEFAULT_HOST}:{_DEFAULT_PORT}"` on your branch and the citation test failed with "'127.0.0.1:8420' was not found within ±5 lines". So: consts for the resolver's fallbacks, literal for `BASE_URL`.

src/no_human/intake/mcp_bridge.py
```diff
@@ -104,6 +121,8 @@ def _ensure_api_reachable() -> None:
 
 
 def main() -> None:
+    global BASE_URL
```
eyalgolan: As a rule of thumb, never use `global` in python. There's a pretty good explanation about it [here](https://www.reddit.com/r/learnpython/comments/osgvd5/comment/h6oek30/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button).

Concretely here: have `_client()` call a `_base_url()` helper that resolves (and caches) the value, and leave `BASE_URL` as the literal default it falls back to. The only existing test that names `BASE_URL` is the caplog assertion in `test_ensure_api_reachable_refuses_on_connect_error`, and that still holds because the default install resolves to the same string.

tests/test_mcp_bridge.py
```diff
@@ -151,3 +151,25 @@ def handler(request: httpx.Request) -> httpx.Response:
 
     _set_transport(handler)
     mcp_bridge._ensure_api_reachable()  # must not raise
+
+
+def test_base_url_follows_server_port_from_config(monkeypatch):
+    """An operator who moves the board must not be left dialling 8420."""
+    import no_human.config as config_module
```
eyalgolan: Imports should always be at the top of the file. See previous comment about this.

Once `load_config` is imported at the top of `mcp_bridge.py`, this import goes away entirely: patch `mcp_bridge.load_config` here instead of `config_module.load_config` (see my comment on `mcp_bridge.py` line 47 for why).

tests/test_mcp_bridge.py
```diff
@@ -151,3 +151,25 @@ def handler(request: httpx.Request) -> httpx.Response:
...
+    class _Config:
+        data = {"server": {"host": "127.0.0.1", "port": 9999}}
+
+    monkeypatch.setattr(config_module, "load_config", lambda **kw: _Config())
+    assert mcp_bridge._resolve_base_url() == "http://127.0.0.1:9999"
+
+
+def test_base_url_falls_back_when_config_unreadable(monkeypatch):
+    """A missing or broken config leaves a default install working as before."""
+    import no_human.config as config_module
```
eyalgolan: Imports should always be at the top of the file. See previous comment about this.

Once `load_config` is imported at the top of `mcp_bridge.py`, this import goes away entirely: patch `mcp_bridge.load_config` here instead of `config_module.load_config` (see my comment on `mcp_bridge.py` line 47 for why).

eyalgolan (changes requested):
@Sreekant13 Thanks for this! It's super exciting to see your first PR for no_human.
The fix does what #8 asked, both new tests fail on the unfixed code (checked, which is exactly what we want), the manifest repin is correct (OK: 1478 file(s) match), and the CLA ledger entry is in the right place.

Please see my inline comments below; they are required, not optional, and two of them carry a trap you would otherwise hit (moving the import breaks your own tests unless you also change what they patch; an f-string `BASE_URL` breaks the citation checker).

In addition, three things before it can land.
The first is a CI failure you could not have seen, because first-time contributor workflows need maintainer approval to run.

1. CI is red on tests/test_readme_claims.py.

Adding a line to the module docstring moved `BASE_URL` from line 29 to line 30, and `docs/security.md` cites `intake/mcp_bridge.py:29` for it. The citation checker fails on the exact-anchor test:

```
FAILED tests/test_readme_claims.py::test_every_line_citation_currently_resolves_exactly
  security.md citation `intake/mcp_bridge.py:29` is 'drifted' (found_line=30, ), not exactly anchored
```

The same drift also fails `tests/test_reanchor_citations.py::test_check_mode_is_clean_on_this_tree`; one re-anchor fixes both.

The repo has a tool for this. It edits two lines (the citation in `docs/security.md` and the matching `CITATION_TABLE` row in the test), after which the manifest needs repinning again:

```
uv run pytest -q tests/test_readme_claims.py tests/test_reanchor_citations.py
```

I ran that sequence on your branch and `tests/test_readme_claims.py` goes to 121 passed, 12 skipped. Run it last, after every other change, because any further line shift in `mcp_bridge.py` above `BASE_URL` moves the anchor again.

2. The silent fallback re-creates the confusion #8 is about.

`_resolve_base_url` swallows every exception and returns the literal with no message. An operator with a YAML typo in `~/.no_human/config.yaml` now gets the same "unreachable at http://127.0.0.1:8420" with no hint that their config was ignored. Please emit a `log.warning(...)` on the fallback path. Logging goes to stderr here, which is safe for a stdio MCP server, while anything on stdout would corrupt the protocol.

Also, `server.get(...)` sits outside the try, so the docstring's promise does not hold for a malformed config. I planted server: 8421 (a scalar instead of a mapping) and the bridge crashes:

```
AttributeError: 'int' object has no attribute 'get'
```

Move the extraction inside the try, or guard with isinstance(server, dict) the way `api/local_boundary.py:allowed_hosts` does. Note that `cli/shell.py:base_url_from_config` already implements this exact resolution. Importing cli from intake would be a layering inversion, so duplicating it is acceptable, but keep the two shapes identical.

3. Two sentences are now stale.

- The nh mcp-serve docstring in cli/commands.py still says "Talks to the existing local HTTP API at 127.0.0.1:8420".
- The "Not egress: loopback" paragraph in docs/security.md still says the bridge talks on 127.0.0.1:8420. It should say server.host:server.port, 127.0.0.1:8420 by default, which is what the reworded egress entries already say.

Optional, not blocking. The two new tests exercise `_resolve_base_url`, not the wiring in main(). A test that monkeypatches `load_config` and `mcp.run`, injects a `MockTransport` recording the request URL, and asserts the reachability GET hits port 9999 would prove the operator-visible behaviour rather than the helper.

On the Windows CRLF note in `check_release_manifest.py --write`: yes, please open that as a separate issue.

Sreekant13:
Thank you for this review. The two traps you flagged both landed exactly as you said they would, and the notes on where each pattern already exists in the repo saved me a lot of guessing. All of it is done in b61f82c.

**Inline comments**

- Imports at the top, in both files. The tests now patch `mcp_bridge.load_config` rather than `no_human.config.load_config`. You were right that they go red otherwise; I reproduced that before fixing it.
- `global` is gone. `_client()` calls a `_base_url()` helper, cached with `functools.lru_cache(maxsize=1)`, and `BASE_URL` stays the literal default it falls back to. I added an autouse fixture that clears the cache around every test, because a resolved value would otherwise leak into whichever test ran next and quietly break the `caplog` assertion in `test_ensure_api_reachable_refuses_on_connect_error`.
- `_DEFAULT_HOST` and `_DEFAULT_PORT` are constants now and the resolver's fallbacks derive from them. `BASE_URL` is still a literal, for the reason you gave.
- Specific exceptions: `OSError`, `yaml.YAMLError`, `ValueError`, `AuthError`, `ConfigError`, plus `AttributeError` and `TypeError` for a shape that parses but is not a mapping. The warning names which one it was.

**1. CI red on test_readme_claims.py**

Fixed, though not the way you expected, and the difference is worth flagging.

Your run saw a one line drift, 29 to 30, comfortably inside the ±5 window. My rework adds imports and constants above `BASE_URL`, so the drift is now 11 lines. `reanchor_citations.py` correctly refuses to guess at that distance and reports it unfixable, which its own docstring says is left for a human. So I re-anchored 29 to 40 by hand, in both `docs/security.md` and the `CITATION_TABLE` row, and `--check` now reports `VERDICT=OK`.

Two smaller things the tool caught along the way, both my fault. My first docstring wording duplicated the `127.0.0.1:8420` token at line 6, and then my explanatory comment above `BASE_URL` duplicated it again, so the checker kept resolving the citation to the wrong line. Both reworded, and the token now appears exactly once in the module.

**2. Silent fallback**

Agreed, and thank you for planting `server: 8421`. That is a real crash and I should have caught it.

Extraction moved behind an `isinstance` check following `api/local_boundary.py:allowed_hosts`. There are now two warning paths: one when `load_config` raises, naming the exception type, and one when `server` parses into something that is not a mapping, naming what it got instead. Both to stderr. The resolution shape matches `cli/shell.py:base_url_from_config`.

**3. Stale sentences**

Both updated, to `server.host`:`server.port` with 127.0.0.1:8420 as the stated default, matching the reworded egress entries.

**Optional test**

Added, since you are right that the previous two only proved the helper. `test_reachability_check_dials_the_configured_port` patches `load_config` and `mcp.run`, injects a `MockTransport` that records the URL, calls `main()`, and asserts the reachability GET hits `http://127.0.0.1:9999/api/tasks`.

**Evidence**

```
$ uv run pytest -q tests/test_mcp_bridge.py tests/test_egress_disclosure.py tests/test_egress_allowlist.py tests/test_mcp_dependency_bound.py tests/test_mcp_registry_listing.py tests/test_readme_claims.py tests/test_reanchor_citations.py
175 passed, 12 skipped in 70.77s (0:01:10)

VERDICT=OK

OK: 1478 file(s) match RELEASE_MANIFEST.txt
```

`tests/test_mcp_bridge.py` goes from 9 to 11. Manifest repinned last, 5 rows, one per file changed in this round.

The `check_release_manifest.py --write` CRLF issue is filed separately as requested.

eyalgolan:
Pushed two small commits onto your branch (maintainer edit), hope that's ok:

- 2b468c7 — the reworded `mcp-serve` docstring went from two lines to three, and `tests/test_structural_budget.py` freezes `cli/commands.py` at 8423 lines (it only ratchets down, so +1 fails). Same sentence, wrapped to two lines. Manifest repinned.
- d01873f — merge of main with the manifest regenerated. Main moved under you and the manifest is a generated file, so it conflicts with every open PR. Nothing you did wrong.

Worth adding `tests/test_structural_budget.py` to your local run whenever a change touches `cli/commands.py`.

Everything else checks out. I also looked at the SDK side: FastMCP hangs a RichHandler on stderr, so the new warning can't leak into the stdio stream. Two nits I'm not going to hold this up for: `host: [1]` in config gives you `http://[1]:9999` (it's your config, and httpx will tell you), and `lru_cache(maxsize=1)` on a no-arg function is just `functools.cache`.

On the merged tree: 217 passed across the nine guard files, manifest OK, `reanchor_citations.py` VERDICT=OK. Approving once CI comes back green. Thanks for sticking with this one.

eyalgolan:
One more on your branch, 7d135c8, and this one's on me: I named `AuthError` and `ConfigError` in the exception list and shouldn't have. Those two aren't "couldn't read the file", they're the product refusing a file it read fine (an `ANTHROPIC_API_KEY` sitting in config.yaml, for instance). Catching them meant the bridge warned and then dialled 8420 while the config said 9001, which is the exact confusion #8 was about, just behind a different door. They propagate now, so `nh mcp-serve` refuses to start like every other entry point does.

While there I added two tests: one goes through the real `load_config` with a real file and a non-default host as well as port (host was never exercised, every stub used 127.0.0.1), the other checks that a config with an API key in it raises out of `_base_url`. Tests 11 to 13. Manifest repinned.

Nothing needed from you. Landing this once the full run comes back.

Sreekant13:
Both your commits pulled, and thank you for them.

`_CONFIG_ERRORS` catching `AuthError` and `ConfigError` was my mistake, and a bad one. You listed those in your review as what `load_config` can raise, and I turned "can raise" into "should swallow" without asking what each one means. They are the product refusing a file it read fine, so my version warned and then dialled the default port while the config named another. That is precisely the bug this PR exists to remove, reachable through a credential guard. Your version is right, and the comment you left above the tuple is the part I should have written myself.

The `test_structural_budget.py` lesson is noted. My three line docstring took `cli/commands.py` to 8424 against a frozen 8423. I have added it to my local run for anything touching that file, and more generally I now content-grep the tests for the path of every file I change rather than picking test files by name. That is what let `test_readme_claims.py` slip past me the first time.

Pushed 437e7fc on top of your 7d135c8:

- Merged `origin/main`. `58e5f6f` landed after your `d01873f`, so `RELEASE_MANIFEST.txt` conflicted again. Resolved the way you did: take main's file, regenerate, repin.
- Took the `lru_cache` nit. A push was needed for the merge anyway, so folding it in costs one CI run instead of two. `cache_clear` still exists so the autouse fixture is unchanged, and the decorator stays one line so the citation does not drift off line 40.
- Left `host: [1]` alone, per your call.

One thing to flag rather than let you wonder. The manifest now carries a row that is not mine, `tests/test_ui_evidence_build_cmd.py`. `58e5f6f` adds that file without pinning it, so a clean checkout of main reports `tests/test_ui_evidence_build_cmd.py: tracked but not listed`. It is warning-level, and main's gate still exits `OK: 1477 file(s) match (1 tracked file(s) unlisted)`, so nothing is broken. Regenerating just picks it up. Happy to drop that row and leave main to repin it if you would prefer the diff stay strictly mine.

```
$ uv run pytest -q tests/test_structural_budget.py tests/test_mcp_bridge.py tests/test_egress_disclosure.py tests/test_egress_allowlist.py tests/test_mcp_dependency_bound.py tests/test_mcp_registry_listing.py tests/test_readme_claims.py tests/test_reanchor_citations.py tests/test_plugin_drift.py tests/test_ui_evidence_build_cmd.py
218 passed, 12 skipped in 91.43s (0:01:31)

VERDICT=OK

OK: 1479 file(s) match RELEASE_MANIFEST.txt
```

Ready for CI whenever you want to approve the run.

eyalgolan (approved):
Approving at 7d135c8. Full gate set is green on the merged tree, and a second reviewer went through the final diff independently. Landing by local squash with you as author, same as the other contributor PRs, so the merge button stays untouched.

Sreekant13:
Thank you for the approval, and for the second pair of eyes on the diff.

One small timing thing worth checking before you squash, since you land these by hand.

Your approval names 7d135c8, but the branch head moved to 437e7fc about two minutes earlier, so the two may have crossed. That commit merges main, because 58e5f6f landed after your d01873f and put RELEASE_MANIFEST.txt back into conflict, and it also takes the lru_cache nit you raised.

So if you squash at 7d135c8 you would get a clean tree, just without the `functools.cache` change you suggested and without the main merge. Nothing breaks either way. I only mention it so you do not end up thinking the nit shipped when it did not.

Happy either way. If you would rather land 7d135c8 as approved, say so and I will open the `functools.cache` change as a separate one line follow up once this is in.

eyalgolan:
Good flag, and yes, they crossed. I'm landing 437e7fc, not 7d135c8. The squash I've built has your merge and the `functools.cache` change in it (its tree is byte-identical to 437e7fc's), so the nit ships with this PR and there's nothing to follow up.

Keep the `tests/test_ui_evidence_build_cmd.py` row. You're right that 58e5f6f left it unpinned, and a regenerated manifest is the honest state of the tree, not a diff to trim back. That one's mine to have caught, thanks for not quietly dropping it.

Full run on the exact landing tree is at 72% and clean so far. Once it finishes this goes in.

eyalgolan:
Landed on main as f24d15c5a, squashed from 437e7fc with you as author. GitHub shows the PR as closed rather than merged because I land by hand instead of the button, but `git log main --author=Sreekant13` will show it. Full run on the landing tree was clean apart from one fixture-provenance test that needs objects only my clone has; that one fails the same way on main and has nothing to do with this change.

Thanks for the work on this, and for flagging the crossed approval. #8 is closed. The #32 PR is welcome whenever you have it.

Sreekant13:
Thanks for landing it, and for the review. It was a genuinely educational one, particularly the point that what `load_config` can raise is not the same question as what the bridge should swallow. I had read your list as a catch tuple and missed that `AuthError` and `ConfigError` are the product refusing a file it read fine.

I will get the #32 PR up shortly, with the `newline="\n"` change to `--write` and nothing else in it.

PR #41, opened 2026-09-04
CLA gate: exempt the repo's own agent account (no-human)

eyalgolan:
Every fleet PR failed only the CLA ledger job: the product's agent account (`no-human`) authors the branch commits, and the gate treated it as a foreign contributor. It is first-party — its work lands identity-normalized to the maintainer — so it is exempted exactly like the maintainer and `[bot]` accounts, in both the ci.yml job and the nudge script, with a regression test.

Also heals RELEASE_MANIFEST.txt: 58e5f6fee175 shipped `tests/test_ui_evidence_build_cmd.py` without its manifest row.


eyalgolan:
Landed on main as 3bfa5aca (supervised; independent review PASS incl. the opener-gated hardening; full suite 11210 green on the exact stacked tree).

PR #43, opened 2026-09-04
Mechanical PR-conflict resolution works on inventory-gated repos (no export_guard.py)

eyalgolan:
**Root cause (live, 3 escalations today):** `resolve_derived_conflict` hard-required `scripts/export_guard.py` (derived_conflict.py, entry gate) — only the private repo has it. On this repo every manifest-only PR conflict failed mechanically with "step 'regenerate'" and escalated a finished task; every landing re-conflicts every open PR, so the queue compounds. Any customer repo using the public tooling hits the same wall.

**Fix:** repos carrying `scripts/check_release_manifest.py` get an inventory backend: `--write` rebuilds the whole manifest from the merged tree, the no-flag run verifies before push; push/update-ref mechanics mirror the export-guard tail. Fail-closed on neither-tool and on classification-eligible conflicts without export_guard.py.

**Evidence:** 4 new end-to-end git-fixture tests run the REAL tool (`4 passed`); ablation: the happy-path test fails against the unfixed module with the exact live escalation shape; neighboring conflict suites `124 passed`.


eyalgolan:
Landed on main as 97a9fd79b69a (supervised; independent review PASS; full suite 11191 green on the exact tree).

PR #53, opened 2026-09-04
Write release manifest with LF line endings on all platforms

### LINKED ISSUE #32: check_release_manifest.py --write produces a CRLF manifest on Windows, rewriting all 1478 rows
Filing this separately as asked on #30.

`scripts/check_release_manifest.py --write` opens the manifest in text mode, so on Windows Python translates every `\n` to `\r\n` on the way out. The manifest is then hashed and compared as bytes, so every row differs from what the repository holds.

The effect is that a Windows contributor who follows the documented workflow produces a diff touching all 1478 rows instead of the handful of files they changed. Committing that would look like a whole tree rewrite in review, and would break the gate for everyone else.

On my first attempt this stacked with a second cause. A checkout made with `core.autocrlf=true`, which is the Git for Windows default, gives every working file CRLF line endings, so the hashes are computed over different bytes than the ones stored. Both had to be fixed to get a five row diff.

What I did locally:

```
git config core.eol lf
git rm --cached -r . && git reset --hard
# then convert the written manifest back to LF
```

The second half is the part the repository could fix. Writing the manifest with an explicit `newline="\n"` would make `--write` produce identical bytes on every platform, and would leave only the `core.autocrlf` half for the contributor to handle.

Two things that might be worth considering alongside it:

- A note in CONTRIBUTING for Windows contributors about `core.autocrlf`, since the failure is silent and looks like the contributor did something wrong.
- A guard in `--write` that refuses, or at least warns loudly, when it is about to rewrite substantially every row, since that is almost always a line ending problem rather than a real change.

I am happy to open a PR for the `newline="\n"` change if you would like it, though it touches shipped tooling so I did not want to bundle it into #30 uninvited.

Environment: Windows 11, Python 3.12, uv. Reproduced on a clean checkout of `main` before any of my changes.

guamiao:
## Problem

On Windows, `scripts/check_release_manifest.py --write` rewrites every row of
`RELEASE_MANIFEST.txt` instead of only the rows for files that actually changed.

## Root cause

`write_manifest()` writes the manifest with
`(root / MANIFEST_NAME).write_text(...)`. `write_text()` opens the file in text
mode, so on Windows Python translates every `\n` in the content to `\r\n` on
the way out. The manifest is then hashed and compared as bytes, so every row
differs from the LF version held in the repository — a Windows contributor
following the documented workflow produces a whole-tree rewrite.

## Fix

Pass `newline="\n"` to `write_text()`, so the manifest is written with LF line
endings on every platform and produces identical bytes regardless of OS.

Fixes #32.

## Verification

On Windows 11, Python 3.12, uv:

$ uv run python -c "import tempfile,os; d=tempfile.mkdtemp(); f1=os.path.join(d,'a.txt'); f2=os.path.join(d,'b.txt'); open(f1,'w').write('x\ny\n'); open(f2,'w',newline='\n').write('x\ny\n'); print('without newline:', open(f1,'rb').read()); print('with newline=\n:', open(f2,'rb').read())"
without newline: b'x\r\ny\r\n'
with newline=\n: b'x\ny\n'

Without `newline`, Windows writes CRLF; with `newline="\n"` it writes LF — the
exact behaviour the fix relies on. The check script itself still runs and
reports correctly after the change.

## CLA

First PR — I have read CLA.md and added `contributors/heyamei690-hash.md` in
this PR.

### DIFF
--- a/contributors/heyamei690-hash.md
+++ b/contributors/heyamei690-hash.md
@@ -1,0 +1,7 @@
+# heyamei690-hash
+
+I have read CLA.md version 1.0 and I agree to it.
+
+- GitHub: @heyamei690-hash
+- Date: 2026-09-04
+- Signing as: myself
--- a/scripts/check_release_manifest.py
+++ b/scripts/check_release_manifest.py
@@ -226,7 +226,7 @@ def write_manifest(root: Path) -> int:
 
     rows = {rel: hash_path(root, rel) for rel in tracked}
     body = "".join(f"{digest}  {rel}\n" for rel, digest in sorted(rows.items()))
-    (root / MANIFEST_NAME).write_text(HEADER + body, encoding="utf-8")
+    (root / MANIFEST_NAME).write_text(HEADER + body, encoding="utf-8", newline="\n")
     print(f"{MANIFEST_NAME}: wrote {len(rows)} row(s)")
     return 0
 


eyalgolan (changes requested):
Hi @heyamei690-hash, thanks for the contribution!
One thing first: #32 was filed by @Sreekant13, who had offered the PR and whom I'd already told to go ahead.

@Sreekant13, your call. If you'd rather take it solo, say so and I'll close this one. If you're happy for @heyamei690-hash to carry it, the review below is the list.

**Good.** `newline="\n"` is the right knob on our Python 3.12 floor, `HEADER` and the body already use bare `\n`, the read side is universal-newline, no CRLF in the branch files, the CLA file matches the template (Name is optional) and its filename matches your handle. It is byte-for-byte the change proposed in #32.

**Blocking.** CI goes red on the manifest gate: the script's hash changed and the new CLA file has no row, but `RELEASE_MANIFEST.txt` was not repinned. On a merge with current main the gate reports:

```
scripts/check_release_manifest.py: content differs from the manifest
```

Merge `main` (the branch forked 11 commits ago), run `uv run python scripts/check_release_manifest.py --write`, commit the two-row diff. Always the last step before pushing.

**Gaps.**
- No test. On Linux CI this fix is invisible. Add a test next to the existing `write_manifest` helper in `tests/test_check_release_manifest.py` that runs `--write` into a tmp repo and asserts `b"\r"` is not in the bytes, then add that file to the Windows job's test list in `.github/workflows/ci.yml` (the explicit `tests/test_windows_*` entries, around line 649). That job is the only place it can bite.
- The verification in the body shows Python's `open()`, not the script. Paste `--write` followed by `git diff --stat` from your Windows checkout, showing only the rows you changed.
- The `core.autocrlf` note for `CONTRIBUTING.md` and the "warn when most rows change" idea from #32 are welcome in the same PR. Not blocking.

Sreekant13:
Hi @heyamei690-hash, and thank you for picking this up. Your diagnosis is right and your one line is exactly the change #32 proposed, so nothing about your reading of the problem is off.

My usual instinct is to let the best version merge whoever wrote it, and I have stood down on someone else's PR before for exactly that reason. Here I think the honest thing to say is that a few of the grounds are still uncovered, and I had already built them before your PR went up: the test that asserts no CR in the written bytes, its entry in the Windows job in `ci.yml` (which is the only lane where that assertion can fail, since on linux it passes for free), the `core.autocrlf` note for `CONTRIBUTING.md`, and the warn-on-a-wide-diff behaviour @eyalgolan and I had settled on in the issue. Plus the manifest repin.

So rather than have you work through all of that in review, I have opened #85 with the whole set. I would rather that than leave you doing several rounds against gates that took me a while to learn.

Please do not read this as your change being unwelcome. It was correct, it was quick, and you spotted a real bug in a repo you had just arrived in. There is other work open that nobody has taken, and if you want a hand getting oriented in this codebase I am happy to help. The manifest and citation gates in particular are not obvious until they bite you once.

@eyalgolan, taking it solo per your offer.

eyalgolan:
> Hi @heyamei690-hash, and thank you for picking this up. Your diagnosis is right and your one line is exactly the change #32 proposed, so nothing about your reading of the problem is off.
> 
> My usual instinct is to let the best version merge whoever wrote it, and I have stood down on someone else's PR before for exactly that reason. Here I think the honest thing to say is that a few of the grounds are still uncovered, and I had already built them before your PR went up: the test that asserts no CR in the written bytes, its entry in the Windows job in `ci.yml` (which is the only lane where that assertion can fail, since on linux it passes for free), the `core.autocrlf` note for `CONTRIBUTING.md`, and the warn-on-a-wide-diff behaviour @eyalgolan and I had settled on in the issue. Plus the manifest repin.
> 
> So rather than have you work through all of that in review, I have opened #85 with the whole set. I would rather that than leave you doing several rounds against gates that took me a while to learn.
> 
> Please do not read this as your change being unwelcome. It was correct, it was quick, and you spotted a real bug in a repo you had just arrived in. There is other work open that nobody has taken, and if you want a hand getting oriented in this codebase I am happy to help. The manifest and citation gates in particular are not obvious until they bite you once.
> 
> @eyalgolan, taking it solo per your offer.

Sure @Sreekant13 , this makes sense. 
@heyamei690-hash thanks for opening this PR. As @Sreekant13 said, your diagnosis is correct, and we would love if you coul tackle on of the other issue!

guamiao:
Thank you for the kind words and for taking the time to explain all of this. I'm glad the fix was on the right track, and I appreciate you building out the full version in #85 — that's a great learning point for me about the manifest/citation gates.
I'd love to contribute to the other open tasks. If there's a good one for someone still getting oriented in the codebase, I'm happy to take a look. Thanks again!

Issue #16, opened 2026-08-20
Port the MCP bridge to the SDK's 2.x API so the `mcp<2` cap can be lifted


eyalgolan:
Follow-up to #15. `src/no_human/intake/mcp_bridge.py` imports `mcp.server.fastmcp`, which the MCP Python SDK removed in 2.0.0, so `pyproject.toml` now caps the dependency at `mcp>=1.28.0,<2`. That cap is correct for a hotfix, but it shouldn't become permanent by neglect — new SDK work happens on 2.x.

**What the 2.x replacement looks like** (verified against `mcp==2.0.0` during the review of the cap):

- `mcp.server.MCPServer` (module `mcp.server.mcpserver`) is the FastMCP equivalent: same `.tool()` decorator, same `.run(transport="stdio")` signature. A rename gets both of our tools registering:
  ```
  tools: [('task_add', ['description','repo_path','title']),
          ('task_status', ['task_id_or_external_id'])]
  ```
- One breaking rename to handle: `Tool.inputSchema` → `Tool.input_schema`.

**Scope**

1. Port `intake/mcp_bridge.py` to `mcp.server.MCPServer`.
2. Raise the floor (`mcp>=2`) and drop the `<2` cap; `tests/test_mcp_dependency_bound.py` documents the rule — the bound and the import must move together, so update `FIRST_SDK_WITHOUT_FASTMCP` / the import assertion in the same change.
3. Re-lock, and re-verify the two lanes that exercise the bridge for real: CI's container handshake (`Dockerfile.mcp` + `packaging/mcp_handshake_probe.py`) and the wheel job's fresh-resolution import.
4. Check the desktop bundles still start (they ship the locked SDK).

Good task for someone who wants a self-contained change with a real end-to-end check attached to it — the handshake probe tells you immediately whether it works.

Sreekant13:
Claiming this one, please.

I probed the 2.x surface myself before commenting, against **2.1.1** rather than the 2.0.0 your note was verified on, since that is what a fresh resolve gets today. Three differences from the issue text, all of them in our favour:

- `from mcp.server import MCPServer` works directly. `MCPServer` is re-exported from `mcp.server`, so the `mcp.server.mcpserver` path is not needed.
- `run()` already defaults to stdio: `run(self, transport: Literal['stdio','sse','streamable-http'] = 'stdio', **kwargs)`. So `mcp.run()` in `main()` needs no change at all.
[...]

Issue #15, opened 2026-08-20
`nh mcp-serve` fails with ModuleNotFoundError on any PyPI install (mcp 2.0.0 dropped mcp.server.fastmcp)


eyalgolan:
**Affects 0.1.1, 0.1.2 and 0.1.3 — anyone who installed from PyPI.**

```
$ uvx no-human mcp-serve
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

Same failure for `nh mcp-serve` after `uv tool install no-human`, and for the Claude Code plugin (`plugins/no-human/.mcp.json` runs `nh mcp-serve`).

**Cause.** `src/no_human/intake/mcp_bridge.py` imports `mcp.server.fastmcp`. The MCP Python SDK removed that path in **2.0.0**, and our requirement was `mcp>=1.28.0` with no upper bound — so a fresh resolve now installs 2.0.0:

```
$ python -c "import importlib.util as u; print(u.find_spec('mcp.server.fastmcp'))"   # with mcp==2.0.0
None
$ python -c "import importlib.util as u; print(u.find_spec('mcp.server.fastmcp'))"   # with mcp==1.29.0
ModuleSpec(name='mcp.server.fastmcp', ...)
```

Nothing caught it because every lane that runs the bridge resolves through `uv.lock`, which pins 1.29.0: CI's container handshake, the desktop bundles, and `uv sync` in a checkout. The one CI step that resolves the way a user does (`uv tool install dist/*.whl`) never touched the bridge.

**Workaround until the fix ships:**

```bash
uvx --with "mcp<2" no-human mcp-serve
# or, for an installed tool:
uv tool install --force --with "mcp<2" no-human
```

The desktop apps are unaffected (they bundle the locked SDK).

**Fix, landing as 0.1.4:** cap the requirement at `mcp>=1.28.0,<2`; add a test that fails if the declared bound ever admits an SDK without the module the bridge imports; and make CI's wheel job import the bridge inside its freshly-resolved tool environment, which is the only CI lane that sees what a user gets.

eyalgolan:
Workaround verified just now against the published 0.1.3 — a real MCP handshake over stdio, not just a start:

```
$ uvx --with "mcp<2" no-human mcp-serve
initialize -> {"name": "no_human-mcp-bridge", "version": "1.29.0"}
tools/list -> ['task_add', 'task_status']
```

(The bridge needs your local no_human running — `nh start` — as it always does.)

[...]

PR #79, opened 2026-09-04
Port intake/mcp_bridge.py to MCP SDK 2.x (MCPServer) and raise mcp>=2

Siddh2024:
## What this changes

Follow-up to #15. Ports `src/no_human/intake/mcp_bridge.py` from the legacy `mcp.server.fastmcp.FastMCP` (removed in MCP SDK 2.0.0) to `mcp.server.mcpserver.MCPServer`, and raises the dependency floor in `pyproject.toml` to `mcp>=2` (dropping the temporary `<2` hotfix cap).

Key changes:
- `src/no_human/intake/mcp_bridge.py`: imports `mcp.server.mcpserver.MCPServer` instead of `FastMCP`.
- `pyproject.toml`: updates dependency to `"mcp>=2"`.
- `uv.lock`: re-locked to `mcp 2.1.1` and `claude-agent-sdk 0.2.152`.
- `tests/test_mcp_bridge.py`: updated schema property access to `tool.input_schema` and tool call return extraction to `result.content[0].text`.
- `tests/test_plugin_drift.py`: updated `_names()` to access `input_schema` (with backward-compatible fallback).
- `tests/test_brain_invariants.py`: updated frozen dependency entry to `"mcp>=2"`.
- `tests/test_mcp_dependency_bound.py`: updated dependency boundary tests to reject `<2` and assert `mcp.server.mcpserver`.

## Evidence

Targeted test suite:
- Command: `pytest tests/test_mcp_bridge.py tests/test_mcp_dependency_bound.py tests/test_plugin_drift.py tests/test_brain_invariants.py -k "test_L1 or mcp" -v`
- Output: `29 passed in 2.68s`

Egress allowlist:
- Command: `pytest tests/test_egress_allowlist.py -v`
- Output: `22 passed in 37.19s`

Handshake probe:
- Command: `python packaging/mcp_handshake_probe.py --command python -m no_human.intake.mcp_bridge`
- Output: initialize ok: no_human-mcp-bridge (protocol 2025-06-18) tools/list ok: ['task_add', 'task_status'] HANDSHAKE OK

## Checklist

- [x] First PR only: I have read `CLA.md` and added my `contributors/<handle>.md` entry to this PR. (Ticking this box is not the record — the file is. CI checks for the file.)


eyalgolan (changes requested):
@Siddh2024 Thanks for taking this on, and for doing it the way #16 asked: the port is right, the handshake probe passes on your branch, and I built a venv from your `uv.lock` and checked that every name we import from `claude_agent_sdk` still exists in 0.2.152. So the core of this is done. What follows is what stands between it and main.

**Three things that block it**

1. The CLA ledger check will fail. The file is `contributors/Siddh2024.md`, and the job lowercases the handle and looks for `contributors/siddh2024.md`. CI runs on Linux, where that's a different file. It also starts with a UTF-8 BOM (`EF BB BF`). `git mv` it to the lowercase name and save it without the BOM.

2. The manifest isn't repinned. `pyproject.toml`, `uv.lock`, the bridge and four test files changed, and the new contributors file has no row, so the File inventory gate reports hash mismatches. Merge `main` first (it moved under you), then `uv run python scripts/check_release_manifest.py --write` and commit the diff. Make it the last step before every push.

3. `mcp>=2` with no upper bound puts back the exact hole #15 was. The cap existed because everything here resolves through `uv.lock` while a `pip install no-human` resolves from PyPI. When the SDK ships 3.0 and moves the module again, we'd ship a broken `nh mcp-serve` a second time and nothing in the tree would notice. Please make it `mcp>=2,<3`, and in `tests/test_mcp_dependency_bound.py` keep a test that the declared bound excludes the next major (the mirror of the prerelease test you removed, with `3.0.0a1`, `3.0.0rc1`, `3.1.0` as candidates). Keep the 2026-08-20 incident paragraph in that module's docstring too; it's the reason the file exists.

**Two more from the issue's scope**

- Item 4, the desktop bundles. They ship the locked SDK, and this PR moves `claude-agent-sdk` from 0.2.121 to 0.2.152 because 0.2.121 pins `mcp<2`. The Desktop, Windows and Linux jobs skip on PRs unless the PR has the `desktop` label. I'm adding it now, so they'll run on your next push. Note in the PR body that they came back green.

- The `getattr(tool, "input_schema", getattr(tool, "inputSchema", None))` fallback in `tests/test_plugin_drift.py` is dead once the floor is 2, and it hides a failure: if `schema` were ever not a dict, `_names()` returns only the tool name and the drift test passes with nothing checked. Use `tool.input_schema` directly, the way `test_mcp_bridge.py` does.

For what it's worth, I looked at the new transitive packages (`httpx2`, `httpcore2`, `truststore`, `opentelemetry-api`, `mcp-types`): all pulled by `mcp 2.1.1` itself, all from `files.pythonhosted.org`, and `httpx2-jsfetch` only installs under emscripten. No objection there.

I'm running the full suite under your lock on my side, since the SDK bump reaches the coder backend. I'll post the result here.

eyalgolan:
Hi Siddh, I want to get this in, so I pushed the review items onto your branch myself rather than wait another day (maintainer edit, one commit on top of yours, nothing of yours rewritten). What it does:

- `mcp>=2,<3`. Your floor is right; the cap is the #15 lesson one major up, and it's back with the reasoning in pyproject.toml. `tests/test_mcp_dependency_bound.py` gets its incident history back plus a test that the bound excludes 3.x and its pre-releases. Both bound tests go red on the mutation they guard against, I checked. uv.lock changes by the specifier line only, the resolution is untouched.
- `contributors/Siddh2024.md` -> `contributors/siddh2024.md`, BOM removed. The ledger job lowercases the handle and CI runs on Linux.
- `tests/test_plugin_drift.py` reads `tool.input_schema` directly. The `inputSchema` fallback is dead on 2.x and would have let the drift test pass with nothing checked.
- A CHANGELOG entry under Unreleased, since this is a runtime-dependency major bump and 0.1.4's entry told people to pin `mcp<2`; and the egress allowlist line that still said FastMCP.
- Manifest repinned under `--strict`.

Your port itself is unchanged and it's good: handshake probe OK, every `claude_agent_sdk` name we import exists in 0.2.152, and the full suite under your lock is clean on my side. I'll approve the CI run so the desktop bundles build on this head, and merge once it's green.

Thanks for the port, it's the right change and it closes #16.

eyalgolan (approved):

PR #85, opened 2026-09-04
Write the release manifest with LF on every platform

### LINKED ISSUE #32: check_release_manifest.py --write produces a CRLF manifest on Windows, rewriting all 1478 rows

Sreekant13:
## What this changes

Fixes #32.

`--write` built the manifest through Python's text layer, so on Windows every row landed with CRLF. The compare side reads back through the same layer and stays quiet, which is why the tool looks healthy while git reports the whole file as changed. Following the documented workflow on Windows produced a diff touching all 1512 rows instead of the four I had actually changed.

`newline="\n"` on that write makes the output byte-identical on every platform.

Alongside it, `--write` now prints one NOTE on stderr when more than a quarter of the existing rows change, naming line endings as the usual cause. A warning and never a refusal, as you asked: a dependency bump or a wide sweep can legitimately touch a large share of the rows, and a `--force` flag would only get typed reflexively. It catches the one case that is never legitimate and is otherwise completely silent.

`CONTRIBUTING.md` gains a short note about `core.autocrlf`, since the failure looks like the contributor did something wrong and they did not.

## Evidence

```
$ uv run python scripts/check_release_manifest.py --write
RELEASE_MANIFEST.txt: wrote 1512 row(s)

$ git diff --stat
 .github/workflows/ci.yml             |  6 +++-
 CONTRIBUTING.md                      | 14 +++++++++
 RELEASE_MANIFEST.txt                 |  8 +++---
 scripts/check_release_manifest.py    | 31 +++++++++++++++++++-
 tests/test_check_release_manifest.py | 56 ++++++++++++++++++++++++++++++++++++

$ uv run pytest -q tests/test_check_release_manifest.py
15 passed, 7 skipped in 11.80s

$ git stash push -- scripts/check_release_manifest.py
$ uv run pytest -q tests/test_check_release_manifest.py
FAILED tests/test_check_release_manifest.py::test_write_produces_lf_bytes_on_every_platform
FAILED tests/test_check_release_manifest.py::test_write_notes_when_it_rewrites_most_of_the_manifest
2 failed, 13 passed, 7 skipped in 10.45s

# That run is on WINDOWS. test_write_produces_lf_bytes_on_every_platform
# passes on mac and linux whatever the script does, which is the whole reason
# it goes in the Windows job. On those platforms only the note test goes red.

$ uv run pytest -q tests/test_ci.py tests/test_ci_network_step_bounds.py tests/test_derived_conflict_inventory.py tests/test_readme_claims.py tests/test_structural_budget.py
285 passed, 19 skipped in 130.62s (0:02:10)

OK: 1512 file(s) match RELEASE_MANIFEST.txt
```

That `git diff --stat` is the whole point of the change: four rows in the manifest, one per file I touched, from a Windows checkout.

The four tests are the LF bytes assertion, byte-for-byte idempotence on a second write, the note firing on a wide diff, and the note staying quiet on a small one. `tests/test_check_release_manifest.py` is added to the Windows job in `ci.yml`, which is the only lane where the line ending assertion can fail. On linux it passes for free.

`tests/test_precommit_manifest_gate.py` fails 10 for me, identically on a clean worktree of `main`. It drives the pre-commit hook through bash, which does not run on this platform, so it is unrelated to this change.

## Checklist

- [x] First PR only: not applicable, my `contributors/sreekant13.md` entry landed with #30.


eyalgolan (changes requested):
Thanks @Sreekant13 , and thanks for how you handled #53 - that was the right call and you were generous about it.

Good PR. This is the version of #32 I wanted: the fix, a test that can actually catch it, the Windows job, the doc note, the warning. Merges clean, inventory is OK on the merge result, CI is green. So close.

There's one thing I need fixed though, and it's a real one.

The warning reads the old manifest with parse_manifest, and parse_manifest bails on anything it can't parse. Think about when people actually run --write: it's after the manifest conflicted on a merge and the file still has <<<<<<< markers in it. You've done this twice yourself on #30. I tried it:

```
$ python scripts/check_release_manifest.py --root <repo> --write     # main
RELEASE_MANIFEST.txt: wrote 3 row(s)

$ python scripts/check_release_manifest.py --root <repo> --write     # your branch
RELEASE_MANIFEST.txt:9: cannot parse '<<<<<<< HEAD' — every row is `<sha256>  <path>`
```

So on main the write clears the conflict, on your branch it dies. That's the recovery path, we can't have the warning break it. Wrap the read of the old rows in a try and if it doesn't parse just skip the note. And add a test for it - manifest with conflict markers, --write returns 0, file comes out clean.

Two small ones while you're in there:

- The CONTRIBUTING note says `git rm --cached -r . && git reset --hard`. reset --hard eats uncommitted work and whoever is reading that is in the middle of something. Add "commit or stash first".
- "both fail on the unfixed code" - I reverted the script here and only the note test went red. The LF test passes on mac and linux whatever you do, which is exactly why it goes in the Windows job. Just say it's Windows-only so nobody goes chasing it.

The rest I checked and it's fine: the threshold only counts rows that already existed and changed (1 of 4 stays quiet, 30 of 100 warns, adding 10 files to 10 stays quiet), the note is on stderr, second write is byte for byte identical.

I'll close #53 when this is in.

Sreekant13:
Thank you, and that is a genuinely good catch. All three are done in 766103b.

**The `parse_manifest` bug.** You are right and it is worse than a rough edge: `--write` is the way out of a manifest left mid-merge, the manifest conflicts on nearly every open PR because it is generated, and I had used that exact recovery twice myself on #30. So the branch broke the path it exists to support. I reproduced your failure before fixing it:

```
$ python scripts/check_release_manifest.py --root <repo> --write
RELEASE_MANIFEST.txt:2: cannot parse '<<<<<<< HEAD' — every row is `<sha256>  <path>`
```

The read now goes through `_previous_rows`, which returns empty when the file cannot be read. `parse_manifest` raising on the first bad row is correct for the check side, so I left it alone and made the write side tolerant instead. Losing the comparison costs the note and never the regeneration.

Two tests, both failing without the change:

- `test_write_recovers_a_manifest_left_mid_merge`: conflict markers in, `--write` returns 0, the markers are gone, **and the checker accepts the result afterwards**. That last assertion is the one that matters, since recovery means a usable manifest rather than merely not crashing.
- `test_an_unreadable_previous_manifest_costs_only_the_note`: an unparseable manifest still writes, and stays quiet.

**The CONTRIBUTING note.** Fixed. It now says to commit or stash first, and says plainly that `reset --hard` discards uncommitted work. You are right that whoever is reading that line is in the middle of something.

**"both fail on the unfixed code".** That was sloppy of me and I have corrected the PR body rather than leave it standing. The LF test passes on mac and linux whatever the script does, which is precisely why it goes in the Windows job; my stash-prove was run on Windows and I wrote it up as though it were universal. The body now says so explicitly so nobody goes chasing it.

Also merged main and repinned; the manifest diff is still four rows, one per file this PR changes.

```
$ uv run pytest -q tests/test_check_release_manifest.py
17 passed, 7 skipped

$ git stash push -- scripts/check_release_manifest.py
$ uv run pytest -q tests/test_check_release_manifest.py
2 failed, 15 passed, 7 skipped

$ uv run pytest -q tests/test_check_release_manifest.py tests/test_ci.py tests/test_structural_budget.py tests/test_readme_claims.py tests/test_derived_conflict_inventory.py
277 passed, 19 skipped

```

And thank you for the note about #53. It seemed the right way to handle it, and I appreciate you saying so.

eyalgolan (approved):
Great work @Sreekant13 , approving and merging.

Sreekant13:
Thanks for merging, and for the review that caught the `--write` recovery break. That one mattered: I had used that exact path twice myself on #30 and still managed to break it, which is a good argument for the reviewer being someone other than the author.

PR #89, opened 2026-09-04
docs: drop the fictional nh integrations verb, complete the docs index

### LINKED ISSUE #11: docs/security.md references a non-existent `nh integrations` command
\`docs/security.md\` (the "Integration health checks" bullet, around line 218) says *"\`nh integrations\` / the board's integrations page authenticate against…"*.

\`\`\`
$ nh integrations --help
Error: No such command 'integrations'.
\`\`\`

The board surface exists (\`/api/integrations\` and \`/api/integrations/{name}/test\` in \`src/no_human/api/app.py\`); only the CLI verb is fictional. The bullet's inline citations into \`src/no_human/integrations/__init__.py\` have also drifted — the status surface is \`list_integrations\` (around line 373).

**Fix:** drop \`nh integrations\` from the sentence (keep the board page) and re-point the two line citations at \`list_integrations\`. Docs only, ~3 lines.

### LINKED ISSUE #12: docs/README.md index misses a dozen published docs, including BACKENDS.md
\`docs/README.md\` lists 11 documents. These published docs are not reachable from it:

\`\`\`
docs/BACKENDS.md   docs/desktop.md   docs/DISTRIBUTION.md   docs/COST_LEVERS.md
docs/NIGHTLY_EVAL.md   docs/NORTH_STAR_BENCH.md   docs/REVIEWER_RECALL_METHOD.md
docs/UNTRUSTED_PR_REVIEW.md   docs/ARCH_REVIEW.md   docs/UI_AUDIT.md
docs/BASELINE_M0.md   docs/design/memory-lifecycle-triage.md
\`\`\`

\`BACKENDS.md\` is the visible gap: \`docs/configuration.md\` sends readers there for \`worker.backend\`, but neither the index, the quickstart nor the README link it.

**Fix:** add rows for at least \`BACKENDS.md\`, \`desktop.md\`, \`DISTRIBUTION.md\` and \`COST_LEVERS.md\` to the table (a second "Reference / reports" table for the audit-style docs is fine). Docs only, ~12 lines.

Sreekant13:
## What this changes

Fixes #11. Fixes #12.

Both are docs-only and both would repin `RELEASE_MANIFEST.txt`, so as two open PRs one was guaranteed to conflict with the other. They are together here for that reason. Say the word and I will split them.

### #11, the fictional CLI verb

`docs/security.md` named `nh integrations`. There is no such command, while the board surface the same sentence names is real (`/api/integrations` and `/api/integrations/{name}/test` in `api/app.py`). The sentence keeps the board half and loses the CLI half.

The citations needed more than a re-anchor, and this is the part worth your eye. Both pointed at `_probe_github_ambient`, whose own docstring says it reports **presence only, never validity, with no network round-trip**. Neither citation supported the claim the sentence actually makes, that the page authenticates against Jira, Linear, CircleCI and Teams to show a live status.

So the bullet now cites `test_integration`, which is the live health check that reaches those services, and the GitHub probe keeps both of its citations against a clause that describes what it really does.

**I did not use `list_integrations` as the issue suggested.** Its docstring reads "Every integration's configured/kind status. Pure; `healthy` is None." Citing a pure function under a heading about what leaves the machine would swap one unsupported citation for another. If you read that differently, say so and I will change it.

### #12, the docs index

The index linked 13 of the 26 docs on disk. `BACKENDS.md` had already been added by the time I picked this up, and two docs have appeared since the issue was filed, `pr-body.md` and `design/lease-takeover-vs-orphan-grace.md`. Both are included.

`desktop.md`, `DISTRIBUTION.md`, `COST_LEVERS.md`, `pr-body.md` and `UNTRUSTED_PR_REVIEW.md` join the main table. The dated measurements and decision records go in a second "Reference and reports" table with a line saying a report may predate the current code, so they do not bury the docs a reader needs early.

Descriptions are taken from each document's own opening lines rather than written from the filename.

## Evidence

```
VERDICT=OK

$ uv run pytest -q tests/test_readme_claims.py tests/test_reanchor_citations.py
128 passed, 12 skipped

$ git diff --stat
 RELEASE_MANIFEST.txt        |  6 +++---
 docs/README.md              | 21 +++++++++++++++++++++
 docs/security.md            | 11 +++++++----
 tests/test_readme_claims.py |  8 +++++---
```

Every link in the index is checked to resolve on disk, and no doc under `docs/` is unreachable from it any more:

```
broken links : none
still missing: none
total linked : 26
```

Nineteen tests fail for me across `test_guard.py`, `test_scope_guard.py`, `test_tamper_guard.py`, `test_agent_commit_identity_enforced.py` and `test_child_env.py`. They fail identically on a clean worktree of `main`, so they are this platform rather than this change.

## Checklist

- [x] Tests cover the change, and they fail on the unfixed code. (Docs only; the citation gate is the test, and the `CITATION_TABLE` row moves with the doc.)
- [x] Test count and assertion count did not go down. The citation table gains a row, 2 to 3.

`RELEASE_MANIFEST.txt` repinned last, three rows, one per file changed.


eyalgolan (changes requested):
Thanks @Sreekant13 . Doing #11 and #12 together was right, two open PRs would just have fought over the manifest.

The index is done as far as I'm concerned. All 26 docs are linked now, I checked every link resolves, and the descriptions match what the files actually say rather than what the filenames suggest. Putting the dated reports in their own table with the "check the date" line is how I'd have done it too. Merges clean, gates are green here.

The security.md bit needs another pass, and it's exactly the sentence you asked me to look at, so good instinct.

You wrote that GitHub is the exception - presence only, no network round-trip. That's only true when GitHub *isn't* configured. Look at what test_integration actually dispatches to for "github":

```
_CHECKERS = { ... "github": _check_github, ... }          # integrations/__init__.py:1568

async def _check_github(config):                            # :1462
    base = _github_status(config)
    if not base.configured:
        return await _unconfigured_or_ambient(base, "github")   # <- the presence probe lives here
    token = os.environ.get("GITHUB_TOKEN") or os.environ.get("GH_TOKEN")
    ...
    u = await _http_get(f"{api}/user", headers=headers, timeout=10.0)
    r = await _http_get(f"{api}/repos/{project}", headers=headers, timeout=10.0)
```

So with GitHub configured, the test button on the board sends your token to api.github.com (or your GHES host) twice. _probe_github_ambient only runs on the unconfigured path. As written, the bullet now says GitHub never leaves the machine, and understating what goes out is the one mistake this page can't make.

Split it in two: configured GitHub gets checked like the others, a live call with the token (cite _check_github:1462); unconfigured GitHub is presence only, no round-trip (keep both _probe_github_ambient citations). And add a CITATION_TABLE row for _check_github like you did for test_integration.

Not citing list_integrations was right, for the reason you gave.

One more thing, for me not you. I moved test_integration down ten lines on a scratch copy and neither the exact-anchor test nor reanchor_citations noticed. So for the `symbol:line` rows the number isn't actually checked. That's a hole in the gate, not in your PR - I'll open an issue for it.

Sreekant13:
You are right, and this was the worst kind of error to make on that page: I understated what leaves the machine.

Fixed in d910192. I confirmed your reading in the code before changing anything. `_CHECKERS["github"]` at :1574 dispatches to `_check_github` at :1462, whose own docstring says "Real GitHub probe: the token authenticates AND the configured repo is reachable". The early return to `_unconfigured_or_ambient` happens only when `base.configured` is false; past that it reads `GITHUB_TOKEN` or `GH_TOKEN` and makes two authenticated `_http_get` calls, to `{api}/user` and `{api}/repos/{project}`.

The bullet now splits it exactly as you described: a configured GitHub is checked like every other integration, with the token going to api.github.com or the GHES host, citing `_check_github:1462`; only an unconfigured one is presence-only, keeping both `_probe_github_ambient` citations. `CITATION_TABLE` gains a `_check_github` row alongside the `test_integration` one.

The mistake underneath it is worth naming, because it is the second time this week I have made it. On #30 I put `AuthError` and `ConfigError` into a catch tuple because your review listed them as things `load_config` raises, without asking what each one *means*. Here I read `test_integration` and `_probe_github_ambient` and stopped, without following the dispatch table to what actually runs for "github". Same shape both times: I trusted the symbol I was pointed at instead of following the call through. I have written that down as a check to run rather than a lesson to remember.

On the gate hole you found, that a `symbol:line` row does not actually verify the line number: that matches what I hit from the other side on #30, where the drift was 11 lines and `reanchor_citations.py` refused to guess and left it to a human. Happy to take the issue if you open it, though it is your call whether you would rather scope that yourself.

```
VERDICT=OK

$ uv run pytest -q tests/test_readme_claims.py tests/test_reanchor_citations.py tests/test_egress_disclosure.py tests/test_egress_allowlist.py
157 passed, 12 skipped in 86.80s

```

And thank you for the note on the index half. Doing #11 and #12 together was mostly about the manifest, so it is good to hear the second table read the way it was meant to.

eyalgolan:
@Sreekant13 Thanks, the GitHub half reads well now. I merged it with main on my side, reanchor is OK, all four cited lines sit where the table says, and the Python job passed.

While I was in that sentence I noticed one more thing, and I want to be clear it predates your PR. The list says "Jira, Linear, CircleCI, GitHub and the Teams webhook". Checking it against the dispatch table:

- Teams isn't actually called. _check_teams is a status view (its docstring says so, and VIEW_ONLY_CHECKS is {slack, teams}), no HTTP at all.
- monday, GitLab and Jenkins are called, with the credential: _check_monday sends a Bearer token, _check_gitlab hits {hostname}/api/v4/user with PRIVATE-TOKEN, _check_jenkins does an authenticated request. They're just missing from the list.

Since you're already rewriting that line, would you mind fixing the list too? The seven that make a live call are Jira, Linear, monday, CircleCI, GitHub, GitLab and Jenkins; Slack and Teams are views and send nothing (the run-time webhook POST is covered in the bullet above). Same one-sentence shape as the GitHub change. I'll approve as soon as that's up.

I'll open an issue for the gate hole. Happy for you to take it if you're interested, no pressure either way.

Sreekant13:
Fixed in 11fbe0c, and thank you. The list was wrong in both directions, which is worse than being incomplete.

I checked each one against `_CHECKERS` rather than taking the old sentence's word for it:

- **Teams was listed and does not call anything.** `_check_teams` reads `_teams_status` and returns; there is no HTTP in the body at all. Its docstring says the omission is deliberate, since the only way to exercise a Workflows webhook is to post a message into a human's channel. It sits in `VIEW_ONLY_CHECKS` with Slack.
- **monday, GitLab and Jenkins were missing and do send the credential.** `_check_monday` a Bearer token, `_check_gitlab` `{hostname}/api/v4/user` with `PRIVATE-TOKEN`, `_check_jenkins` an authenticated request.

The bullet now names the seven that make a live call (Jira, Linear, monday, CircleCI, GitHub, GitLab, Jenkins) and says plainly that Slack and Teams send nothing, with the run-time webhook POST already covered by the bullet above. `CITATION_TABLE` gains a `VIEW_ONLY_CHECKS` row so that claim is anchored like the others.

One thing from my own side worth admitting: my first pass at checking this counted `_http_get` occurrences near each `_check_*` and got a false positive on Teams, because the match was inside the docstring. Reading the body is the only thing that actually answers the question. That is the same mistake as citing the symbol the docs named without following the dispatch, just one level down.

```
VERDICT=OK

$ uv run pytest -q tests/test_readme_claims.py tests/test_reanchor_citations.py tests/test_egress_disclosure.py tests/test_egress_allowlist.py
158 passed, 12 skipped

```

On the gate hole, yes, I would like to take it if you open the issue. Having hit it from the other side on #30, where the drift was 11 lines and the tool correctly refused to guess, I have a fairly clear picture of where the seam is.

eyalgolan (approved):
Another great PR @Sreekant13 . Approving and merging 🚀

Sreekant13:
Thanks for merging, and for both rounds of the review. The GitHub half and then the live-check list were each wrong in a way I would not have found by re-reading my own diff, so I appreciate you going through that sentence twice.

PR #92, opened 2026-09-04
nh onboard: show why a proving candidate failed

### LINKED ISSUE #18: `nh onboard` reports a failed proving command without a line of its output
When `nh onboard` proves candidate commands, a failure prints the exit code and nothing else:

```
proving (running each candidate):
  ✗ [FAILED] test: pytest -q  (from python/pytest, exit 1)
...
test command NOT proven — profile is not usable until it runs clean.
```

Refusing to confirm an unproven test command is correct. Refusing with **no diagnostic** is where a new user stops: they cannot tell a missing dependency from an import error from a genuinely failing test. In the run that found this, the cause was a one-line fixture problem the captured stderr would have named immediately. This is [KI-4 in docs/KNOWN_ISSUES.md](https://github.com/no-human-ai/no_human/blob/main/docs/KNOWN_ISSUES.md).

**The data is already there.** `ProveOutcome` captures the combined output ([src/no_human/onboard.py:120-125](https://github.com/no-human-ai/no_human/blob/main/src/no_human/onboard.py#L120-L125) — the `output: str` field) but the renderer prints only `p.summary` ([src/no_human/cli/commands.py:3083](https://github.com/no-human-ai/no_human/blob/main/src/no_human/cli/commands.py#L3083)), which formats verdict, kind, command, source, and exit code — never `output`.

**A fix has to prove** (acceptance criteria from KI-4):
1. A failed proving candidate prints the last N lines of its combined output, attributed to the command (a passing one stays quiet).
2. Output is truncated, not unbounded — a 10,000-line pytest failure must not flood the terminal.
3. Nothing secret is echoed: the proving subprocess inherits the process env, which by then holds loaded `.env` values — the fix must not print environment contents, only the command's own output.

**Verification:** a test that onboards a fixture repo whose declared test command fails with a known stderr line, and asserts that line (truncated) appears in the rendered output; plus the existing onboard tests stay green (`uv run pytest -q tests/test_onboard.py`). No credential needed ([CONTRIBUTING.md](https://github.com/no-human-ai/no_human/blob/main/CONTRIBUTING.md)).

Sreekant13:
## What this changes

Fixes #18. KI-4.

`ProveOutcome.output` already held the combined output; the renderer only ever printed `p.summary`. A failed candidate now prints the tail of its own output, attributed to the command.

The formatting lives on `ProveOutcome` rather than in the renderer, so `cli/commands.py` stays at 8567 lines and its frozen budget is untouched. The call site is the same length it was.

## Before and after

Same fixture both times: a repo whose `test_demo.py` imports a package that is not installed.

Before, on `main` today:

```
proving (running each candidate):
  ✗ [FAILED] test: pytest -q  (from python/pytest, exit 1)

proposed profile:
test command NOT proven — profile is not usable until it runs clean.
```

After:

```
proving (running each candidate):
  ✗ [FAILED] test: pytest -q  (from python/pytest, exit 1)
      ... 4 earlier line(s) not shown
      Traceback:
      ...\importlib\__init__.py:90: in import_module
          return _bootstrap._gcd_import(name[level:], package, level)
      E   ModuleNotFoundError: No module named 'psycopg2'
      =========================== short test summary info ===========================
      ERROR test_demo.py
      !!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
      1 error in 0.28s
```

## Against the three criteria

1. **Failed candidates only.** `failure_tail()` returns `""` when the candidate passed, and also when the command said nothing, so a clean prove run reads exactly as it did before this existed.
2. **Bounded, on both axes.** The last 12 lines, each clipped to 200 characters. A 10,000-line pytest failure renders as 13 lines; a single 5,000-character line renders as one clipped line. Whatever is dropped is announced (`... 9988 earlier line(s) not shown`, `... (+4800 chars)`) rather than silently cut, because a truncated diagnostic that looks complete is its own trap.
3. **Nothing from the environment.** Only `output` is rendered, which is the command's own stdout and stderr. The proving subprocess inherits the process environment, which by then holds the values loaded from `~/.no_human/.env`, and nothing derived from it is printed. `test_nothing_from_the_environment_is_rendered` puts sentinels in the env and asserts the rendered tail is exactly the command's line and nothing more.

One thing I said I would check before pushing rather than assume: whether the captured output can itself contain a secret. It can, if the proved command echoes one of its own (a tool printing its config, say). That is the command's own output rather than anything this code reaches for, and showing it is what the issue asks for, but it is a real property of the change and I would rather name it than leave you to find it. If you would prefer a scrub pass over the tail as well, say so and I will add one.

The output is also escaped before it reaches the console. `[` opens a Rich markup tag, and an unclosed `[red]` inside a traceback would otherwise be swallowed or would raise.

## Evidence

```
$ uv run pytest -q tests/test_onboard.py
23 passed in 20.43s

$ git stash push -- src/no_human/onboard.py src/no_human/cli/commands.py
$ uv run pytest -q tests/test_onboard.py
9 failed, 14 passed in 9.09s

$ uv run pytest -q tests/test_structural_budget.py tests/test_onboard.py tests/test_onboarding_api.py tests/test_profile.py tests/test_gui_prove_e2e.py tests/test_adoption_harness.py tests/test_profile_divergence.py tests/test_project_config.py
11 failed, 157 passed in 35.22s

```

`tests/test_onboard.py` goes from 14 tests to 23, and all nine new ones fail without the change.

The 11 failures are in `test_onboarding_api.py` (the `discover` group), `test_gui_prove_e2e.py` and `test_project_config.py`. They fail identically on a clean worktree of `main`: 11 failed there too, with 148 passed against my 157, the difference being exactly these nine new tests. They are this platform, not this change.

## Checklist

- [x] Test count and assertion count did not go down. `test_onboard.py` 14 to 23.

`RELEASE_MANIFEST.txt` repinned last, three rows, one per file changed.


eyalgolan (changes requested):
Thanks @Sreekant13 , this is a good one. Keeping the formatting on ProveOutcome so commands.py doesn't grow was the right call, the truncation says what it dropped instead of hiding it, and the env sentinel test is the kind of test I like to see. Merged it with main on my side, everything in the onboarding area passes, and the nine new tests go red without the change.

Three things and then it's in.

The escaping has a small gap. Replacing `[` with `\[` is fine until there's already a backslash there, which is what a Windows path looks like at the end of a traceback line. I ran one through your renderer:

```
'ends with one backslash\\'   ->   'ends with one backslash[/]'
```

The backslash eats the closing tag, so the line loses its last character and prints a literal `[/]`. rich.markup.escape handles this case (I checked, same input comes out intact), so I'd use that instead of the hand-rolled replace, and add a trailing-backslash line to test_command_output_cannot_inject_console_markup so it stays that way.

On the scrub pass you offered: yes please. There's already the right tool in the tree, `_redact` in agent/verification_receipts.py. It masks live env values (and their base64 forms) plus the usual credential shapes on a command line, and the module is stdlib-only so importing it from onboard.py costs nothing. Run the tail through it before rendering. A proved command that prints its own config is exactly the case you described, and a key in a terminal is the one class of bug we treat as a P1 here.

Last one, KI-4 in docs/KNOWN_ISSUES.md still says `**Status:** open`. The convention there is `**Status:** CLOSED <date> by <what>`, KI-3 and KI-5 have the shape. A line pointing at this PR is enough, the rest of the entry can stay as the record.

For what it's worth, the board was never affected. The onboarding stream in api/app.py already sends the last 4000 chars of each proof's output, so this really was CLI-only.

Sreekant13:
All three done in fff7367, and thank you for the escaping one in particular.

**The escaping gap.** I reproduced your case before touching anything, and it is exactly as you described:

```
input        : 'ends with one backslash\'
mine         : '[dim]ends with one backslash\[/]'
rendered     : 'ends with one backslash[/]'      <- tag eaten, last char lost
rich.escape  : 'ends with one backslash\\'      <- intact
```

The hand-rolled replace is gone in favour of `rich.markup.escape`. `test_a_line_ending_in_a_backslash_keeps_its_closing_tag` covers it and asserts on two levels: that the markup still ends in a closing tag, and that a real `Console` renders no literal `[/]` and keeps the trailing backslash. Rendering it was the part that would have caught this the first time; asserting on the markup string alone would not have.

**The scrub pass.** In, using `_redact` as you suggested rather than anything of my own. It runs over the output before splitting, so it covers live env values, their base64 forms, and the command-line credential shapes. Imported inside the method so `onboard.py`'s import cost is unchanged. The test sets a real `sk-ant-` value in the environment, has the command echo it in a config dump, and asserts it does not reach the tail while the surrounding lines still do.

Worth saying: I raised this possibility in the PR body and then left it as a question rather than acting on it. Given that a key in a terminal is a P1 here, offering it and waiting was the wrong instinct. Thank you for the pointer to the existing helper; I would have written a worse one.

**KI-4** now reads `**Status:** CLOSED 2026-09-04 by ProveOutcome.failure_tail, which renders the tail of a failed candidate's own output under the summary line`, matching the shape KI-3 and KI-5 use, with the rest of the entry kept as the record of what a fix had to prove.

Also merged main, and `cli/commands.py` is still 8567 so the frozen budget is untouched.

```
$ uv run pytest -q tests/test_onboard.py
25 passed in 16.55s

$ git stash push -- src/no_human/onboard.py
$ uv run pytest -q tests/test_onboard.py
FAILED tests/test_onboard.py::test_a_line_ending_in_a_backslash_keeps_its_closing_tag
FAILED tests/test_onboard.py::test_a_credential_in_the_command_output_is_masked
2 failed, 23 passed

OK: 1518 file(s) match RELEASE_MANIFEST.txt
```

The 7 failures in `test_onboarding_api`'s discover group are identical on a clean worktree of current main (ce4ff43), so they are this platform. I re-checked that against today's main rather than reusing the earlier result, since main has moved fourteen commits.

Good to know the board was never affected. That is the kind of scoping detail I would not have found on my own.

eyalgolan (approved):
Great work as always @Sreekant13 . Approved.

PR #95, opened 2026-09-04
Citation gate: check the line in a path:symbol:line row

### LINKED ISSUE #93: Citation gate does not check the line number in `path:symbol:line` rows
`tests/test_readme_claims.py` has two forms of citation row. A bare `path:line` row is exact-anchored: any drift fails `test_every_line_citation_currently_resolves_exactly` (that is what caught `intake/mcp_bridge.py:29` moving to 30 in #30). A `path:symbol:line` row is not: once a symbol is present, the line number is never looked at, by the test or by `scripts/reanchor_citations.py`.

Repro on a clean checkout of `main` (aa407cffc):

```
$ uv run pytest -q tests/test_readme_claims.py -k "citation or exactly"
52 passed, 2 skipped
VERDICT=OK

# move test_integration ten lines down
$ python - <<'PY'
p="src/no_human/integrations/__init__.py"; L=open(p).read().split("\n")
i=next(k for k,l in enumerate(L) if l.startswith("async def test_integration"))
for _ in range(10): L.insert(i, "# planted")
open(p,"w").write("\n".join(L))
PY

$ uv run pytest -q tests/test_readme_claims.py -k "citation or exactly"
52 passed, 2 skipped
VERDICT=OK
```

`docs/security.md` cites `integrations/__init__.py:test_integration:1591`; after the plant the function starts at 1601 and nothing goes red.

Positive control, same tree: insert one line above `BASE_URL` in `intake/mcp_bridge.py` (cited as a bare `intake/mcp_bridge.py:40`) and the exact-anchor test fails immediately. So the gate works for bare rows and is silent for symbol rows.

Expected: for a `symbol:line` row, resolve the symbol, then give the cited line the same exact / drifted / missing verdict a bare row gets, and have `reanchor_citations.py --apply` rewrite the number. The symbol rows for `integrations/__init__.py` in `CITATION_TABLE` (two on `main` today, five once #89 lands) are a ready-made fixture set.

Found while reviewing #89. The 11-line refusal Sreekant13 hit on #30 was the bare-row path working as designed; this is the symbol-row path not working at all.

Sreekant13:
## What this changes

Fixes #93.

Two commits, deliberately separable: **the gate and the tool**, then **the 15 numbers that gate exposes**. The second is machine-produced and reads on its own.

### The gate

A `path:symbol:line` row resolved the symbol and then threw the number away, in the test and in `reanchor_citations.py` alike. Reproduced with your plant: `integrations/__init__.py:test_integration:1591` stays green with the function on 1601.

The line is now verified, with the same verdict a bare row gets. Inside the drift window it warns; beyond it, it fails. There is no `missing` case, because the symbol and the token have both already resolved by that point.

Three decisions I would rather state than bury:

- **Anchored on the token, not the definition.** This is forced, not preferred. `:_probe_github_ambient:549` cites a docstring sentence inside a function that starts at 515. Anchoring on the `def` would report that row as 34 lines wrong when it is exactly right. All five `integrations` rows are exact under token anchoring.
- **A range's end is not verified.** A symbol's length changes whenever its body is edited, so tracking the end would report an unrelated edit as citation rot. On a rewrite the end shifts by the same delta as the start, so `829-830` becomes `875-876` and the span is preserved.
- **An injected `source_text` is never line-checked.** That buffer is synthetic and its line numbers mean nothing. Checking them fails `test_symbol_citation_resilience`, whose entire purpose is that a symbol citation survives a shift. I hit that on the first run.

### The tool

`plan()` now handles both forms, and its docstring says why they are fixable for different reasons. A bare row is anchored by **proximity**, so a distant match is a guess and is refused, which is the behaviour that correctly left my 11-line drift on #30 to a human. A symbol row is anchored by the **symbol**, which resolves however far the code has moved, so its number can be rewritten exactly at any distance.

That distinction is what makes the second commit mechanical rather than a judgement call.

### One bug found on the way

`reanchor_citations.py` had the same CRLF defect that #32 fixed in `check_release_manifest.py`: `write_text` without `newline="\n"`. On Windows a 15-citation change landed as **3836 insertions**, including 1024 lines of `security.md` that were only line endings. With `newline="\n"` the same change is **27 insertions**. Fixed here since I was in the writer anyway; say the word if you would rather it were split out.

## The 15 rows, and the alternative I did not take

The second commit is `--apply` output with no hand edits. These rotted precisely because nothing looked at them:

```
cli/commands.py:bench_run:7123      -> 7633   (+510)
:GitRepo.fetch:916                  -> 1103   (+187)
cli/commands.py:approve:4955        -> 5085   (+130)
northstar_card.py:render_northstar_md:1474-1478 -> 1555-1559
```

I had offered you a choice on the issue between one PR and two, and said I leaned to two so the gate could land warning-only first. **That option turns out not to exist**, and it is worth recording why: `tests/test_reanchor_citations.py::test_check_mode_is_clean_on_this_tree` asserts `--check` is clean on the real tree. The moment the tool can see symbol rows, `--check` is dirty until the numbers move. So the tool change and the re-anchor have to land together, and once they do there is nothing left for a warning-only phase to buy.

They are still two commits, so if you would rather review them apart, or want the re-anchor as its own PR on top of the gate, that split is one `git revert` away and I am happy to do it.

## Evidence

```
$ uv run pytest -q tests/test_readme_claims.py tests/test_reanchor_citations.py
137 passed, 12 skipped in 22.09s          # zero warnings

VERDICT=OK

OK: 1519 file(s) match RELEASE_MANIFEST.txt

$ uv run pytest -q tests/test_readme_claims.py tests/test_reanchor_citations.py tests/test_structural_budget.py tests/test_ci.py tests/test_derived_conflict_inventory.py
275 passed, 12 skipped in 130.45s
```

Before the second commit the same citation suites pass with **15 warnings**, one per stale row, which is the gate doing its job.

Seven new tests: the planted shift, an exact row staying silent, token-not-definition anchoring, a row with no line at all, a range's start being what is read, injected source being skipped, and a beyond-window row failing rather than warning.

## Checklist

- [x] Test count and assertion count did not go down. `test_readme_claims.py` gains 7.


eyalgolan (changes requested):
Thanks @Sreekant13 , this is good work, and you built it the way I'd have asked for it: the symbol resolves, the token anchors the number, start only, injected source left alone.

I merged it with main on my side and then went at it with plants. Ten lines under test_integration turns the per-row test red and `--check` reports FAIL. Three lines gives one warning and `test_check_mode_is_clean_on_this_tree` goes red, so a small drift can't ride through CI as a warning nobody reads. The bare-row control still fails. And `--apply` on the ten-line case rewrites the doc and the table to :1601 exactly. The 15 re-anchors are numbers only, I diffed both docs. Good call on `newline="\n"` in the writer while you were in there, same bug as #32, no reason to make it its own PR.

You were also right that the two-PR option didn't exist. `test_check_mode_is_clean_on_this_tree` makes the tool and the re-anchor one change, and that's fine by me.

One thing before I approve. Three places still describe the old contract:

- the comment above `_LINE_CITATION_RE`: "as a purely advisory, never-checked hint"
- the `_SYMBOL_CITATION_RE` header right under it: "parsed but never checked — the whole point is that it may go stale"
- `_check_citation`'s docstring: "parsed and then ignored — the whole point of a symbol citation is to survive that range going stale"

After this PR each of those says something the code no longer does, and this repo is picky about that. Rewrite them to the new contract: the symbol resolves the citation and survives a move; the number is checked against the token's line, warns inside the window, fails beyond it, and `--apply` rewrites it. Same commit as the behaviour so a reader never sees the two disagree.

Small thing, not blocking. `_symbol_first_line` finds the vicinity by searching for its content and takes the first match. When the AST path resolved the symbol it already has `node.lineno`; using that and falling back to the search only for the regex path would make two identical function bodies in one file anchor correctly. Happy to leave that for a follow-up.

Sreekant13:
Thanks for going at it with your own plants. That is a better check than anything I could have written for you, and it is good to know the three line case turns `test_check_mode_is_clean_on_this_tree` red rather than sitting in a warning nobody reads.

All three comments are rewritten, in the same commit as the behaviour.

While I was there I found a fourth of the same kind, and one of them was mine:

- the comment above the symbol branch in `_check_citation`, "never consulted for pass/fail, only parsed so the format allows it"
- `_advisory_line`, the helper I added in this PR. The name says the number is advice, and after this PR it is enforced. Renamed to `_cited_line`, along with its test and the one call in `reanchor_citations.py`.

That is the same defect one layer down, so I would rather not leave it for you to find later.

## I took the `_symbol_first_line` note rather than deferring it

You are right, and it turned out to be a real anchoring bug rather than a tidy up, so I did not want to ship it knowingly wrong.

A top level pair cannot actually collide, because the `def` line carries the name and makes the two vicinities differ. A method pair can:

```python
class Alpha:
    def probe(self):
        """MARKER PHRASE"""
        return 1

class Beta:
    def probe(self):
        """MARKER PHRASE"""
        return 1
```

Those two vicinities are byte identical, so the content search returned 5 for both, and `Beta.probe` on line 11 would have been cited as 5. Before and after, on that fixture:

```
OLD content-search:  Alpha.probe -> 5   Beta.probe -> 5
NEW AST lineno:      Alpha.probe -> 5   Beta.probe -> 11
```

`_symbol_vicinity_span` now returns the number the AST already had. `_symbol_vicinity` stays as a thin wrapper so the other four callers are untouched. The content search survives only for the regex fallback path, which recognizes a symbol without ever numbering it, and there the first match is genuinely the best answer available. One test, `test_twin_methods_anchor_to_the_right_one`, which asserts the two vicinities are equal first so it is clear what is being defended against.

It is in the same commit. If you would rather it were separate, or out, say so and I will move it.

## Rebased rather than merged

To get the comment fix inside the behaviour commit I rebased onto main at 0.2.0 instead of merging again, so the branch is three clean commits and the merge commits are gone. Force pushed. That also meant re-running `--apply` against the new main rather than replaying the old numbers: still 15 rows, and the manifest is re-pinned last.

The re-anchor commit is 27 insertions and 27 deletions across the two docs and the table.

## Re-ran your plants on the rebased branch

Ten lines above `test_integration`:

```
$ uv run pytest -q tests/test_readme_claims.py -k test_integration
FAILED ...[security.md:integrations/__init__.py:test_integration:1591]

DRIFT: security.md `...:test_integration:1591` -> `...:test_integration:1601` (would re-anchor)
VERDICT=FAIL

$ uv run python scripts/reanchor_citations.py --apply
applied 1 re-anchor(s)
docs/security.md:406:  (`integrations/__init__.py:test_integration:1601`)
tests/test_readme_claims.py:2023:    ("security.md", "integrations/__init__.py:test_integration:1601",
```

Three lines: one warning at "3 line(s) out", the per row test still passes, and `test_check_mode_is_clean_on_this_tree` fails. Matches what you saw.

Worth recording, because it cost me two attempts: my first plant inserted the blank lines in the middle of `async def`, which broke the parse. The gate did not go red for the reason I expected, it went to the regex fallback with an AST parse warning. So the fallback path does what it says on a genuinely unparseable file.

## Evidence

```
$ uv run pytest -q tests/test_readme_claims.py tests/test_reanchor_citations.py
138 passed, 12 skipped in 20.61s          # zero warnings

VERDICT=OK

OK: 1519 file(s) match RELEASE_MANIFEST.txt

$ uv run pytest -q tests/test_structural_budget.py tests/test_ci.py tests/test_derived_conflict_inventory.py
138 passed in 97.92s
```

eyalgolan (approved):
That's the whole thing, thanks @Sreekant13 . Nice catch on the fourth passage, and `_cited_line` is the honest name now.

Glad you took the twin-methods one instead of parking it. I merged the rebased branch with main, ran the citation suites with warnings promoted to errors (clean), then did the plants again. Ten lines under test_integration fails the per-row test and `--check`, and `--apply` rewrites both places to :1601. Three lines trips `test_check_mode_is_clean_on_this_tree`. The bare-row control still fails. And if I make `_symbol_first_line` ignore the AST lineno, your twin test goes red on Beta.probe. That's everything I wanted to see.

Approving.

eyalgolan (approved):

Sreekant13:
CI is green on your merge commit. Worth flagging why it was not the first time, since it looks like a real intermittent rather than noise.

The first run on `2e3d303c` reached 93% at 09:42:58, then produced no output at all for 22 minutes and was cancelled at 10:05:18 against `timeout-minutes: 30`. The re-run on the same commit passed `Python` in 6m3s. So it hung rather than ran slow, and it did not reproduce.

I checked the things that would have made it ours before concluding that. Your merge is byte-identical to a clean auto-merge (`git merge-tree --write-tree 0e448df 9320964` gives the same tree, `7eea8ac3`), the citation suites pass on the merged tree with zero warnings, and the only source file main added was `integrations/health.py`, which no citation points at. My `0e448df` had also passed CI twice before your merge.

Nothing needed from me. Flagging the hang only because it costs the full 30 minutes each time it happens, and a silent stall leaves no traceback to work from.

Ready whenever you are.

Issue #19, opened 2026-08-31
Re-measure the KI-1 concurrent-commit crash now that Store writes are serialized, then re-enable or retire the deselected test


eyalgolan:
[KI-1 in docs/KNOWN_ISSUES.md](https://github.com/no-human-ai/no_human/blob/main/docs/KNOWN_ISSUES.md) records a real product defect: with two tasks running concurrently against one `Store` (`concurrency.enabled: true`, `max_workers > 1`), an attempt could die on

```
sqlite3.OperationalError: cannot commit transaction - SQL statements in progress
```

with the traceback entirely in shipped code (`Orchestrator._run_attempt` → `Store.update_attempt`). The only test that drives that configuration, `tests/test_scheduler.py::test_two_repos_run_concurrently_in_worktrees`, measured a 3/8 serial failure rate on 2026-07-30 and is **deselected in CI** to keep the badge truthful.

Since then, a mitigation landed: every `Store` write goes through the `serialized_write` lock in `src/no_human/core/db.py`. **The failure rate has not been re-measured since.** KI-1's own status line says: re-measure, then re-enable or close.

**The task** — measurement first, code second (possibly no code at all):
1. Re-measure: run `uv run pytest -q tests/test_scheduler.py::test_two_repos_run_concurrently_in_worktrees` serially, ≥10 times, on current `main`, and record pass/fail per run. (It reproduces without xdist — the concurrency that matters is inside the test, two asyncio tasks sharing one `Store`.)
2. If it now passes consistently: re-enable it in CI (remove the deselect in `.github/workflows/ci.yml`), and move KI-1 to closed with the new measurement in the entry.
3. If it still fails: KI-1's "what a fix has to prove" section defines the bar — the numbers you measured become the before/after evidence for whoever fixes it (which could be you, as a follow-up).

Either outcome is a complete, mergeable contribution: this issue is about replacing a stale measurement with a current one and making CI's selection honest again. Note the docs/KNOWN_ISSUES.md rule: an entry leaves the file when the defect is fixed, not when it stops being convenient.

**Verification to include in the PR:** the raw per-run results (command + tail of output per run), machine/OS/Python version, and — if re-enabling — a full-suite run showing the test selected and green (`uv run pytest -q -n 4`; see [CONTRIBUTING.md](https://github.com/no-human-ai/no_human/blob/main/CONTRIBUTING.md) for why not `-n auto`). No credential needed.

[...]

PR #102, opened 2026-09-06
Bundled-CLI check: assert the directory and stem, not the POSIX name

Sreekant13:
## What this changes

`test_bundled_cli_skip_marker_is_honest` asserts the located binary ends with `os.path.join("_bundled", "claude")`. The SDK's own locator does not promise that name: `SubprocessCLITransport._find_bundled_cli` looks for `claude.exe` on Windows and `claude` everywhere else, and it does so identically in 0.2.121 and 0.2.152.

So the assertion fails on Windows while the locator is working correctly. No issue for this one; I found it while probing the SDK for #16 and it is unrelated to that port.

The rest of the module is fine on Windows. Before this change it ran **4 passed, 1 failed** there; after it, all 5 pass. The `Python` job is `ubuntu-latest` only, so CI has never seen it.

**Checked as parent directory plus stem, rather than re-deriving the platform rule.** The binary's name belongs to the SDK, and a second copy of that rule in the test is a second thing to drift out of sync. `Path(p).stem` is `"claude"` for both spellings. The check still catches a locator that returned the wrong thing: a system CLI at `/usr/bin/claude` has parent `bin`, not `_bundled`, and still fails.

The skip `reason` carried the same POSIX-only path. That test's own docstring requires the reason to say where the bundle is expected, so it now names both spellings instead of only the one that is wrong half the time.

## Evidence

Same commit, same machine, only the file reverted between the two runs:

```
$ git stash push tests/test_local_backend_loopback.py     # main's version
$ uv run pytest -q tests/test_local_backend_loopback.py
FAILED tests/test_local_backend_loopback.py::test_bundled_cli_skip_marker_is_honest
1 failed, 4 passed in 8.19s

$ git stash pop                                            # with the fix
$ uv run pytest -q tests/test_local_backend_loopback.py
5 passed in 7.77s
```

The failure itself:

```
assert _BUNDLED_CLI_PATH.endswith(os.path.join("_bundled", "claude"))
E   assert False
E    +  where False = '...\\claude_agent_sdk\\_bundled\\claude.exe'.endswith('_bundled\\claude')
```

Gates:

```
$ uv run pytest -q tests/test_structural_budget.py tests/test_ci.py tests/test_readme_claims.py
267 passed, 12 skipped in 84.83s

OK: 1522 file(s) match RELEASE_MANIFEST.txt
```

That both SDK versions use the same name, from their own sources rather than from inference:

```
0.2.121  cli_name = "claude.exe" if platform.system() == "Windows" else "claude"
0.2.152  cli_name = "claude.exe" if platform.system() == "Windows" else "claude"
```

## Checklist

- [x] Tests cover the change, and they fail on the unfixed code. The changed test IS the coverage, and the run above shows it red on main's version and green on this one.
- [x] Test count and assertion count did not go down. Test count is unchanged at 5; the assertion count in that test goes from 2 to 3.

`RELEASE_MANIFEST.txt` re-pinned last.

One note in case you would rather not take this: it is a test-only fix for a platform your CI does not run, so if you consider the POSIX name a deliberate assumption, say so and I will close it. I raised it because the project ships a Windows desktop app and a Windows bundle job, and because a red test on a contributor's machine that is green in CI is the kind of thing that teaches people to ignore red.


Sreekant13:
@eyalgolan thank you for the collaborator invite, and congratulations on getting the MCP port landed in #79. You beat me to that one fairly.

A small process question rather than anything about this change. I pushed this branch from my fork, the same as every PR so far, even though the invite means I could push a branch to the repo itself now. That felt like the sort of thing worth asking about rather than assuming, since it changes what shows up in your branch list and in your notifications.

Would you rather contributors with write push feature branches to the repo directly, or do you prefer the fork flow to stay as it is? I am happy either way, and I will keep using the fork until you say otherwise. Same for the merge button: it is available to me on an approved PR now, and I would rather you kept landing them unless you tell me differently.

No rush on this or on the PR itself.

eyalgolan (approved):
Approving, thanks @Sreekant13 . I read _find_bundled_cli in both SDK versions and it's exactly as you said, and your check keeps what the test was there to prove (a /usr/bin/claude still fails it). Good catch on a platform our CI doesn't run this file on.

On the process question: either is fine. Fork branches keep the repo's branch list quiet, so if you don't mind the extra remote, keep doing that. If you ever push to the repo directly, prefix the branch with your handle so it's obvious whose it is.

On the merge button: yes, please use it. That's what the write access is for. When a PR of yours has my approval, nothing was pushed after it, the branch is up to date and CI is green, press "Squash and merge" yourself. CONTRIBUTING.md has the exact list. This one qualifies as soon as my approval is in.

Sreekant13:
Thank you @eyalgolan, and thank you for reading `_find_bundled_cli` in both versions rather than taking my word for it. Merged with the button.

Thank you also for answering the process question so concretely. Recording what I am taking from it, so you do not have to repeat yourself: I will stay on fork branches, and if I ever push one to the repository directly I will prefix it with my handle. On merging, I will press squash-merge on my own PRs once your approval is in, nothing has been pushed since it, the branch is current with main, and CI is green, and I will leave it to you whenever any of those does not hold.

One correction I owe you, because I had it wrong and said so out loud elsewhere. I had assumed from the history that this repository lands PRs as merge commits, since #85, #89 and #92 all have two parents with you as the author. CONTRIBUTING.md says plainly that squash is the only method the repository allows, and #95 and #79 both landed as single-parent squashes keeping the contributor as author. Those three were you hand-landing outside the button, and I generalised from the artifacts instead of reading the rule that produced them. Corrected on my side, and worth stating here in case it saves someone else the same wrong inference.

On to #19. The serial half is measured, and the `-n 4` half is running now.

Issue #104, opened 2026-09-06
test_the_unfixed_readline_raises_the_exact_asyncio_valueerror can deadlock and hang the whole job to the 30-minute ceiling


Sreekant13:
## What happened

While running the `-n 4` half of the #19 measurement I hit a full-suite run that produced no output for **three hours**. It was not slow, it was deadlocked, and I caught it alive and took a stack before killing it.

The hang is in `tests/test_codex_oversized_jsonl_line.py::test_the_unfixed_readline_raises_the_exact_asyncio_valueerror` (line 138).

## The evidence

Four xdist workers. Three were idle waiting for work. The fourth, after 3h06m:

```
Thread 44186 (idle): "MainThread"
    select (selectors.py:468)
    _run_once (asyncio/base_events.py:1961)
    run_forever (asyncio/base_events.py:645)
    run_until_complete (asyncio/base_events.py:678)
    run (asyncio/runners.py:118)
    inner (pytest_asyncio/plugin.py:905)
    pytest_pyfunc_call (_pytest/python.py:167)
    ...
    run_one_test (xdist/remote.py:227)
```

Identified from the frame locals:

```
kwargs: {"oversized_event_cli":
  "/tmp/pytest-of-root/pytest-148/popen-gw3/test_the_unfixed_readline_rais0/fake-codex"}
```

State of the world at that moment:

```
process table   the fake-codex child is GONE, not even a zombie
/proc/44186/fd  17 -> pipe:[7241321]   (the child's stdout, read end, still open)
pipe holders    pid 44186 only -- NOTHING holds the write end
wchan           do_epoll_wait
threads         2
```

## Why I think it is `await proc.wait()`

The test does, with no timeout on anything:

```python
    first = await proc.stdout.readline()
    ...
        await proc.stdout.readline()
    finally:
        if proc.returncode is None:
            proc.kill()
        await proc.wait()
```

With the child dead and nothing holding the pipe's write end, that pipe is at EOF, so both `readline()` calls would return immediately and `drain()` could not block. By elimination the only await that can block in that state is `proc.wait()`, waiting on a returncode future that nothing will resolve. I am inferring that rather than observing it: py-spy shows OS thread stacks, and a suspended coroutine does not appear in one.

The consequence is the part I am confident about. **There is no timeout anywhere in this test**, so when the await does not resolve the test does not fail, it hangs, and it takes the whole job with it to the 30 minute `timeout-minutes` ceiling.

## What I could NOT reproduce, which matters

[...]

PR #103, opened 2026-09-06
nh test --help: describe mode selectors, not stale counts

ShamikOfficial:
## What this changes

Fixes #10.

`nh test --help` printed hard-coded headcounts (`711` / `721` / `10`) from the `test` command docstring. Those numbers were already wrong against a suite of ~9,400 tests, and would keep rotting.

The help now describes what each mode *selects*, matching `scripts/run_tests.sh`:

- `fast` — `not slow and not nightly` (the PR lane)
- `full` — everything
- `slow` — `slow`

The script-missing fallback in `commands.py` used `not slow` only; it now uses `not slow and not nightly` so it agrees with the shell script. Line count of `cli/commands.py` stays at the frozen 8567.

## Evidence

```
$ uv run pytest -q tests/test_cli_commands.py::test_test_cmd_help tests/test_cli_commands.py::test_test_cmd_help_describes_selectors_not_counts
2 passed in 6.35s

$ git stash push -- src/no_human/cli/commands.py
$ uv run pytest -q tests/test_cli_commands.py::test_test_cmd_help_describes_selectors_not_counts
FAILED — assert 'not slow and not nightly' in help output
1 failed

$ uv run python scripts/check_release_manifest.py
OK: 1523 file(s) match RELEASE_MANIFEST.txt

$ python -c "from pathlib import Path; print(len(Path('src/no_human/cli/commands.py').read_text(encoding='utf-8').splitlines()))"
8567
```

`tests/test_cli_commands.py` gains one test; the existing help smoke test still passes.

## Checklist

- [x] Tests cover the change, and they fail on the unfixed code.
- [x] Test count and assertion count did not go down. `test_cli_commands.py` gains 1 test.
- [x] Lockfiles committed if a dependency changed. (None changed.)
- [x] Nothing in the diff reads or writes a credential inside the repo.
- [x] This does not conflict with the project's standing constraints: the agent never merges (it opens a PR and stops; merge is a human action), review stays evidence-based (independent fresh-context reviewer, never a numeric self-scoring gate), and the locally installed product keeps its lean stack (SQLite only, no extra services, and no re-implementation of what the coding backend's SDK already provides).
- [x] First PR only: I have read `CLA.md` and added my `contributors/shamikofficial.md` entry to this PR. (Ticking this box is not the record — the file is. CI checks for the file.)

`RELEASE_MANIFEST.txt` repinned last, three rows for the files changed plus the new CLA ledger entry.

PR #108, opened 2026-09-06
KI-1 re-measured: select the concurrency test again

### LINKED ISSUE #19: Re-measure the KI-1 concurrent-commit crash now that Store writes are serialized, then re-enable or retire the deselected test

Sreekant13:
## What this changes

Fixes #19. The KI-1 failure rate is re-taken now that `serialized_write` has landed, and the test is selected again.

## The measurement

All on `580a879`, every run with the test selected:

| condition | failures | environment |
| --- | --- | --- |
| this test alone, serial | **0 / 300** | Linux, matched to the `Python` job |
| this test alone, serial | **0 / 100** | Windows 11 host |
| whole suite, `-n 4` | **0 / 13** | Linux, matched to the `Python` job |

Against the 3/8 the entry recorded, the probability of 300 consecutive passes is `0.625^300`, about `6e-62`.

**Why 300 and not the 10 the issue asked for.** Zero failures in 10 bounds the true rate near 30% under the rule of three, which cannot tell "fixed" apart from "fails one run in twenty" — and the second one comes back as CI flake after you have merged. At 300 the bound is near **1%**. That is the number I am willing to put behind a re-enable.

**Why the Linux environment is built the way it is.** KI-1 is a SQLite defect and SQLite ships with the interpreter, so the three candidates were not the same build:

```
python:3.12-slim (Debian's python)         SQLite 3.46.1
Windows host                               SQLite 3.49.1
ubuntu:24.04 + uv-managed python  <- CI    SQLite 3.53.1
```

Your job uses `astral-sh/setup-uv` with `.python-version` and `uv sync --frozen`, so it runs uv's managed CPython. The container does the same, and clones onto its own overlayfs rather than a bind mount from Windows, because a 9p/virtiofs mount has different fsync and locking semantics and I would have been measuring the mount.

## The issue understated the scope, and one part of that is risky

The deselect is in **four** places, not the one the issue names, and three of them assert a rate the measurement contradicts:

| file | what was there |
| --- | --- |
| `.github/workflows/ci.yml:258` | the deselect the issue names |
| `.github/workflows/ci.yml:225-235` | comment: "in roughly a third of runs" |
| `scripts/run_tests.sh:65` | **a second deselect**, nightly lane |
| `CONTRIBUTING.md:186` | "a real, open defect ... about a third of runs die" |

**The `run_tests.sh` half is the risky one and it is a separate commit so you can drop it alone.** `scripts/nightly_eval.sh:138` runs that lane and captures its exit code, and its header line 30 says "Exit code IS the verdict". So unlike CI, where a flake costs one red job, a flake there reddens the nightly verdict and masks the eval signal — which is exactly why the deselect was put there. My bound is near 1%, not zero, and the runs were on WSL2. If you would rather let `main` accumulate history first, `git revert` the second commit and keep the rest.

## A test so this cannot drift again

The two deselect lists were kept in step by hand and nothing checked them. That is not hypothetical: this issue named only `ci.yml`. `tests/test_deselect_lists_agree.py` now pins them to each other, with a control so a broken extractor cannot make the equality pass vacuously.

Worth separating: this pins the **deselect lists**, not the marker **selectors**. `ci.yml:247` and `CONTRIBUTING.md:164` both say `tests/test_test_lanes.py` pins the selectors against each other, and that file is not in the tree. I have raised that separately rather than widen this PR.

## Read this before merging: your PR check will not exercise this test

The test is `@pytest.mark.slow`, and the `pull_request` lane runs `-m "not slow and not nightly"`. So it is **not selected on this PR**, and a green check here says nothing about the change. The lanes that do select it are push-to-main and `workflow_dispatch`.

That is why I pushed this branch to the repository rather than my fork, prefixed with my handle as you asked: `workflow_dispatch` needs the ref to exist here, so you can run the nightly lane on the branch before merging rather than finding out on `main`.

## Evidence

```
    tests/test_readme_claims.py tests/test_derived_conflict_inventory.py \
    tests/test_deselect_lists_agree.py tests/test_ci_network_step_bounds.py
284 passed, 12 skipped in 134.99s

OK: 1523 file(s) match RELEASE_MANIFEST.txt
```

I cannot show you a green **full** suite locally: on this Windows host the suite has 25 environmental failures (no node for the JS suites, no tzdata, running as root, the bundled CLI cannot spawn), and they are the same 25 every run. Your own push-to-main run on `main` is green with the full suite at `-n 4`, which is the evidence that those 25 are mine.

## One thing that happened during the measurement, unrelated to KI-1

The first `-n 4` batch stalled for three hours on a deadlocked test, which is #104. I restarted with a watchdog that dumps stacks and kills, so a hang became data rather than a stalled harness. Hung attempts produce no verdict for the KI-1 test, so they are **excluded** rather than counted as passes: 13 completed runs of 15 attempts, 2 hung. Every completed run had `KI1=PASS`.

## Checklist

- [x] Tests cover the change, and they fail on the unfixed code. `test_deselect_lists_agree.py` fails if either list moves without the other; verified by reverting the `run_tests.sh` commit locally.
- [x] Test count and assertion count did not go down. Two tests added; one test re-enabled in two lanes.


Sreekant13:
Merged `main` in after #103 landed, so the branch is current.

One thing worth surfacing rather than leaving implicit: that merge touched `RELEASE_MANIFEST.txt` and git auto-resolved it without a conflict. A clean auto-merge of a content-hash file is not the same as a correct one, since git only knows the lines do not overlap and nothing checks that the hashes still describe the tree. So I re-ran the gate rather than trusting the absence of a conflict:

```

$ uv run pytest -q tests/test_deselect_lists_agree.py tests/test_ci.py \
    tests/test_readme_claims.py tests/test_structural_budget.py
269 passed, 12 skipped in 79.58s
```

The reminder about the PR lane still stands: this test is `@pytest.mark.slow`, so a green check here does not exercise it. The branch is on the repository as `sreekant13/ki1-remeasure-and-reenable` specifically so you can `workflow_dispatch` the nightly lane against it if you want that signal before merging.

eyalgolan (changes requested):
Thanks @Sreekant13 . This is the measurement I wanted, with the limits written next to the numbers, and the deselect-list test is the right thing to leave behind. I checked what I could from here: the lock is where you say, nightly_eval really does make the lane's exit code its verdict, the lanes test that ci.yml and CONTRIBUTING cite is indeed missing, your new test goes red when I drop one deselect from run_tests.sh, and I ran the KI-1 test myself 100 times serially on macOS with the same SQLite 3.53.1 you measured against: 100/100.

One thing to fix before it lands. run_tests.sh line 58 says "10 whole-suite -n 4 runs"; ci.yml, CONTRIBUTING and KNOWN_ISSUES say 13. One number in four places, and drift between copies is exactly what your new test is about, so let's not ship it with one.

On the nightly lane: I'm going to take your own advice and revert the second commit for now. The CI re-enable already closes #19, and I'd rather see a couple of weeks of push-to-main history before a residual one-in-a-hundred can redden the nightly verdict. When main has run it clean for a while, re-selecting it there is a one-line PR and it's yours if you want it. If you'd rather do the revert yourself so the branch stays entirely yours, say so; otherwise I'll push a revert of it onto the branch before merging.

Fix the number, push, and I'll approve.

Sreekant13:
Both fixed, and thank you for running it 100 times on macOS against the same SQLite build. That is a platform and a filesystem I could not reach, and it is the one caveat in KNOWN_ISSUES I most wanted covered by someone else's machine rather than by my own hedging.

## The drifted number

You are right, and it is the exact failure this PR adds a test for. I updated three copies with a search and replace and the fourth was phrased differently, so it survived. That is drift between copies, in the PR about drift between copies.

All copies now read 13. There are **five**, not four: `run_tests.sh`, `ci.yml`, `CONTRIBUTING.md`, `KNOWN_ISSUES.md`, and the exception entry in the test itself, which quotes the figure in its reason string.

## The nightly lane

Taking your decision, and doing the change here so the branch stays mine, since you offered.

**A bare revert of that commit does not work, and I checked rather than assumed:**

- **It turns the branch red.** `tests/test_deselect_lists_agree.py` pins the two lists to each other. Holding the test back in one lane *is* a difference between them. You verified that test goes red when you drop one deselect from `run_tests.sh`; the same property makes the revert red.
- **It reinstates a claim this PR disproves.** The original comment says KI-1 "fails ... in roughly a third of runs".

So instead the deselect stays in the nightly lane, and its comment now records the measurement, says why *this* lane is the one that waits, and says what removes it:

> REMOVE THIS DESELECT once main has run it clean for a while; that is a one-line change here plus deleting the entry from the test above.

## The test knows about the divergence rather than being weakened by it

`_NIGHTLY_ONLY` holds the one entry with its reason. Four properties, not one:

- an **undocumented** nightly-only deselect still fails
- a difference in the **other** direction (ci.yml deselecting something the nightly lane runs) fails with **no exception mechanism at all**, because that one is a local red CI can never reproduce
- a **stale** entry fails: if the deselect goes and the entry stays, `test_every_documented_exception_is_still_needed` names it
- the control still fails a deselect pointing at a file that does not exist

An intended difference and an accidental one look identical in a diff. Only one of them has an entry here.

`KNOWN_ISSUES.md` and `CONTRIBUTING.md` no longer say the test is selected in both places, because it is not.

## Evidence

```
$ uv run pytest -q tests/test_deselect_lists_agree.py tests/test_ci.py \
    tests/test_readme_claims.py tests/test_structural_budget.py
270 passed, 12 skipped in 83.56s

$ bash -n scripts/run_tests.sh
(syntax OK)
```

Yes to the follow-up PR re-selecting it in the nightly lane once `main` has clean history. The comment and the test entry both point at it, so it will not need rediscovering.

eyalgolan (approved):
@Sreekant13 That's better than the revert I asked for, and you were right that the bare revert would have gone red on your own test and put the "a third of runs" sentence back. An exception list that names the divergence, refuses the other direction, and fails when the entry outlives the deselect is the right shape. I went at it from all four sides on my side: an undocumented nightly-only deselect, the deselect gone with the entry kept, a ci.yml-only deselect, and a bogus path; each one turns exactly the test you'd expect red. All five copies of the number say 13 now, bash -n is clean, manifest matches.

Approving. This one's yours to merge: nothing was pushed after my approval, it's up to date, and CI is green, so press the button. And yes to the follow-up that re-selects it in the nightly lane once main has some history.

Sreekant13:
Merged, thank you. Landed as `aab935b6`, and #19 closed with it.

Recording the follow-up so it does not get lost: re-selecting the test in `run_tests.sh`'s nightly lane once `main` has run it clean for a while. Both the comment in that file and the `_NIGHTLY_ONLY` entry in the test point at it, and `test_every_documented_exception_is_still_needed` will fail if someone removes the deselect without removing the note, so the two cannot drift apart in the meantime.

I will leave it a couple of weeks of push history as you suggested rather than pick a date, and open it as the one-line PR then.

Issue #111, opened 2026-09-06
CONTRIBUTING's test counts are stale (2,980 vs 11,710; 538 vs 1,576) and the documented web command fails on a fresh clone


Sreekant13:
`CONTRIBUTING.md:110` states two figures about the Python suite. Both are stale, and one is out by about 4x.

> About 2,980 tests. On a 4-core machine `-n 4` brings a full run to roughly four minutes

## The count

```
$ uv run pytest --collect-only -q | tail -1
11710 tests collected in 7.15s
```

2,980 documented against 11,710 collected.

## The timing

Your own `Python` job is the cleanest available 4-core `-n 4` measurement, on the push-to-main lane where no marker filter applies, so it runs the whole suite:

```
run 33949024684, job "Python", step "Run tests"
  started   2026-09-05T06:09:24Z
  completed 2026-09-05T06:16:59Z
  = 7m35s, for the step alone, excluding checkout and uv sync
```

"Roughly four minutes" against 7m35s on exactly the hardware shape the sentence names.

## This is the same defect as #10, one file over

#10 was `nh test --help` freezing "721 tests", and #103 fixed it by describing what each mode *selects* instead of hardcoding a count:

```
fast  — selects 'not slow and not nightly' (the PR lane)
full  — everything
slow  — selects 'slow'
```

That is exactly the right shape, and CONTRIBUTING.md is the other copy of the same claim. It was not in #10's scope, so it was not touched.

Worth noting that #10's own text said "the suite has ~9,400 tests" and it is 11,710 today, which is the argument for the #103 approach rather than a fresher number: any count committed to a file starts going stale the same day.

## Suggested shape

Drop both figures and point at the command that answers the question, in the same spirit as #103:

```bash
uv run pytest --collect-only -q | tail -1   # how many tests there are today
```

If a rough sense of scale is wanted in prose, "over 11,000" ages better than a precise number, and the timing claim could name your CI job (which is public and dated) rather than an unattributed four minutes.

## The web section has the same two problems

`CONTRIBUTING.md:216-222` documents:

```bash
cd web
npm ci
npm test
```

> 538 tests.

Measured on a fresh clone:

```
# tests 1576
# pass  1570
# fail  6
```

So the count is out by ~3x, and **the documented command does not pass on a fresh clone**, on any OS. Two of the six failures are `cancelFlow.test.mjs` and `eventLabels.test.mjs` failing closed with:

> `web/dist/assets is missing - run npm run build in web/ first`

[...]

PR #113, opened 2026-09-06
Board tests: import file:// URLs, not raw absolute paths

### LINKED ISSUE #112: Board tests dynamically import an absolute path, which fails on Windows with ERR_UNSUPPORTED_ESM_URL_SCHEME
Two board tests do a dynamic `import()` of an absolute filesystem path. On Windows a drive letter parses as a URL scheme, so Node's ESM loader refuses it and the tests fail before asserting anything.

```
$ cd web && npm ci && npm test
# tests 1576
# pass  1570
# fail  6
```

Four of those six are this:

```
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data,
and node are supported by the default ESM loader. On Windows, absolute paths
must be valid file:// URLs. Received protocol 'e:'
```

`e:` is the drive letter of my checkout.

## Where

`src/viteProxyTarget.test.mjs` builds an absolute path and imports it three times (lines 20, 33, 45):

```js
const CONFIG_PATH = join(dirname(fileURLToPath(import.meta.url)), "..", "vite.config.js");
...
const mod = await import(`${CONFIG_PATH}?case=default`);
```

`src/markdownSafety.test.mjs:44` does the same with a generated shim, and takes the whole file down with it:

```js
const file = join(dir, `markdownSafety-${process.pid}.mjs`);
...
return (await import(`${file}?t=${Date.now()}`)).default;
```

On POSIX the path starts with `/` and happens to resolve; on Windows it starts with `E:\`, and `E:` is read as the scheme.

## The fix, verified

`node:url` exports `pathToFileURL` for exactly this, and both files already import `fileURLToPath` from it, so it is one more name on an existing import.

I patched `viteProxyTarget.test.mjs` locally to confirm rather than assume:

```
before:  # tests 3   # pass 0   # fail 3
after:   # tests 3   # pass 3   # fail 0
```

The change was `import { fileURLToPath, pathToFileURL } from "node:url";` and `await import(`${pathToFileURL(CONFIG_PATH).href}?case=default`)`. I have reverted it; happy to send it as a PR covering both files if you want it.

Note `markdownSafety.test.mjs` also rewrites import specifiers into generated source at line 38 using `join(here, rel)`, which is the same shape and probably needs the same treatment, but I did not get far enough to see it fail because the outer import dies first.

## Why CI has never seen it

The `Web board` job is `ubuntu-latest`, like the `Python` job. Nothing runs this suite on Windows, and the project ships a Windows desktop app.

## Not the same as the other two failures

The remaining two (`cancelFlow.test.mjs`, `eventLabels.test.mjs`) fail on purpose with `web/dist/assets is missing — run npm run build in web/ first`, which is a deliberate fail-closed and correct. That one is a documentation gap and I have raised it separately.

Found while checking the test counts in CONTRIBUTING.md against reality.

Sreekant13:
## What this changes

Fixes #112.

I offered this in the issue and you have not answered yet, so this is speculative. It is a two-file test-only change and easy to close if you would rather do it differently.

On Windows an absolute path starts `E:\`, and Node's ESM loader reads the drive letter as a URL scheme, so a dynamic `import()` of one dies before the test asserts anything:

```
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data,
and node are supported by the default ESM loader. On Windows, absolute paths
must be valid file:// URLs. Received protocol 'e:'
```

A POSIX path happens to work because it starts with `/`, which is why the ubuntu-only `Web board` job has never seen it.

`pathToFileURL` is the standard-library answer, and both files already import `fileURLToPath` from `node:url`, so it is one more name on an existing import.

## Two files, and the second needed it twice over

`viteProxyTarget.test.mjs` imports `vite.config.js` three times by path.

`markdownSafety.test.mjs` imports a generated shim, which took the **whole file** down rather than one test. It also writes absolute paths into that generated ESM as import specifiers, and those needed fixing for a second reason as well as the first: a Windows path inside an emitted JS string literal has its backslashes read as escape sequences.

## Evidence

`cd web && npm test`, on Windows, same clone:

```
before   # tests 1576   # pass 1570   # fail 6
after    # tests 1584   # pass 1582   # fail 2
```

The two files alone:

```
$ node --test src/viteProxyTarget.test.mjs src/markdownSafety.test.mjs
# tests 12
# pass 12
# fail 0
```

**The count rises, which is the part I would look at hardest.** `markdownSafety.test.mjs` previously died at load, so its tests never ran on Windows at all. They were invisible, not passing — the same shape as the concern in `ci.yml`'s own comment, that a node running in neither lane looks exactly like a node that passes.

The two remaining failures are `cancelFlow`/`eventLabels` failing closed on a missing `web/dist`, which is deliberate and correct. That is the documentation gap in #111, not this.

```
```

## What I have NOT verified, and how this PR verifies it

**I have not run these two files on Linux.** `pathToFileURL` is cross-platform and produces `file:///...` on POSIX too, so I expect no change there, but I did not measure it and I am not going to imply I did.

This PR's own `Web board` job is exactly that check, and unlike the `Python` lane on my last PR it does select these tests, so a green check here is real evidence rather than a formality.

## Also in the diff

A duplicated comment line in `markdownSafety.test.mjs`: the same sentence appeared twice, once without its trailing semicolon. It is two lines from the code being changed, so I removed it rather than leave it for a separate PR. Say the word if you would rather it were not bundled.

## Checklist

- [x] Tests cover the change, and they fail on the unfixed code. The changed tests ARE the coverage; before/after counts above are on the same clone.
- [x] Test count and assertion count did not go down. It rises by 8, because a file that could not load now loads.


Sreekant13:
The Linux gap I flagged is now closed by the PR's own run rather than by my expectation: **`Web board` passed on ubuntu-latest in 28s**, and that job does select both changed files, unlike the `Python` lane on #108.

So the change is measured on both sides now: Windows went 6 failures to 2 (the two remaining being the deliberate `web/dist` fail-closed, which is #111), and Linux is unchanged and green.

eyalgolan (approved):
Approving, thanks @Sreekant13 for another great PR. Checked it from both sides here: the two files and the whole board suite pass on macOS with your version and with main's, and a raw import of an E:\ path throws ERR_UNSUPPORTED_ESM_URL_SCHEME even on this Mac, so the loader rule is what you said it is. The markdownSafety rewrite going through pathToFileURL too was the right call; a half-fixed file would have been worse than none. Yours to merge.

PR #115, opened 2026-09-07
Bound the awaits in the oversized-JSONL characterization test

### LINKED ISSUE #104: test_the_unfixed_readline_raises_the_exact_asyncio_valueerror can deadlock and hang the whole job to the 30-minute ceiling

Sreekant13:
## What this changes

Fixes #104.

This was the only async test in the file with no timeout on its awaits. Unbounded, it can deadlock a whole `-n 4` run to CI's 30-minute ceiling. As you said, it is the odd one out, so this matches the file's existing idiom rather than inventing a pattern: `asyncio.wait_for(..., 30)`, already used at lines 115, 197, 228, 279 and 338.

## What the deadlock looked like

Caught live with py-spy after 3h06m:

```
three xdist workers  idle in execnet, waiting for work
the fourth           select() inside an asyncio loop, via pytest_asyncio
the fake-codex child GONE from the process table, not even a zombie
its stdout pipe      still open in the worker, NOTHING holding the write end
/proc/<pid>/wchan    do_epoll_wait
```

With the child dead and the pipe at EOF, both `readline()` calls return immediately and `drain()` cannot block, so by elimination the await that hangs is `proc.wait()`, on a returncode future nothing resolves.

**That last step is inference, not observation** — py-spy shows OS thread stacks and a suspended coroutine is not in one. The consequence is not inference, and it is what this fixes: with no bound the test does not fail, it hangs, and it takes the job with it.

## Evidence

Verified on Linux. The whole file fails on Windows for unrelated pre-existing reasons, and I checked that against unmodified `main` rather than assuming it was mine: **11 failures there too**.

```
upstream main, whole file        11 passed
with this change, whole file     11 passed
```

And that the bound actually **fires**, with the fake CLI patched to sleep instead of emitting:

```
with this change    fails in 35s, asyncio TimeoutError, traceback attached
upstream main       rc=124 -- killed by `timeout 120`, still running
```

That is the before/after: a named failure with a traceback, against a process that had to be killed.

```
```

## What this does not do

It does not diagnose the race, and is not meant to. Rate measured on `580a879` during #19: **two hangs in fifteen** whole-suite `-n 4` runs. Not reproducible in isolation — 75 targeted runs of the test alone, the file serially, and the file under `-n 4` produced **zero** hangs, so I still cannot hand you a cheap recipe.

What changes is that the next occurrence leaves a traceback instead of a silent half hour, which is what makes the root cause chaseable at all.

## Checklist

- [x] Tests cover the change, and they fail on the unfixed code. The bound is proved to fire against a hanging child, and `main` is proved to hang on the same input.
- [x] Test count and assertion count did not go down. Unchanged at 11.


eyalgolan (approved):
Approving, thanks @Sreekant13 . I did the same experiment you did, on my side: made the fake CLI sleep before emitting, and with your change the test fails in 30 seconds with a TimeoutError and a traceback, while main's version was still sitting there when I killed it at 60. That's exactly the trade we wanted. Matching the file's own wait_for idiom was the right call. Yours to merge.

Issue #105, opened 2026-09-06
Command guards tokenise with POSIX shlex, so Windows path separators are deleted and the venv guard fails open


Sreekant13:
Both command guards tokenise with `shlex` in POSIX mode, where `\` is an escape character. On Windows that silently deletes every path separator, so a native path can never resolve, and the venv guard's one allow-and-log fallback stops being an edge case and becomes the default.

## Reproduction

```python
from no_human.agent.venv_install_guard import denial_reason, _lex
env = {"PATH": ""}

_lex(r"C:\Users\me\proj\.venv\Scripts\pip.exe install requests")
# ['C:Usersmeproj.venvScriptspip.exe', 'install', 'requests']

denial_reason(r"C:\Users\me\proj\.venv\Scripts\pip.exe install requests",
              cwd=r"C:\Users\me\proj", env=env)
# None        <- allowed
```

`guard.py` does the same, via `shlex.split` at `:438` and `shlex.shlex(cmd, posix=True, ...)` at `:486` and `:885`:

```python
from no_human.agent.guard import _segment_tokens
_segment_tokens(r"rm -rf C:\Users\me\important")
# [['rm', '-rf', 'C:Usersmeimportant']]
```

## Why it fails open rather than closed

`_resolve_installer` treats an unresolvable installer path as "structure genuinely cannot decide" and allows, which is documented and reasonable when the path is merely absent. On Windows the path is *destroyed before resolution is attempted*, so it can never resolve, and the fallback fires every time.

The pure-Windows form is the quieter of the two, which is the part I would want fixed first:

| command | lexed argv[0] | outcome |
| --- | --- | --- |
| `C:\...\.venv\Scripts\pip.exe install x` | `C:...venvScriptspip.exe` | allowed, **no warning at all** |
| `C:\...\proj/.venv/bin/pip install x` | `C:...proj/.venv/bin/pip` | allowed, warning logged |

The silent one is silent because after mangling there is no `/` in the token, so `_resolve_installer` takes the `shutil.which` branch, and `_is_installer_name("C:Usersmeproj.venvScriptspip.exe")` is False, so not even the WARNING fires. A guard that cannot fire looks exactly like a guard that passed.

Neither module has any platform handling: no `os.name`, no `sys.platform`, no `ntpath`.

## Observed effect on the suite

`tests/test_guard.py` has **15 failures on Windows**, and every one is a "must be refused" assertion, i.e. the guard allowing what it should deny:

```
[...]

Issue #107, opened 2026-09-06
guard.py does not strip .exe when matching installer names, so v1 misses 'pip.exe install'


Sreekant13:
`venv_install_guard.py` strips a `.exe` suffix before deciding whether a name is an installer. `guard.py` never does. So the two layers disagree about Windows-shaped argv[0], and the v1 lexical layer misses installs the v2 resolution layer would catch.

## Reproduction

```python
from no_human.agent.guard import _is_pkg_manager_name, _pkg_install_match
from no_human.agent.venv_install_guard import _is_installer_name, _basename

name              guard._is_pkg_manager_name    venv_guard._is_installer_name(_basename)
pip               True                          True
pip.exe           False                         True     <- disagree
uv.exe            False                         True     <- disagree
python.exe        False                         True     <- disagree
python3.12.exe    False                         True     <- disagree
pip3.exe          False                         True     <- disagree

_pkg_install_match(["pip", "install", "foo"])                 -> ['foo']
_pkg_install_match(["pip.exe", "install", "foo"])             -> None
_pkg_install_match(["python.exe", "-m", "pip", "install", "foo"]) -> None
```

`venv_install_guard._basename` does this and `guard.py` has no equivalent:

```python
def _basename(path: str) -> str:
    name = os.path.basename(path)
    if name.lower().endswith(".exe"):
        name = name[:-4]
    return name
```

## Why this is worth its own issue

It needs no path and no Windows-specific string. A bare `pip.exe install foo` is missed by v1 on any platform, so this is independent of the POSIX-`shlex` issue I filed separately, and independent of which shell the coder uses.

On its own it may be harmless: your comment at `_MAX_INTERPRETER_PEEL` says a v1 miss "falls through to `None` here — an ordinary v1 miss, not a crash — leaving v2's resolution-based check as the backstop". That is exactly right on Linux.

The reason I am raising it anyway is that on Windows **v2 is the layer the shlex issue disables**. So for `C:\proj\.venv\Scripts\pip.exe install foo` both layers miss for different reasons: v1 because argv[0] ends in `.exe`, v2 because the path lost its separators. Defence in depth stops being in depth when both layers fail on the same input for unrelated reasons.

## Suggested shape

Give `guard.py` the same suffix strip `venv_install_guard._basename` already has, so the two layers agree on what an installer is called. That is a small change and testable without any platform-specific setup, but it touches a security path, so I would rather you decided the shape than have me guess at it.

Found while running the suite on Windows for #19.

[...]

PR #116, opened 2026-09-07
Treat a Windows volume or share root as a filesystem-wide scan target

### LINKED ISSUE #106: root_scan_denial does not treat a Windows drive root as a filesystem root (C:/ is allowed)
`root_scan_denial` blocks a filesystem-wide scan rooted at `/`, but a Windows drive root is not recognised, so the breaker does not fire.

## Reproduction

```python
from no_human.agent.guard import root_scan_denial

root_scan_denial("grep -r secret /", None)
# "filesystem-wide scan blocked: grep -r secret /. Codebase exploration must be
#  repo-scoped ..."

root_scan_denial(r"grep -r secret C:\ ", None)   # -> None
root_scan_denial("grep -r secret C:/", None)     # -> None
```

## This is not the shlex issue

I filed a separate issue about POSIX `shlex` eating backslashes on Windows, and it would be easy to assume this is a symptom of it. It is not: **`C:/` contains no backslash**, survives lexing intact as `C:/`, and is still allowed. So this is a genuine gap in what counts as a filesystem root, independent of tokenising, and it would remain after the shlex issue is fixed.

The same presumably applies to any `<drive>:\`, `<drive>:/`, and to a UNC root `\\server\share`, though I only measured `C:` in both separator forms.

## Why it matters at all on a Linux-only CI

The rationale in the denial message is about cost and scope, not just correctness: "Scanning / reads the whole machine, burns your turn budget, and answers nothing about this task." All of that is equally true of `C:\` on a Windows box, and the project ships a Windows desktop app.

## Suggested shape, if you want it

Treat a bare drive specifier as a root alongside `/`: `C:`, `C:\`, `C:/` (and the same for any letter). The `C:` case matters because after POSIX lexing that is what `C:\` reduces to, so handling it also closes the backslash form without depending on the other fix landing first.

Happy to send this one as a small PR with tests if you would like it, since unlike the shlex issue it does not depend on knowing which shell the coder uses on Windows. Say the word and I will, or close this if a Windows-rooted scan is out of scope.

Found while running the suite on Windows for #19.

Sreekant13:
## What this changes

Fixes #106. Both separator forms and UNC, as you asked.

```
blocked    C:   C:\   C:/   c:\*   C:/*   D:
           \\server\share   //server/share   and their trailing forms
allowed    C:\repo   C:/repo/src   \\server\share\proj   //server/share/proj
           and every POSIX case, unchanged
```

The bare `C:` form is not padding: `guard.py` tokenises with POSIX shlex, which eats the backslash, so a `C:\` operand arrives already reduced to `C:` (#105). Matching it means this works whether or not that separate defect is fixed.

## Why a new module rather than a helper in guard.py

`agent/guard.py` is frozen at **2892 lines** by `FROZEN_FILE_LINES`, a budget that only ratchets down, and it sits **exactly on the line**. My first attempt put the helper there and the gate failed at `+30`.

So the logic is in `agent/fs_roots.py`, and `guard.py` gains **zero lines**:

```
guard.py 2892 -> 2892 lines (delta 0)
```

All three edits there are in place: the existing `from . import venv_install_guard` grew a second name, the non-slash early return now answers the Windows question instead of assuming repo scope, and the POSIX root test also answers for a forward-slash UNC root, which does start with `/` and so never reaches the other branch.

If you would rather this lived in `guard.py` and the budget moved by a line or two, say so and I will do that instead. I did not want to raise a ratchet on my own initiative.

## Scope I deliberately did not take

Roots only. `C:\Users` is the Windows analogue of `/Users` in `_SYSTEM_ROOTS`, but a Windows system-root list needs the same cwd-before-roots ordering care your comment there explains, and getting it wrong denies every ordinary repo-scoped scan on the platform. That is wider than this issue.

## One gap this cannot close, pinned as a test rather than left implicit

`root_scan_denial` parses each segment with `shlex.split`, and a trailing lone backslash raises:

```
'grep -r secret C:\'   -> ValueError: No escaped character
```

which the loop answers with `continue`. The segment is skipped and the command allowed **before any operand is examined**, so `grep -r secret C:\` is still allowed for a reason that has nothing to do with roots. The escaped and slash forms of the same target *are* denied:

```
'grep -r secret C:\\'  -> denied
'grep -r secret C:/'   -> denied
```

`test_a_trailing_lone_backslash_is_skipped_before_any_operand_check` pins that so it stays visible rather than being assumed fixed by this change. Deciding what an unparseable command should do is a wider security call, so I have raised it on the #105 thread rather than change fail-open semantics here.

## Evidence

```
$ uv run pytest -q tests/test_guard.py -k "windows or posix_root or posix_exempt or trailing_lone"
26 passed, 204 deselected

# against UNMODIFIED guard.py, the same selection:
12 failed, 14 passed

$ uv run pytest -q tests/test_guard.py
15 failed, 215 passed      # the same 15 as main: pre-existing Windows, #105

    tests/test_readme_claims.py tests/test_intake_repo_scope.py \
    tests/test_derived_conflict_inventory.py
290 passed, 12 skipped

```

## Checklist

- [x] Tests cover the change, and they fail on the unfixed code. 12 of the 26 fail against unmodified `guard.py`.
- [x] Test count and assertion count did not go down. 26 added.


eyalgolan (changes requested):
Thanks @Sreekant13 , and the fs_roots split to keep guard.py on its budget was the right instinct; I'd rather that than raise the ratchet. The Windows side is right and I checked the roots and the inside-a-share cases myself.

One thing I'd like changed before it lands, and it's a POSIX side effect rather than anything about Windows. The UNC pattern matches any path with a doubled leading slash and exactly two components, and on Linux and macOS those are ordinary paths. On your branch:

```
//tmp/x       blocked   (main: allowed, /tmp is an exempt prefix)
//home/dev    blocked   (main: allowed)
```

A doubled slash is rare but it's a legal spelling and the guard shouldn't start denying it on a platform where a UNC root doesn't exist. guard.py already has `_IS_WINDOWS` at the top, kept as a constant so a test can flip it: use it to gate the forward-slash UNC form, keep the drive-letter and backslash forms as they are (a `C:` operand can't mean anything else on any platform), and add `//tmp/x` to the "still allowed on POSIX" cases. The pinned trailing-backslash gap stays where you put it; that one is #105's.

One more thing you can't see from your branch: #117 also appends to tests/test_guard.py, so whichever of the two lands second will need a quick merge of main. I'll take them in the order they're ready.

Sreekant13:
You are right, and it is worse than a spelling quibble: it inverted an exemption.

`/tmp` is in `_SCAN_EXEMPT_PREFIXES` precisely so a scan there is allowed, and my version denied `//tmp/x`. An exemption that can be spelled around is worse than no exemption at all.

Measured in a worktree at `origin/main` rather than inferred, because my first attempt to check this stashed only uncommitted work and so measured my own branch:

| operand | main | first version | now |
| --- | --- | --- | --- |
| `//tmp/x` | allowed | **BLOCKED** | allowed |
| `//home/dev` | allowed | **BLOCKED** | allowed |
| `/tmp/x` | allowed | allowed | allowed |
| `/home/dev` | blocked | blocked | blocked |
| `C:/` | allowed | blocked | blocked |

## What changed

The UNC pattern is now two, split on exactly the line you drew:

```
_UNC_BACKSLASH_ROOT_RE   \\server\share    NOT gated
_UNC_SLASH_ROOT_RE       //server/share    gated on is_windows
```

A leading pair of backslashes is not an ordinary POSIX path spelling, so that form cannot mean anything else anywhere. Drive specifiers stay ungated for the same reason: a `C:` operand cannot mean anything else on any platform.

`is_windows_filesystem_root` now takes `is_windows` as a keyword argument and `guard.py` passes `_IS_WINDOWS`. Both call sites are still edited in place, so **guard.py 2892 -> 2892, delta 0**.

## Tests

The forward-slash forms moved out of the unconditional lists into platform-explicit ones, and four tests were added:

- `//tmp/x`, `//home/dev`, `//server/share` allowed with `is_windows=False`
- the same spellings denied with `is_windows=True` (on Windows the first component is a host, not a directory, and that includes `tmp`)
- drive and backslash forms asserted under **both** values, pinning that they are deliberately ungated
- an end-to-end test that monkeypatches `guard._IS_WINDOWS` in both directions

That last one is what makes the POSIX behaviour testable from a Windows machine, which is exactly why you keep that as a constant. Thank you for pointing at it; I had not spotted it.

34 in that selection, all passing. `tests/test_guard.py` still reports the same 15 pre-existing Windows failures as main and no new ones. The pinned trailing-backslash gap is untouched.

## On the overlap with #117

Noted, and thank you for the warning. Both PRs append to `tests/test_guard.py`, so whichever lands second needs a merge of main. I will not pre-merge either, since that would just create the conflict earlier and force a re-approval; tell me which you take first and I will merge main into the other within a few minutes of it landing.

Sreekant13:
Merged `main` and resolved the `tests/test_guard.py` conflict you predicted. #117 landed first, so this was the one that needed care.

Two hunks, and **neither side was chosen over the other**:

- the import line: `fs_roots, guard` versus `guard, venv_install_guard` became the union, `from no_human.agent import fs_roots, guard, venv_install_guard`
- the appended sections: #106's "Windows filesystem roots" and #107's ".exe-suffixed installer names" both landed at the end of the file, so git saw one overlapping hunk. **Both blocks are kept**, in order, with the section separator restored between them.

Verified that nothing was dropped rather than trusting the merge:

```
$ uv run pytest -q tests/test_guard.py -k "<both sections>"
57 passed, 204 deselected        # 34 from #106 + 23 from #107

$ uv run pytest -q tests/test_guard.py
15 failed, 246 passed            # the same 15 pre-existing Windows failures (#105)
```

246 is exactly 223 + 23, so both sets are present and running.

`RELEASE_MANIFEST.txt` also conflicted, since both PRs re-pinned rows. Regenerated rather than hand-merged, because a hand-merged content-hash file can be internally consistent and still wrong: `--strict` reports OK on 1526 files.

This still needs your look, since the platform gate was the change you requested.

eyalgolan (approved):
Approving, thanks @Sreekant13 . Splitting the UNC pattern in two and gating only the slash form on the constant is exactly the shape I had in mind, and keeping the backslash and drive forms ungated is right; there's nothing else those can mean.

I went at it on my side: on this Mac `//tmp/x` and `//home/dev` are allowed again, they're denied only when I flip `_IS_WINDOWS`, and if I take the gate back out your four new tests go red. The merge is clean too: both test blocks are there exactly once, the import line is the union, and guard.py differs from main by exactly your three lines, still on 2892. Yours to merge.

PR #117, opened 2026-09-07
Strip .exe when matching installer names, so v1 agrees with v2

### LINKED ISSUE #107: guard.py does not strip .exe when matching installer names, so v1 misses 'pip.exe install'

Sreekant13:
## What this changes

Fixes #107.

The name predicate was only half of it, and the other half is the part that actually mattered. `_pkg_install_match` computed `PurePosixPath(argv[0]).name` and compared it against bare strings (`"pip"`, `"pip3"`, `"uv"`, …), bypassing `_is_pkg_manager_name` entirely, so `pip.exe install foo` matched **nothing at all**:

```
before   ['pip.exe', 'install', 'foo']                 -> None
         ['python.exe', '-m', 'pip', 'install', 'foo'] -> None
after    ['pip.exe', 'install', 'foo']                 -> ['foo']
         ['python.exe', '-m', 'pip', 'install', 'foo'] -> ['foo']
```

Three sites now read the name through `venv_install_guard._basename`: the predicate, the interpreter peel, and the verb table. That helper also uses `os.path.basename` rather than `PurePosixPath`, which additionally splits a backslash path on Windows.

## Every edit in guard.py is in place

Same constraint as #116: `agent/guard.py` is frozen at **2892 lines** and sits exactly on the budget. So the five-line comment above `_PKG_MANAGER_NAMES` was rewritten to five lines that also cover `.exe`, and the one-line predicate body stayed one line via a walrus.

```
guard.py 2892 -> 2892 lines (delta 0)
```

## One gap deliberately not closed, pinned as a test

`_basename` lowercases only to **test** for the suffix and returns the name otherwise unchanged, so `PIP.EXE` becomes `PIP`, which is not in `_PKG_MANAGER_NAMES`. Windows filenames are case-insensitive, so that runs the same binary and is missed.

I found this because I wrote the opposite test first, assuming case-insensitivity, and it failed. Rather than widen the tables to make my assumption true, I inverted it: `test_an_uppercase_name_is_still_missed_and_that_is_pinned_not_fixed` records the actual behaviour.

Not fixed here because case-folding the name tables would have to move in **both** layers together to keep them agreeing, and widening what counts as an installer is a maintainer's call rather than mine. Say the word and it is a small follow-up.

## Evidence

```
$ uv run pytest -q tests/test_guard.py -k "two_layers_agree or windows_shaped_install or read_only_and_unrelated or uppercase_name"
23 passed, 204 deselected

# the same selection against UNMODIFIED guard.py:
11 failed, 12 passed

$ uv run pytest -q tests/test_guard.py
15 failed, 212 passed      # the same 15 as main: pre-existing Windows, #105

$ uv run pytest -q tests/test_structural_budget.py tests/test_ci.py
134 passed

```

The read-only control matters as much as the positive cases: `pip.exe list`, `uv.exe lock`, `uv.exe venv`, `notpip.exe install` and `pipx.exe install` are all still unmatched, so the suffix strip does not widen **what** counts as an install.

## Checklist

- [x] Tests cover the change, and they fail on the unfixed code. 11 of 23 fail against unmodified `guard.py`.
- [x] Test count and assertion count did not go down. 23 added.

`RELEASE_MANIFEST.txt` repinned last, and verified with `--strict` rather than the default mode after that bit me on #116.


eyalgolan (approved):
Approving, thanks. Reading the name through `_basename` at all three sites is the fix I'd have written, and I checked the one place this widens: `_is_pkg_manager_name('/usr/bin/pip')` is True now where it was False, but its only caller feeds it the token after `-m`, so nothing a user runs changes except the Windows spellings you meant to catch. Pinning the uppercase case as "still missed" instead of quietly folding the tables was the right call; that's a maintainer decision and you left it to me. Yours to merge.

Sreekant13:
Merged #115, and that put the other four behind, so this is one note rather than four.

`main` requires branches to be up to date, so landing any one of these makes the rest `BEHIND`, and updating them is a push, which per CONTRIBUTING means the most recent push needs approving by someone other than whoever pushed it. Four approved PRs therefore cost three extra review rounds no matter what order they go in. That is inherent to the settings, not something any of us did wrong, but it is your time so I would rather flag it than quietly queue it up.

I have merged `main` into all four and pushed:

| PR | state |
| --- | --- |
| #117 `.exe` strip | main merged, checks re-running |
| #118 lanes guard | main merged, checks re-running |
| #119 CONTRIBUTING figures | main merged, checks re-running |
| #116 Windows roots | main merged, plus the UNC platform gate you asked for |

Each was re-verified after the merge rather than assumed: `check_release_manifest.py --strict` clean on all four, and the suite each one touches re-run. The manifest auto-merged in every case, and since a clean auto-merge of a content-hash file is not the same as a correct one, I ran `--strict` rather than trusting the absence of a conflict.

Two things you might prefer instead, and either is fine by me:

1. **You land the remaining three yourself in one pass**, which skips the re-approval round-trips entirely.
2. **I stop merging and leave them to you** whenever more than one of mine is approved at once.

On #116 specifically: the platform gate is pushed and the reply is on that PR. `//tmp/x` and `//home/dev` are allowed again on POSIX, measured against a worktree at `origin/main` rather than inferred, and the drive and backslash forms stay ungated. It still needs your look since you had requested changes.

#117 and #116 both touch `tests/test_guard.py`, as you warned. They merged cleanly against each other here, but #116 is the one that will need care if #117 lands first, and I will handle that within a few minutes of it landing.

Sreekant13:
Correction to what I said above, before you act on it: **I was wrong about the re-approval cost.**

All three of #117, #118 and #119 came back `CLEAN` / `MERGEABLE` / `APPROVED` after I merged `main` in and pushed. The approvals were not dismissed, and each still shows a single APPROVED review. So updating a behind branch costs mechanical work on my side and **nothing of your review time**.

I took CONTRIBUTING's "if `main` moved, update the branch, which means a fresh approval" at face value and told you it would cost three review rounds, without checking what the protection actually does. That is the same mistake I made about squash-versus-merge earlier: read the sentence, but measure the behaviour.

So please ignore the two options I offered. There is nothing to decide, and I will merge the three approved ones now unless you say otherwise.

You may still want to look at that CONTRIBUTING parenthetical, though, since it describes a stricter rule than the branch protection enforces.

PR #118, opened 2026-09-07
Add tests/test_test_lanes.py, the guard ci.yml and CONTRIBUTING already cite

### LINKED ISSUE #109: ci.yml and CONTRIBUTING.md both cite tests/test_test_lanes.py, which does not exist
Two places name a test that pins the lane selectors against each other. The file is not in the tree.

`.github/workflows/ci.yml:247`:

> every node is in exactly one, which `tests/test_test_lanes.py` pins and the collection counts reconcile

`CONTRIBUTING.md:164`:

> `tests/test_test_lanes.py` pins the selectors in `ci.yml` and `scripts/run_tests.sh` against each other, and the two lanes' collection counts must sum to the unfiltered total

```
$ ls tests/test_test_lanes.py
ls: cannot access 'tests/test_test_lanes.py': No such file or directory

$ grep -rl "not slow and not nightly" tests/
(no matches)
```

No test file contains the selector string at all, so nothing is pinning the marker expressions, and nothing checks that the two lanes' collection counts sum to the unfiltered total.

## Why this one seems worth raising

Both passages are making the same argument, and it is the repo's own:

> a mistyped marker expression drops a node out of *both*, and a test that runs nowhere looks exactly like a test that passes

A guard cited by name in two places, that does not exist, is the same failure one level up. A reader who checks whether the lanes are protected finds a sentence saying yes.

I checked before asking: `git log --all -- tests/test_test_lanes.py` returns nothing in the history I have fetched, so it looks like the prose landed and the file never did, rather than the file being removed. My clone is a shallow-ish fetch of `main`, so that is suggestive rather than conclusive.

## Not the same as what I just added

PR #108 adds `tests/test_deselect_lists_agree.py`, which pins the `--deselect` lists in `ci.yml` and `run_tests.sh` to each other. That is a sibling of what is described here, not the same thing: deselect lists are node ids, the missing guard is about the `-m` marker expressions and the collection-count arithmetic. I deliberately did not widen that PR to cover this, and its commit message says so.

## Shape, if you want it

Two properties, both cheap and both mechanical:

1. The `fast` selector in `run_tests.sh` and the `pull_request` selector in `ci.yml` are the same string, and likewise `nightly` versus `workflow_dispatch`.
2. `collect-only` under each of the two selectors sums to the unfiltered total, which is the property that actually catches a node falling out of both lanes.

Point 2 costs three collection passes, so it probably wants a `slow` marker.

Happy to send it, but I would rather not guess at the intended name and shape of a file your own docs already describe, so tell me whether to write `tests/test_test_lanes.py` as specified or whether the prose is what should change.

Found while re-reading these two files for #19 / #108.

Sreekant13:
## What this changes

Fixes #109. Written under the cited name, `tests/test_test_lanes.py`, rather than editing the two citations away — so both sentences in `ci.yml:247` and `CONTRIBUTING.md:164` become true.

Thank you for the history; it explains the shape. Nothing here touches the export builder or a scrub workflow, which is what made the private version unshippable.

## The two properties, as four tests

| test | what breaks it |
| --- | --- |
| `test_the_pr_lane_is_spelled_identically_in_both_files` | `run_tests.sh fast` drifting from `ci.yml`'s pull_request selector |
| `test_the_nightly_lane_is_spelled_identically_in_both_files` | the same for `nightly` / workflow_dispatch |
| `test_the_push_lane_still_applies_no_marker_filter` | a third `-m` appearing, i.e. push silently stopping running everything |
| `test_every_node_is_in_exactly_one_lane` | a node falling out of both lanes |

Measured on this tree: **11467 + 243 = 11710**.

The count check is `slow`, as you asked. That also puts it in the nightly lane, which is the lane whose exit code `scripts/nightly_eval.sh` folds into its verdict — so a partition that silently stops holding is caught by the run that cares most about it.

## Mutation-tested, not merely run

A guard that has never been seen to fail is the thing this issue is about, so I broke each property in turn and restored the tree after each:

```
run_tests.sh fast    -> "not slow"        1 failed
run_tests.sh nightly -> "slow"            1 failed
ci.yml push lane gains a -m filter        1 failed

unmutated, the same three                 1 passed each
```

## Evidence

```
$ uv run pytest -q tests/test_test_lanes.py
4 passed in 25.84s

    tests/test_readme_claims.py tests/test_test_lanes.py
271 passed, 12 skipped

OK: 1525 file(s) match RELEASE_MANIFEST.txt
```

## One implementation note

`_workflow_selectors` deliberately returns the two `-m` expressions **in source order** and asserts there are exactly two, rather than searching for each lane by name. The push lane's branch is the empty string and carries no `-m` at all, so "exactly two" is what states that a push runs the whole suite; if someone gives push a filter, the count becomes three and `test_the_push_lane_still_applies_no_marker_filter` says so by name.

## Checklist

- [x] Tests cover the change, and they fail on the unfixed code. Mutation results above; this PR *is* tests, so each was broken deliberately.
- [x] Test count and assertion count did not go down. 4 added.

`RELEASE_MANIFEST.txt` repinned last, with the new file staged **before** the repin so it is actually listed, and verified with `--strict` — that ordering is what bit me on #116.


eyalgolan (approved):
Approving, and thank you for writing it under the cited name so the two sentences that pointed at nothing now point at something. I ran the partition check here and the arithmetic holds on my tree too (11450 + 244 = 11694), and both mutations you listed go red on exactly the test they should. The "exactly two -m expressions means push runs everything" reasoning is a nice way to pin the third lane without naming it. Yours to merge.

Sreekant13:
The `Python` failure here was not this PR, and not the change that preceded it. Recording the check rather than just re-running.

```
FAILED tests/test_guard.py::test_the_gate_mention_scan_is_not_quadratic
AssertionError: 0.717s on an 800-line script  (bound is 0.6s)
```

That is a wall-clock bound on a shared runner. This PR adds only `tests/test_test_lanes.py` and touches no product code, but its branch does contain #117, which routes guard tokens through `_basename` — so the honest question was whether I had made the guard slower.

Measured, on the same machine, in worktrees:

| tree | min | median | max |
| --- | --- | --- | --- |
| `d0dcb537` main, WITH #117 | 0.265s | **0.266s** | 0.297s |
| `bbc3d14` without #117 | 0.281s | **0.282s** | 0.296s |

No regression; marginally faster if anything, and both less than half the bound. Two more data points agree: **main's own run on `d0dcb537` is green**, and #119 passed this same test on a branch that also contains #117.

So it is a runner-speed flake. I have merged `main` in and pushed, which re-runs it.

Worth flagging separately, and I have not opened an issue because it may be known: that test asserts a wall-clock duration inside CI, and its own comment says "Loose bound — the shape, not a machine." 0.717s against a 0.6s bound is the kind of miss a noisy runner will produce periodically. If it has flaked before, a shape-based assertion (segments scanned, not seconds) would hold the same property without the timing sensitivity. Happy to look if that is useful, or to leave it.

Issue #10, opened 2026-08-20
`nh test --help` prints stale test counts (721) — the suite has ~9,400 tests


eyalgolan:
\`\`\`
$ nh test --help
  Modes:
    fast  — 711 tests, ~28s (skip slow integration tests)   full  — 721 tests,
    ~3min (all tests, parallel)   slow  — 10 tests, ~3min (eval replay +
    integration only)
\`\`\`

Source: the \`test\` command docstring in \`src/no_human/cli/commands.py\`. On the same tree:

\`\`\`
$ python -m pytest --collect-only -q | tail -1      # 9413 tests collected
$ python -m pytest --collect-only -q -m "not slow"  # 9176/9413
$ python -m pytest --collect-only -q -m "slow"      # 237/9413
\`\`\`

**Fix:** replace the hard-coded counts and timings with what each mode *selects* (\`fast\` = \`not slow and not nightly\`, \`slow\` = \`slow\`, \`full\` = everything) so the help never goes stale again. ~6 lines.

eyalgolan:
Want this one? Comment here to claim it — no assignment needed. I respond within 24 hours, and review newcomer PRs within 48. A claim with no visible progress or reply for 14 days goes back up for grabs. [CONTRIBUTING.md](https://github.com/no-human-ai/no_human/blob/main/CONTRIBUTING.md) has the credential-free dev setup — the test suite runs without any API key.

ShamikOfficial:
Claiming this. I'll replace the hardcoded counts/timings in the `nh test --help` docstring with the mode selection markers (`fast` = `not slow and not nightly`, `slow` = `slow`, `full` = everything) so the help stays accurate as the suite grows, plus a regression test that fails if the docstring drifts back to hardcoded counts.

ShamikOfficial:
PR up: https://github.com/no-human-ai/no_human/pull/102 — docstring now names the selectors each mode uses (aligned with `scripts/run_tests.sh`), fallback markers match, and a regression test fails if the hard-coded counts return.

PR #119, opened 2026-09-07
CONTRIBUTING: drop the frozen test counts, add the build step web needs

### LINKED ISSUE #111: CONTRIBUTING's test counts are stale (2,980 vs 11,710; 538 vs 1,576) and the documented web command fails on a fresh clone

Sreekant13:
## What this changes

Fixes #111. Same treatment as #103: the numbers go, the command that answers the question stays, and the timing claim names your CI job instead of an unattributed figure.

| claim | measured |
| --- | --- |
| Python "About 2,980 tests" | **11,710** collected (~4x) |
| Python "roughly four minutes" | **7m35s** for CI's own `Run tests` step |
| web "538 tests" | **1,584** (~3x) |
| desktop "167 tests" | **391** (~2.3x) |

The web and desktop figures are from running the documented commands, as you asked, not from counting declarations by hand.

For timing I used your `Python` job rather than my machine: on a push to `main` it applies no marker filter, so its `Run tests` step is a whole-suite `-n 4` run on a four-core `ubuntu-latest` runner — dated, public, and exactly the hardware shape the old sentence described.

## The part that matters is not the numbers

The documented web block was:

```bash
cd web
npm ci
npm test
```

and on a fresh clone that reports failures **on any platform**. Two tests assert against the built bundle and fail closed when `web/dist/assets` is absent. That fail-closed is right — an unbuilt tree reads as "cannot verify", not "clean" — but nothing told the reader to build.

Verified rather than assumed:

```
cd web && npm ci && npm test                    1584 tests, 6 fail
cd web && npm ci && npm run build && npm test   1584 tests, 0 fail
```

Four of those six were the Windows ESM bug from #113, already on `main`; the other two are the `dist` fail-closed this documents. `npm run build` is now in the block, with a sentence saying why it is not optional.

## On keeping rough scale

I kept "over 11,000", "over 1,500", "nearly 400" in prose, and pointed each at the command that prints the real figure. A precise count is stale the day after it is written — #10 said "~9,400" and it is 11,710 now — while an order of magnitude ages gracefully and still tells a newcomer what they are in for.

If you would rather have no figure at all, that is a one-line change in each place and I am happy to make it.

## Evidence

```
$ uv run pytest -q tests/test_readme_claims.py
133 passed, 12 skipped

```

## Checklist

- [x] Tests cover the change, and they fail on the unfixed code. Documentation only; the claims were verified by running each documented command, before and after.
- [x] Test count and assertion count did not go down. Unchanged.

`RELEASE_MANIFEST.txt` repinned last, verified with `--strict`.


eyalgolan (approved):
Approving, thanks. Everything in it survives being checked: the latest push-to-main Python step took 7m55s, so "roughly eight minutes" is honest; a fresh `npm ci --ignore-scripts` in desktop gives 395 tests, so "nearly 400" is honest; and the build step is the part that actually matters, since I reproduced the two fail-closed failures without it yesterday. Keeping the order of magnitude in prose and pointing at the command for the number is the right balance. Yours to merge.

PR #131, opened 2026-09-07
Reflect PR checks in merge policy

### LINKED ISSUE #124: Merge policy shows 'ready — 6 of 6' while the PR's required GitHub checks are red: the ci rule reads task.context.ci_status, which nothing writes
Caught live on PR #122 today: the task reached `awaiting_approval` with merge policy "ready — 6 of 6 rules satisfied", while the PR's required **File inventory** check was red on GitHub (two tracked files missing from `RELEASE_MANIFEST.txt` — real, actionable, exactly what that check exists to catch). The human gate was shown a clean verdict that the repo's own CI contradicted.

The mechanism, from reading the code rather than guessing:

- The `ci` rule's verdict comes from `facts.ci_state` (`src/no_human/core/merge_policy.py`, `_check_ci`), and under the default `success_or_unknown` a `None` state returns `True, "ci: none reported (tolerated)"`.
- `ci_state` is populated from `task.context.get("ci_status")` (`src/no_human/core/orchestrator.py`, the `GateFacts` construction around line 20175).
- I can find no writer of that key. `_run_ci` (orchestrator, around line 10377) records `ci_status` onto the **attempt row** via `store.update_attempt`, not into `task.context` — and it only runs at all when a `ci_runner` integration is configured. The other `ci_status` writers are the `pr_outcomes` table in `core/db.py`. Grep for a `task.context["ci_status"]` assignment comes up empty (and the same grep does find the attempt-row and pr_outcomes writers, so it's not the instrument).

Two consequences:

1. On a repo whose CI is GitHub Actions on the PR (this repo), the `ci` rule can never observe anything: it is permanently "none reported (tolerated)". The rule's tolerance for *unknown* was designed for "no CI configured", but here CI exists, ran, and failed — and the policy still said ready.
2. Even where `ci_runner` is configured, the policy reads a different location than `_run_ci` writes, so the rule looks structurally disconnected from the one component that produces the value. If there's a writer I missed, this issue should be closed with the file:line; the observed PR #122 behaviour still needs an explanation in that case.

What I'd want (smallest honest fix first): when a task enters `awaiting_approval` with a delivered PR, poll the PR's check rollup (the `gh`/API machinery the pipeline already uses for PR delivery and `pr_watch`) and write the aggregate into `task.context.ci_status`, then recompute the persisted merge-policy verdict for that head — so "ready" on the board means the checks the human can see on GitHub agree. Strict mode (`ci: success`) should also then be a real option for this repo instead of a rule that would block every task on "none reported". A failing rollup should read exactly like the other failed rules do, with the failing check's name in the detail.

Verification for a PR closing this: a fixture task whose delivered PR head has a failing check → policy shows the `ci` rule failed with the check name (the plant); the same head with checks green → 6 of 6 (control); a repo with genuinely no CI → "none reported (tolerated)" unchanged, so the tolerance still means what it was designed to mean. Suite green (`uv run pytest -q -n 4`; see CONTRIBUTING.md for why not `-n auto`).

For the record, the live incident resolved fine — the manifest rows were regenerated and CI went green — but only because a human went and looked at GitHub despite the board saying ready. That's the exact dependence this product exists to remove.

zerone0x:
## Summary
- poll the delivered PR's GitHub check rollup before moving a task to awaiting approval
- restamp task.context.ci_status and the current-head merge policy verdict from those PR checks
- preserve failing or pending check names in the CI rule detail, while keeping no-CI/unknown tolerant behavior unchanged

Closes #124

## Verification
- uv run pytest -q tests/test_merge_policy.py -q
- uv run pytest -q tests/test_merge_policy_wiring.py


eyalgolan (changes requested):
Hi, and thanks for taking #124 on; you read the mechanism correctly, and the adapter that names the failing check in the verdict detail is the right instinct.

I have to close this one, and it's timing rather than anything wrong with your reading. I landed my own fix for #124 as #132 (a496a5dc): the same behaviour, split into `vcs/ci_rollup.py` so the orchestrator's frozen line budget doesn't move, with the failing check names carried through to the detail and tests across ci_rollup, merge_policy, the wiring and pr_watcher. Your branch now conflicts with that in all three source files (merge_policy.py, orchestrator.py, test_merge_policy.py), and on its own it would also trip the structural-budget gate, since orchestrator.py is frozen in tests/test_structural_budget.py and the inline block adds 59 lines to it.

If you spot a gap in what landed, a case your version handles and mine doesn't, please open an issue with the case and I'll take it seriously. And if you'd like to pick up something else, the open issues are up for grabs; a first PR needs the `contributors/zerone0x.md` file the bot described, one line of agreement, once.

Sorry for the collision, and thanks again.

PR #134, opened 2026-09-07
Measure the gate-mention scan's shape, not the runner's speed

### LINKED ISSUE #125: test_the_gate_mention_scan_is_not_quadratic flakes on shared runners: 0.64-0.72s against a 0.6s wall-clock bound, 3 of the last 12 CI runs
Three of the last twelve ci.yml runs failed on the same assertion, all on unrelated changes:

```
FAILED tests/test_guard.py::test_the_gate_mention_scan_is_not_quadratic - AssertionError: 0.641s on an 800-line script
FAILED tests/test_guard.py::test_the_gate_mention_scan_is_not_quadratic - AssertionError: 0.677s on an 800-line script
FAILED tests/test_guard.py::test_the_gate_mention_scan_is_not_quadratic - AssertionError: 0.717s on an 800-line script
```

One of them cost a JS-only PR (#122, no Python in the diff) a full suite rerun today.

The test (tests/test_guard.py:381) guards a real regression class — a quadratic that once made `guard.evaluate` take 3.4s against a 138ms baseline inside a PreToolUse hook — and its own docstring says the right thing: "Loose bound — the shape, not a machine." But the assertion is a machine bound: `elapsed < 0.6` wall-clock via `time.monotonic()`, and GitHub's shared runners now land 6-20% over it on good code. A genuine reintroduction of the quadratic would blow past 3s; everything between 0.6s and that is noise territory where only flakes live.

Suggested fix, in line with the docstring: measure the shape instead of the machine. Run the same `guard.evaluate` on a 400-line and an 800-line script and assert the ratio stays near-linear (say `t800 / t400 < 3`); a same-class quadratic doubles the input and quadruples the time, so it still goes red with margin, and runner speed cancels out of the ratio. (`time.process_time()` for both measurements would additionally drop scheduler noise.) If a ratio feels too clever, the fallback is an honest slack bound — 2.0s still sits 40% under the regression it exists to catch — but the ratio is the version that actually matches "the shape, not a machine".

Verification for a PR closing this: re-introduce the quadratic locally (unmask the command once per segment, as the docstring describes) and show the new assertion goes red — the plant; on current code, run the test 20x serially and show 20 passes — the flake control; full suite green (`uv run pytest -q -n 4`; see CONTRIBUTING.md for why not `-n auto`).

Sreekant13:
Closes #125.

The docstring already said the right thing, "the shape, not a machine", and the assertion underneath it was a machine. Replaced with the ratio you suggested.

## The change

`guard.evaluate` runs on a 400 line and an 800 line script of the same shape, and the assertion is `t800 / t400 < 3.0`. A linear scan doubles, a quadratic quadruples, and the runner speed cancels out of the ratio.

Measured with `time.process_time()` rather than `time.monotonic()`, as you suggested, so a scheduler preemption inside either half does not land in the ratio.

## Measurements

All on one machine, current code unless stated.

| | t400 | t800 | ratio |
|---|---|---|---|
| linear, 15 samples | 0.140 to 0.157 s | 0.265 to 0.375 s | **1.70 to 2.39** |
| quadratic put back by hand, 3 samples | 4.38 to 4.58 s | 17.2 to 17.7 s | **3.87 to 3.97** |

3.0 sits between them: 26% clear of the observed linear worst case, 22% under the observed quadratic best. The median linear ratio is 2.00, exactly as a linear scan should give.

I tried min of k sampling first, thinking the spread was scheduler noise. It is not: min of 3 moved the worst case only from 2.38 to 2.25 and tripled the runtime, so it is not in the change.

## Verification

**The plant.** Reintroduced the exact quadratic, the whole command unmasked once per segment rather than hoisted, and the ratio assertion goes red with the diagnosis in the message:

```
AssertionError: 4.328s at 400 lines, 16.891s at 800, ratio 3.90: doubling the
input did much more than double the work, so the gate-mention scan is unmasking
the whole command once per segment again
```

**The flake control.** 20 serial runs on current code, 20 passes, 0 failures.

**The flake reproduced, which I did not expect to manage.** Running the same
measurement while a saturating `pytest -n 4` held the CPU puts the old bound
red 7 times in 10, at 0.641 to 0.735 s. That is the same 0.64 to 0.72 s band
you saw on the shared runners, so the flake is CPU contention and it is
reproducible on demand rather than only on GitHub. The new bound passes 10 of
10 through the same load, worst ratio 2.31:

```
run  1  ratio 1.52   wall t800 0.578s
run  4  ratio 2.00   wall t800 0.641s  OLD BOUND RED
run  6  ratio 2.00   wall t800 0.703s  OLD BOUND RED
run  8  ratio 1.88   wall t800 0.735s  OLD BOUND RED
...
new ratio bound (<3.0): 10/10 pass, worst 2.31
old wall bound (<0.6s):  3/10 pass
```

**Full suite, and I am not going to call it green, because on Windows it is
not.** `uv run pytest -q -n 4` on this branch gives 499 failed, 10645 passed,
645 skipped. So I ran the same command on unmodified `origin/main` in a
detached worktree and got 497 failed, 10636 passed, 656 skipped, then diffed
the two failure sets by name rather than by count.

* `tests/test_guard.py`: **15 failures on both, the same 15**, and this test
  is in neither list.
* Only on main, 2: `test_get_task_phases_populated` and
  `test_stale_reports_never_used_and_never_archives`. Ordinary churn.
* Only on this branch, 4: `test_sdist_members_are_exactly_the_allowlist`,
  `test_wheel_installed_in_a_clean_venv_serves_the_board`,
  `test_wheel_in_a_clean_venv_answers_the_first_commands`,
  `test_the_worktree_can_actually_build_after_provisioning`.

Those 4 looked like a coherent cluster rather than churn, so I chased them.
All 4 fail from one root cause, `hatch_build.py.BoardStaleError`: `web/dist`
is stale against `web/src` in my long lived checkout, which a freshly created
worktree cannot be. Checking `origin/main` out **in that same checkout** and
running `test_sdist_members_are_exactly_the_allowlist` reproduces the failure
on unmodified main, so it is my working copy, not this commit. The diff here
touches no file under `web/`. All 4 are also `@pytest.mark.slow`, so the
`pull_request` lane does not run them at all.

The honest summary is therefore: this branch and main are indistinguishable
on Windows apart from build tests that fail identically on main in the same
checkout, and CI is the authority for the Linux result.

## One judgement call beyond what you asked for, easy to drop

A ratio cannot see a UNIFORM slowdown, which the old bound could, so I kept a wall clock check as a backstop at 3.0 s. That is roughly 10x what the 800 line scan costs, so it should never flake, but it is genuinely loose: it only catches something on the order of a 10x regression.

I did not want to ship it as untested code, so I planted a uniform slowdown too, the same linear work repeated 400 times. The ratio assertion stayed green at 2.06 to 2.24, correctly, and the backstop fired on its own:

```
AssertionError: 4.391s of CPU on an 800-line script, against roughly 0.3s when
this bound was written. The ratio above passed, so this is not the quadratic;
something has made the whole scan far slower
```

So the two assertions catch different things and each stays quiet on the other's plant. If you would rather have only the ratio, say so and I will drop the second assertion and its paragraph in the docstring.

## One note on platforms

`process_time` has 15.625 ms resolution on Windows, which quantizes a 140 ms measurement. It does not affect CI, since the windows job runs only `test_windows_portability.py`, `test_windows_credential_file.py` and `test_check_release_manifest.py`, and this test is Linux only there. It does mean the numbers above, taken on Windows, are the pessimistic case: the ratio spread on the Linux runners should be tighter than 1.70 to 2.39, not wider.

`RELEASE_MANIFEST.txt` repinned, `check_release_manifest.py --strict` clean.


eyalgolan (approved):
Approving, thanks Sreekant. I ran both of your plants on my side and each fires with the right diagnosis: the per-segment unmask put back gives ratio 3.90 (0.859s at 400 lines, 3.353s at 800), and a uniform per-segment slowdown leaves the ratio alone and trips the backstop at 3.437s. Ten serial runs pass here, and five more pass while a `-n 4` run of test_guard, test_vcs and test_readme_claims saturates the machine, which is the contention shape from the issue. Keep the backstop; it is cheap and it catches the one thing a ratio cannot.

Reading the Windows suite by name rather than by count, and not calling it green, was the right way to report it. The branch merges clean with current main; once you update it I'll re-approve (the update dismisses this one, that's the branch rule, not a change of mind). Then it's yours to merge.

Sreekant13:
Merged. It was already 0 commits behind main, so no update was needed and your approval stood; nothing to re-approve.

Thanks for running both plants independently. Your numbers are the useful part: 0.859s at 400 and 3.353s at 800 on your machine against 0.140/0.265 on mine, roughly 6x slower, and the ratio still lands at 3.90 against my 3.87 to 3.97. That is the property the change is built on, measured on a second machine, which is better evidence than anything I could produce alone.

Backstop stays.

Issue #128, opened 2026-09-07
A task attempt's env setup rewrites the shared .venv's editable path to its own ephemeral worktree, breaking every nh invocation once that worktree is gone


eyalgolan:
Caught live tonight, mid-approval. `nh approve` worked at 20:15; at 21:5x every `nh` invocation failed with `ModuleNotFoundError: No module named 'no_human.cli'`.

The chain, verified step by step on the affected machine:

- `.venv/lib/python3.12/site-packages/_editable_impl_no_human.pth` (mtime 20:56) contained a single path: `~/.no_human/worktrees/05b710174e4a4136b33f09ed10833cd5.88111.a7e38550/src` — the src of a task attempt's worktree, not the repo checkout the venv belongs to. The attempt for that very task had run its env setup inside its worktree against the shared `.venv` and the editable install repointed the whole venv at the worktree.
- That worktree no longer exists (a sibling `...125763f3` for the same task does — the attempt created more than one). With the `.pth` path dead, `import no_human` fell back to the namespace directory physically present in site-packages, which holds only the data dirs (`migrations`, `web_dist`) — hence `no_human.cli` not found, `no_human.__file__ = None`, `__path__ = [.../site-packages/no_human]`.
- Restoring the `.pth` to the repo's own `src` immediately fixed `nh` (`no_human.cli` imports from the checkout again, `nh --version` prints 0.2.0). The long-running server on 8420 was unaffected throughout because its process had loaded modules before the rewrite — which is exactly what makes this failure quiet: the board keeps working while every fresh CLI process is broken.

So the defect class: an attempt whose target repo is no_human itself (dogfooding) runs `uv`/pip editable install in its worktree, uv resolves the project's venv to the primary checkout's `.venv`, and the install mutates shared state that outlives the worktree. Any operator command between that install and... forever (the worktree is deleted at cleanup) hits a broken CLI.

What a fix has to establish: an attempt's environment work must never write outside its own worktree — either an isolated per-attempt venv for self-targeted repos, or the install pinned to the worktree's env (`UV_PROJECT_ENVIRONMENT`/`VIRTUAL_ENV` pointing inside the worktree), with a regression test that runs a task-shaped editable install in a temp worktree against a repo with a populated `.venv` and asserts the repo venv's `.pth` (and site-packages generally) is byte-identical before and after (the plant: the current behaviour fails that assertion).

[...]

PR #145, opened 2026-09-08
Scope an attempt's environment commands to its own worktree

Sreekant13:
Part of #128, not a fix for the observed incident. See the scope note at the end.

I reproduced your incident in a controlled repo before writing anything, because the quiet part (the server keeps serving while every fresh CLI is dead) makes it easy to fix the wrong thing.

## The reproduction

A demo package installed editable into its own `.venv`, a linked worktree, then `uv pip install -e .` from inside the worktree with `VIRTUAL_ENV` inherited, exactly as `env_setup` ran it:

```
BEFORE  ...\repro128\primary\src
AFTER   ...\repro128\wt\src
```

Then removing the worktree and running the primary interpreter:

```
ModuleNotFoundError: No module named 'demopkg'
```

Same chain you described, same symptom.

## The fix

`core/worktree.py` gains `isolate_attempt_env`, which pins `VIRTUAL_ENV` and `UV_PROJECT_ENVIRONMENT` to `<worktree>/.venv`.

Both keys move together. uv honours either, so pinning one and leaving the other still reaches the shared venv.

**The venv has to be created, not just named, and that is measured rather than assumed.** My first attempt pinned the variables and nothing else. `uv pip install` does not build a missing environment:

```
error: Failed to inspect Python interpreter from active virtual environment at `.venv\Scripts\python.exe`
  Caused by: Python interpreter not found at `...\wt2\.venv\Scripts\python.exe`
```

The shared venv was safe, but the attempt's own command now failed. So a pin without a venv behind it converts a corrupting install into a failing one. `isolate_attempt_env` creates the venv when the worktree looks like a Python project.

**A creation failure does not fall back to the shared venv,** because falling back is the bug. The pin stands, the attempt's command fails loudly, and the operator's venv survives.

The redirect applies only when all three hold, so ordinary runs are untouched:

* the path is a LINKED worktree, its `.git` a file carrying a `gitdir:` line. `isolation.enabled: false` runs the attempt straight in the operator's own checkout, and repointing their venv would be a worse bug than this one.
* the inherited environment actually names a venv. Nothing to displace otherwise.
* that venv lies outside the worktree. One already inside it is the isolation this exists to create.

This is the residual `agent/venv_install_guard.py` already names in its own module docstring, and it says why no command pattern can close it: the information does not exist before the command runs, only environment scoped BEFORE the command can. This is that scoping for the commands the orchestrator itself runs. The coder session's own environment is the larger half of that note and I have not attempted it here.

## A second call site, which the issue did not name

I wrote the wiring test to assert over EVERY shell `subprocess.run` in `orchestrator.py` rather than only the one from the incident, and it went red before I had planted anything. `env_teardown` at line 5166 has the same defect: shell commands, `cwd` set to the worktree, no `env=`. Both call sites are wired now.

## Verification

**The plant is inside the end to end test.** It hashes every file under the shared venv's `site-packages`, runs the shape the code had before this fix and asserts the digest CHANGES, restores, then runs the same install through `isolate_attempt_env` and asserts it does not. If the defect ever stops reproducing, that test fails rather than passing vacuously.

**Both wiring plants fire and each names its own line:**

```
dropping env= from env_setup     -> "orchestrator.py line 5063 runs a shell command with no env="
dropping env= from env_teardown  -> "orchestrator.py line 5166 runs a shell command with no env="
```

The wiring test reads the AST, not the source text. A substring search would be satisfied by the word appearing in a comment, which has caught this repo before.

**No new failures.** The directly affected files (`test_worktree_isolation.py`, `test_worktree_teardown.py`, `test_venv_install_guard.py`, `test_structural_budget.py`, plus the new file) give 17 failed, 62 passed on this Windows host, and those 17 are identical BY NAME to the 17 the same files produce on unmodified `origin/main` here (the diff is empty). I am comparing names rather than counts on purpose. `test_structural_budget.py` passes, so the frozen budgets hold: `core/orchestrator.py` is edited entirely in place and is still 21788 lines.

`RELEASE_MANIFEST.txt` regenerated with the new test file staged first, `--strict` clean.

## One cost you should weigh, and an alternative

A Python repo whose `env_setup` installs will now build a venv per attempt instead of reusing the shared one. On my trivial demo package that was about 3 seconds; on no_human itself it is a full dependency install, which is a real cost on every dogfooding attempt.

Your issue offered two shapes, "an isolated per-attempt venv for self-targeted repos" or "the install pinned to the worktree's env". I built the second because it establishes the invariant for every repo rather than only the self-targeted case, and because it needs no heuristic for deciding what "self-targeted" means. If the per-attempt cost is not one you want to pay generally, the narrower version is a smaller change on top of this: keep `isolate_attempt_env` and gate its call on the worktree's repo being the one this process is installed from. Say which you prefer and I will follow it.


Sreekant13:
CI caught something I did not know about and it changed the fix for the better.

`tests/test_egress_allowlist.py` flags any exec whose program it cannot name. My first version built the worktree venv with `uv venv`, and `uv venv` MAY fetch an interpreter, so that is a real network call sitting inside an attempt's environment setup. I had not thought about it as egress at all.

So the venv is now built with `sys.executable -m venv`, which provably cannot leave the machine, and there is no network in this path to declare. The channel still reports `<dynamic>` because argv[0] is `sys.executable` rather than a literal name, so `core/worktree.py` gains an allowlist entry saying exactly that, plus a test that pins the argv so a future edit back to `uv venv` is a visible one rather than a silent reintroduction.

Force pushed. The gate was right and the product is better for it.

eyalgolan (changes requested):
Thanks Sreekant. Reproducing first and pinning fail-closed were the right instincts, and the CI-driven switch away from `uv venv` made it better. Three things before it lands, and the first is a real defect.

**1. The frozen desktop bundle.** `_create_venv` runs `[sys.executable, "-m", "venv", target]`, and in the PyInstaller build `sys.executable` is the `nh` binary itself. I ran that argv against nh here:

```
$ nh -m venv /tmp/probe
Usage: nh [OPTIONS] COMMAND [ARGS]...
Try 'nh --help' for help.
(exit 2, no venv created)
```

Then the pin stands, so in the packaged app every env_setup command on any repo with a pyproject.toml fails instead of running. `testing/repro_gate.py` already solves this exact case (`_pytest_python`: under `sys.frozen`, the repo's own venv interpreter, then python3/python on PATH); the same fallback here, plus a test with `sys.frozen` monkeypatched asserting argv[0] is not the nh binary, closes it.

**2. Main moved under you today.** 5760cef4f added `run_setup_commands` in core/worktree.py, a second orchestrator-run shell path (the profile's `setup_cmds`, run through `runner._run_shell` with `runner._env_for(worktree)`), and `_env_for` inherits the process `VIRTUAL_ENV` when the worktree has no venv yet. Your AST test only walks orchestrator.py, so it cannot see it. After the rebase, route that env through `isolate_attempt_env` as well. The rebase also has an import conflict (main's worktree import is a multi-line block now); fold `isolate_attempt_env` into an existing line, or the structural budget trips by one. I hit both on my trial merge.

**3. What this actually closes.** I went back to the task in the issue. No task in my database has ever had `env_setup` configured (0 of 939), and the receipts show the write came from the coder session itself: `uv run --active ruff check ...` inside the worktree, at the minute the .pth changed. I reproduced it in a throwaway repo: `uv run --active` with an inherited `VIRTUAL_ENV` rewrites that venv's editable path to the worktree; plain `uv run` does not. That is the half you explicitly left out, and it belongs to the venv install guard rather than this PR, so nothing to do here except change "Closes #128" to "Part of #128" and say the same in the two docstrings. Establishing the invariant for the orchestrator's own commands is still worth having.

On the cost question: keep the general version. env_setup is rare enough that a per-attempt venv is a cost nobody pays today, and the invariant is worth more than a heuristic.

For the record, everything else checked out on the trial merge: 119 passed across the isolation, egress, worktree and guard files plus your slow end-to-end test with real uv, and dropping `env=` from env_setup fails the wiring test naming line 5156.

Sreekant13:
All three addressed, and the first one was a real defect I shipped. Thank you for running that argv.

**1. The frozen bundle.** `_builder_python` now falls back the way `repro_gate.py::_pytest_python` does: the venv being displaced first (the caller only pins when the inherited environment names one, so it exists by construction and is a real interpreter), then `python3`/`python` on PATH, then None, which keeps the fail-closed contract rather than reaching for the shared venv. Two tests with `sys.frozen` monkeypatched: one asserts argv[0] is the displaced interpreter and not `sys.executable`, one asserts that with no interpreter at all the pin still stands and no venv appears. Reverting to `sys.executable` fails the first.

**2. `run_setup_commands`.** Routed through `isolate_attempt_env` after the rebase. You were right about why I could not have caught it: my AST test only walks orchestrator.py, so there is now a second one walking worktree.py, and unrouting the call fails it. The import conflict resolved as you said, folded onto an existing line rather than added as a new one; orchestrator.py sits at main's 21880.

**3. Scope.** Now "Part of #128" in the title line of the PR body and in both docstrings, with your measurement recorded: 0 of 939 tasks have ever configured `env_setup`, and the receipts put the write in the coder session. I would not have found that from the code, and it changes what this PR honestly claims.

Keeping the general version per your call on the cost question.

Gate after the rebase: 11 passed in the isolation file including the slow end-to-end with real uv, and the worktree/venv-guard/budget files give 17 failures identical BY NAME to current main on this Windows host, re-measured against 5760cef4f rather than reusing the older baseline.

Sreekant13:
Rebased onto ce4d4a7 (which is #151 landing) and pushed; CI is running again.

Worth noting what just happened, since it is the case #151 was about. My previous push showed ONLY the CLA nudge check and no CI at all. That is the conflict signature: GitHub cannot build the merge ref for a conflicting PR, so every `pull_request` workflow is skipped silently, while `cla-nudge.yml` runs on `pull_request_target` against the base and fires anyway. It reads as a half-finished run rather than a blocked one.

The gap was one commit and the file that collided was the generated manifest, so it resolved by regenerating. But #151's own rule would have flagged this branch for a different reason: the landing touched orchestrator.py and so does this branch, which is a real overlap rather than a manifest one.

eyalgolan (changes requested):
Thanks Sreekant, all three landed exactly as asked and I checked each on a trial merge with current main: the frozen-build call with `sys.frozen` set and `sys.executable` pointing at nh now pins and builds the venv from the displaced interpreter; unrouting `run_setup_commands` fails your second AST test; putting `sys.executable` back in `_builder_python` fails both frozen tests; 124 passed across the isolation, egress, worktree, guard and setup-cmds files plus the slow end-to-end with real uv; orchestrator.py sits at main's count.

One more thing, and I only found it by running the plain-pip shape rather than the uv one. The comment above `_VENV_KEYS` says both variables are honoured by uv and pip, and that is true of uv but not of pip: pip installs into the interpreter it runs under and never reads `VIRTUAL_ENV`. What routes a bare `pip` is PATH, and in the environment this PR produces PATH still leads with the shared venv's bin, because the server was started from it and `_env_for` only prepends a worktree venv that already exists at the time it is called, which yours is not yet. I reproduced it with two real venvs:

```
VIRTUAL_ENV=<wt>/.venv UV_PROJECT_ENVIRONMENT=<wt>/.venv PATH=<shared>/.venv/bin:$PATH  pip install -e .   (cwd = worktree)
which pip            -> <shared>/.venv/bin/pip
shared .pth after    -> <wt>/src          (the #128 rewrite, pin and all)

control, PATH=<wt>/.venv/bin:$PATH, same command
shared .pth after    -> <primary>/src     (untouched)
worktree .pth        -> <wt>/src
```

So `isolate_attempt_env` should also put the worktree venv's bin directory (`Scripts` on Windows) at the front of PATH once the venv exists, and the comment should say pip follows PATH. The end-to-end test can then run a bare `pip install -e .` alongside the `uv pip` one; the venv module seeds pip, so it is there. That closes the last way an orchestrator-run command reaches the shared venv, and then I'm happy to approve.

Sreekant13:
Done, and you were right that the pin was only half the job. That comment was mine and it was wrong: I wrote that both variables are honoured by uv and pip without checking pip, which reads neither.

`isolate_attempt_env` now also puts the worktree venv's bin (`Scripts` on Windows) at the FRONT of PATH, once the venv exists on disk. Not before that: pointing PATH at a directory that was never built would shadow nothing and swap a clear failure for a confusing one. The comment now says pip follows PATH.

Two unit cases pin the PATH order and the never-built case, and there is a second slow end to end case running a BARE `pip install -e .` alongside the uv one. That test is the one that was failing. With the prepend removed it reports:

```
a bare pip still resolves outside the worktree
(...\primary\.venv\Scripts\pip.EXE), so it would install into the shared
venv no matter what the pin says
```

One deliberate choice in that test worth flagging, because the obvious version would have been silently useless: it asserts resolution with `shutil.which(..., path=isolated["PATH"])` rather than by spawning a bare `pip`. Windows resolves a bare program name from the PARENT process PATH, not from the env handed to the child, so a spawn-based test would pass on Windows while testing nothing. It then installs through the resolved path and re-digests the shared venv.

Gate after rebasing onto current main: 14 passed in the isolation file including both end to end cases, and the worktree, guard, budget and egress files give 17 failures identical by name to current main on this host. `--strict` clean, `reanchor_citations.py` VERDICT=OK.

Also, thanks for merging #156.

eyalgolan (approved):
Approving, thanks Sreekant, and thanks for taking the pip finding as a fact to measure rather than a comment to answer. The prepend-only-once-the-venv-exists choice is the right one, and the reason you gave for asserting resolution with `shutil.which(path=...)` instead of spawning a bare `pip` is exactly the kind of thing that keeps a test honest on the platform it runs on.

Checked on your head, which is level with main: manifest strict-clean at 1552 files, VERDICT=OK, orchestrator.py at main's count, 127 passed across the isolation, egress, worktree, guard, budget and setup-cmds files, and both slow end-to-end cases pass with real uv and real pip. Three plants, each red on exactly the tests you'd want: dropping the PATH prepend fails the PATH-order unit test and the bare-pip end-to-end test with the "still resolves outside the worktree" message; unrouting `run_setup_commands` fails its AST test; putting `sys.executable` back in `_builder_python` fails both frozen-build tests.

Yours to merge.

Sreekant13:
Merged. Thank you for three rounds of genuinely load-bearing review on this one.

Worth stating plainly for the record: each round found something my own verification could not have. The frozen build because my tests never run frozen; `run_setup_commands` because my AST test only walked orchestrator.py and the path landed mid-review; the pip half because I wrote a comment asserting pip reads VIRTUAL_ENV without checking, and every test I had used uv, which does.

#128 stays OPEN on purpose. What landed is the invariant for the commands the orchestrator runs itself. The incident you traced belongs to the coder session (`uv run --active`), and `agent/venv_install_guard`'s own docstring already names that residual and says why no command pattern can close it. Happy to take that half as its own piece of work if you want it; otherwise I will leave it with you rather than assume.

PR #149, opened 2026-09-08
Node-suite failures caused by missing build prerequisites (app-builder-lib absent, web/dist missing) are judged 'tests failed' and retried; classify them as environment errors

eyalgolan:
## Acceptance criteria
- A faithful-size incident TAP (417 tests, real failures at ordinals 182/343, both signatures outside the last 8000 bytes of output) classifies as environment through both orchestrator call sites; a real assertion failure and a Python `ModuleNotFoundError` do not; a coder-owned failing test is never excused regardless of its text.
- Full Python suite green (whole summary line quoted below); strict `RELEASE_MANIFEST.txt` check green; `_run_attempt` stays at its frozen 2210 lines / CC 257; `src/no_human/api/app.py` stays at 6140 lines.

## Changes
Independent review of the original fix (f999a591) found it classified from `TestRunResult.output`, which is truncated to the last 8000 bytes — the real incident's failing-test signatures don't live in that tail on a 400+ test run, so the classifier would have missed the actual case it was built for. This round fixes that plus five other findings, without renaming `runner.missing_prerequisite_reason` / `orchestrator._environment_test_failure` or touching `app.py`/`desktop/`/`web/`.

- `src/no_human/testing/runner.py`
  - `_tap_failure_blocks` (runner.py:403) parses `not ok` TAP blocks off the **full**, untruncated `node --test` output before any of `run_tests`'s 6 truncating return sites; the new `TestRunResult.failure_blocks` field carries them through.
  - `missing_prerequisite_reason` (runner.py:583) and `prerequisite_reason_for` (runner.py:597) classify from `result.failure_blocks`/`result.traceback_excerpts`, never from `result.output`.
  - `_MISSING_PREREQUISITE_RULES` (runner.py:565): the "is missing — run `npm run build`" rule now matches the real em-dash byte sequence (`\xe2\x80\x94`, U+2014) as it actually appears in `web/src/cancelFlow.test.mjs`/`eventLabels.test.mjs`, in addition to the ASCII hyphen. Added the reversed ENOENT/path rule for `spawnSync .../node_modules/.bin/foo ENOENT`.
- `src/no_human/core/orchestrator.py`
  - `_environment_test_failure` (orchestrator.py:11141) now takes `owned_failing` and returns `None` immediately if it's non-empty — a coder-owned failing test is never excused as environment no matter what its output says. It also sets a terminal `status="failed"`, `infra_failure=1` on the attempt row (previously left `in_progress`), and its blocker question now says "primary checkout" (the resume path runs in a new worktree seeded from there via `runner._ensure_node_deps`/`_ensure_forced_build_artifacts`), not "this checkout".
  - Both call sites now compute run attribution (`_owned_failing_tests`, orchestrator.py:11258) **before** consulting the environment classifier: the single-run site at orchestrator.py:6329-6331, and the layered-plan site at orchestrator.py:11587-11589 inside the new `_layered_tests_failed_outcome` (orchestrator.py:11549).
  - `_layered_tests_failed_outcome` (orchestrator.py:11549) and `_failed_tests_outcome` (orchestrator.py:11636) were extracted from `_run_attempt`'s two failure-billing call sites so `_run_attempt` itself stays net-zero against its frozen budget (2210 lines / CC 257, unchanged in `tests/test_structural_budget.py`); only the file-level `core/orchestrator.py` entry moved to reflect the extracted code.
- `tests/test_missing_prereq_env_classification.py`: rebuilt around `_incident_tap()`, a 417-test TAP with failures at ordinals 182/343 whose byte offsets from the end are pinned >8000; drives `runner.run_tests` end-to-end (BLOCKER-1), pins the real em-dash bytes (BLOCKER-2), drives the layered call site via `orch._run_attempt` (MAJOR-5), and pins that Python `ModuleNotFoundError`/"No module named" text is *not* matched (MAJOR-6, no rule added for it — out of scope).
- `tests/test_structural_budget.py`: only `FROZEN_FILE_LINES["core/orchestrator.py"]` moved (documented inline); `_run_attempt`'s own frozen entries and all `MAX_*` thresholds are untouched.

## How I verified this
```
$ uv run pytest -q tests/test_missing_prereq_env_classification.py
15 passed

$ uv run pytest -q tests/test_structural_budget.py tests/test_base_tree_gate.py tests/test_handoff.py tests/test_infra_not_work.py tests/test_e2e_orchestrator.py tests/test_runner.py
347 passed

$ uv run pytest -q -n 4
11628 passed, 220 skipped, 8 warnings in 670.54s (0:11:10)

OK: 1534 file(s) match RELEASE_MANIFEST.txt

$ wc -l src/no_human/api/app.py
6140 src/no_human/api/app.py
```

Also mutation-checked (broke each fix, confirmed the corresponding test goes red, restored byte-for-byte, confirmed green again): removing the layered call site, reverting the em-dash rule to ASCII-only, and adding a "No module named" rule.

## Assumptions
- The environment excuse is gated on `owned_failing` (from `_owned_failing_tests`) at both call sites, not on the broader `newly_failing` set — matching the existing single-run site's pre-review behavior and the layered site's new one.


eyalgolan:
Superseded by later rounds of the environment-error classification lineage; landed on main as e9dd5e07.

PR #151, opened 2026-09-08
Rebase a stale base when the gap is small but touches the branch's own files

### LINKED ISSUE #141: Base-staleness rebase gates on commit count, but tonight's three conflict rounds all came from 1-3 commit gaps that touched the branch's own files
BASE_STALENESS_REBASE_THRESHOLD (src/no_human/core/orchestrator.py, ~line 926) rebases a retry's branch only past a 5-commit gap, and the comment records why: 1 would rebase on nearly every retry because most gaps are noise — an unrelated commit landing on main mid-attempt — and the incidents that set the number (db9da7f7 at 5 commits/6.9M tokens, e6719bd8 at 140 commits/37.9M) were both about big gaps. That reasoning is sound and this issue is not asking to lower the threshold.

But in one evening (2026-09-07/08) three tasks each paid a full conflict-resolution round — re-implement, re-review, re-test — on gaps the threshold correctly ignored, because the gap was small and *related*:

- 0e1edabb: repro gate errored 'repro tests not found at base — the base moved under this ticket' after one landing (d071a751) touched the same web/src area mid-attempt.
- 05b71017: approve's full gate on the diverged tree failed against 4cbf73e3, which had changed the same file (vcs/manifest_repair.py) two commits earlier.
- 82644133: same again — its desktop/updater.mjs + updatePolicy.mjs edits collided with 72ca3a0b (the updater-feed landing), one commit earlier.

The pattern: when the fleet lands several fixes in the same subsystem back-to-back — exactly what a release push looks like — every follower pays one round. Commit count can't see this; the signal that would is *file overlap*: at the `_refresh_stale_base` call site, the branch's changed files and the base-gap's changed files are both one `git diff --name-only` away. A gap below the threshold whose files intersect the branch's files is not noise; it is a guaranteed conflict round waiting at the approve gate.

Suggested shape: keep the >=5 unconditional rebase; below it, rebase when `changed_files(branch) ∩ changed_files(base_gap)` is non-empty (RELEASE_MANIFEST.txt excluded from the intersection — the generated pin-file intersects everything and gets regenerated anyway). Same never-fail contract as today: measurement or rebase failure degrades to advisory. The comment block at ~1153 records an older incident where rebase rounds under an already-reviewed PR looped until killed — whatever guard exists there today must keep applying, i.e. this only widens the *pre-review* rebase, never a post-review one.

Verification for a PR: a fixture where main gains one commit touching a file the branch also touches → rebase happens (the plant); one commit touching a disjoint file → no rebase (the control that preserves the 'most gaps are noise' decision); a >=5 disjoint gap still rebases as today. Evidence in the PR: the three task ids above as the motivating incidents.

Sreekant13:
Closes #141.

The `>= 5` unconditional rebase is untouched, and this only widens the pre-review rebase.

## The shape

New `core/base_staleness.py` holds the decision, so it is testable without a repository:

* `overlapping_paths` is pure set logic
* `base_gap_overlap` asks git
* `should_rebase(behind, threshold, overlap)` keeps the threshold and adds the overlap
* `staleness_record` builds the `task.context` payload, now carrying `overlapping_files`

That last field matters more than it looks. Without it a reader sees a rebase on a 1-commit gap and has no way to tell how it tripped a 5-commit rule.

New `GitRepo.files_changed_since_fork(base, on_base=...)` returns one side's changed files. Three dot on purpose, for the reason `vcs/git.py::head_commit` already records: a two-dot `base HEAD` reports commits that landed on base since we branched as deletions of ours, so it returns the UNION of both sides and every gap would look like an overlap.

## The manifest exclusion is not hypothetical

You called this out in the issue and you were right to. Measured live while writing the change, on my own `fix-128` branch:

```
commits behind main: 3

base gap files (HEAD...origin/main)      my branch files (origin/main...HEAD)
  CHANGELOG.md                             RELEASE_MANIFEST.txt
  desktop/main.mjs                         src/no_human/core/orchestrator.py
  desktop/updater.mjs                      src/no_human/core/worktree.py
  ... 24 files                             tests/test_attempt_venv_isolation.py
  RELEASE_MANIFEST.txt                     tests/test_egress_allowlist.py
```

The intersection is `RELEASE_MANIFEST.txt` and nothing else. Without `GENERATED_PATHS` that branch would rebase on a gap it shares nothing real with, and since the manifest changes on nearly every landing and nearly every branch, so would almost everything else. The threshold would be undone quietly rather than loudly.

## Verification

Real git repositories, not stubs, because the risky part here is the range syntax rather than the set logic.

**Your plant and your control, both pinned:**

| case | behind | overlap | rebases |
|---|---|---|---|
| one commit touching a file the branch touches (the plant) | 1 | `src/app.py` | yes |
| one commit touching a disjoint file (the control) | 1 | none | no |
| five disjoint commits (today's rule) | 5 | none | yes |
| both sides touched only the manifest | 1 | none | no |

**Three mutations, each red on exactly the right tests:**

```
decide on the bare count again    -> AST wiring test fails
GENERATED_PATHS = frozenset()     -> both manifest tests fail
range switched to two dot         -> the CONTROL fails
```

The third is the one I care about most: with a two-dot range, a *disjoint* gap reads as an overlap, so the control flips and it rebases on everything. That is exactly the "most gaps are noise" decision the threshold was bought with, so it is the failure mode worth a dedicated test rather than a comment.

**Scope, per your ground rule.** `_refresh_stale_base` is called once per attempt from `_run_attempt` before the coder starts, so this widens only the pre-review rebase. The 2026-08-13 incident where rebase rounds under an already-reviewed PR looped until killed runs through `_MECHANICAL_FEEDBACK_SOURCES` round accounting, which is a different mechanism and is untouched.

**Never-fail contract preserved end to end.** An unmeasurable base returns `[]` rather than raising, matching `commits_behind`'s own convention, so a measurement failure degrades to today's count-only behaviour instead of rebasing on a guess.

## Suite

`tests/test_base_staleness_overlap.py` 11 passed. `test_structural_budget.py` and `test_egress_allowlist.py` clean, so the frozen budgets hold and `core/orchestrator.py` is edited entirely in place, still 21788 lines.

`tests/test_vcs.py` gives 21 failed / 82 passed here, and the same 21 **by name** on unmodified `origin/main` in this same checkout (diff empty). Those are Windows host failures unrelated to this change.

One note on that comparison, since it nearly misled me: I had a baseline of 12 for that file, measured against `93e87c5`. Main has since landed its own `manifest_repair` change, which took it to 21. Re-using the old number would have shown 9 phantom regressions that were not mine. Baselines expire when main moves; this one was re-measured against current main.


Sreekant13:
CI caught two things I had not run, both mine, and the first one is the one worth flagging.

**`tests/test_retry_base_staleness.py` already exists** and already drives the real `_refresh_stale_base`, asserting `task.context['base_staleness']` by exact equality. My new `overlapping_files` key broke it. I should have read that file before writing a line of this, and I did not.

Fixed, and it now also carries the integration case for this change rather than only my own unit file: a 1 commit gap whose file collides with the branch's own is rebased through the real orchestrator, while the disjoint 4 commit case sitting right beside it still is not. That pair is a better guard than anything in my new file, because it is the actual decision path.

**Second:** `docs/security.md` cites `vcs/git.py` by symbol AND line, and the 25 lines I added pushed two citations out by exactly 25. `scripts/reanchor_citations.py --apply` re-anchored the doc and the mirrored `CITATION_TABLE` together.

Re-ran everything CI flagged: 175 passed, 12 skipped. Force pushed.

Sreekant13:
Rebased onto current main. The conflict was `RELEASE_MANIFEST.txt` and nothing else, so I regenerated it rather than hand merging a generated file; `--strict` clean, `reanchor_citations.py` VERDICT=OK, 175 passed / 12 skipped.

Worth recording precisely, because it is tempting to overclaim it as evidence for this PR and it is not quite that.

This branch is 1 commit behind and the intersection with the base gap is `RELEASE_MANIFEST.txt` alone. Under this PR's rule that is the CONTROL, not the plant: `GENERATED_PATHS` excludes it, the overlap is empty, and `should_rebase` correctly says no. That is the right answer. What I hit was git being unable to textually auto-merge a generated file, which is a mechanical repin, not the re-implement plus re-review plus re-test round the issue is about.

So this is not a case the PR would have prevented. It is a second, cheaper data point for the same exclusion: the manifest intersects nearly everything, and if it counted, this branch would have been rebased on a gap it shares no real work with.

eyalgolan (dismissed):
Approving, thanks Sreekant, this is the shape I was hoping for. On a trial merge with current main: 295 passed across the two staleness files, test_vcs, test_readme_claims, the structural budget and the egress allowlist. The two-dot plant flips the control and the disjoint-gap test exactly as you described, and the bare-count plant fails the AST wiring test plus the new integration case in test_retry_base_staleness. orchestrator.py stays in place at main's count. `GENERATED_PATHS` carrying only the manifest is right, and the `overlapping_files` field in the record is the thing I would have asked for if you hadn't added it.

One thing before the button: main moved again under you (5760cef4f), and this time the conflict is in tests/test_readme_claims.py rather than the manifest. Main re-anchored the `merge_stack_run` row to 2931 while your branch re-anchored the `GitRepo.fetch` row to 1128. Keep both, regenerate the manifest, and `reanchor_citations.py` reports VERDICT=OK; that is what I got resolving it that way here. The rebase will dismiss this approval, so ping me and I'll put it back, it's a ten-second look.

And thanks for the note about the manifest-only intersection being the control rather than the plant. That is exactly the distinction.

Sreekant13:
Rebased onto 5760cef4f. Resolved exactly as you described: kept `GitRepo.fetch:1128` (mine, my git.py insertion moved it) and `merge_stack_run:2931` (main's), regenerated the manifest, and `reanchor_citations.py` reports VERDICT=OK. 175 passed / 12 skipped locally, strict inventory clean, orchestrator.py still in place at main's count.

The rebase dismissed your approval, so this is the ping. Ready when you are.

eyalgolan (approved):
Re-approved, thanks for the quick turnaround. I checked the rebased head rather than taking the ping on faith: it's a clean rebase (the only differences from your previous head outside the manifest and the claims table are main's own changes), both citation rows are where we agreed, VERDICT=OK, the manifest is strict-clean at 1537 files, orchestrator.py is at main's count, and the same 295 tests pass with the two-dot plant still failing exactly the three it should. Once the Python check finishes it's yours to merge.

PR #152, opened 2026-09-08
No-op send-back resume, round 2: resume no-human/d83c7bb6 @ 019c9129 - validated status write, timestamp comparison parsed (not lexical across ISO-T vs SQLite formats), rule that covers the incident, guards observed by tests

eyalgolan:
## Acceptance criteria
- the incident timeline (send-back 00:06:05, consuming attempt failed post-review, zero-diff retry) seeded with production-format started_at lands awaiting_approval; same-day stale feedback and a first attempt with a PR still fail; a send-back resume without a PR still fails; no validate=False; no mechanical_round stamp; tests observe each guard alone
- full Python suite green with the whole summary line; strict OK; app.py 6140; _run_attempt growth minimal via delegation; PR body describes the rule

## Changes
Round 2's rule for landing a zero-diff send-back resume as "no changes needed"
compared `send_back_feedback[-1]["at"]` against the HEAD commit's **git
committer date** (`git log -1 --format=%cI`). That rule misses its own
incident (task 82644133): feedback arrived at `00:06:05Z`; the consuming
attempt committed at `00:19:40Z` and PASSED review; the next attempt produced
a zero diff on the same HEAD and should have landed "no changes needed" —
but `feedback_at (00:06:05) > head_at (00:19:40)` is false, so it didn't.
Git commit metadata was never the right signal: it records when code was
written, not when it was reviewed and accepted.

This PR replaces that rule with one keyed only on no_human's own audit trail
(`Orchestrator._send_back_resume_round` in `core/orchestrator.py`):

- the newest `send_back_feedback` entry must be **older** than the `"at"` of
  the last entry in `review_history`,
- that last review entry must have `"passed": True`,
- and its `"sha"` must equal the current `repo.head_sha()` — a PASS on a
  different commit (e.g. a stale/replayed history entry) must not count.

All timestamps are parsed with `_parse_iso` and compared as instants, never
as raw strings, so the comparison is correct across the `...T...+00:00` and
SQLite `... ...` timestamp formats that show up in practice.

`_land_no_changes_needed` was also hardened so it never leaves a fake open
`task_phases` row for a round where no tests actually ran: it opens the
TESTING phase, immediately closes it via `close_phase(task.id, "no_tests_run",
reason=...)`, and only then sets `AWAITING_APPROVAL` — with the
`update_attempt` / `merge_context` / `emit` writes ordered after both
`set_status` calls succeed, so a CAS-refused status write leaves no stray
"no changes needed" marker behind. An unreachable dead-code guard was
removed, and the new re-entry point is registered (with a stated
disposition) in both `REGISTRY` and `STOP_REGISTRY` in
`tests/test_resume_entry_registry.py`.

`core/task.py`, `api/app.py`, and `core/db.py` are untouched by this change.

## Corrected assumption
Round 2 answered "what does the acceptance criterion `app.py 6140` mean" by
assuming it was a line-count **ceiling** ("6140 lines or fewer"). That is
wrong: `app.py` is out of scope for this task and must stay **untouched**,
i.e. exactly 6140 lines before and after — not a target to stay under.
`wc -l src/no_human/api/app.py` reports `6140` unchanged by this diff.

## How I verified this
<details><summary><b>Tests</b></summary>

```
$ uv run pytest tests/test_e2e_orchestrator.py::test_send_back_resume_with_no_changes_returns_to_awaiting_approval -q
1 passed in 1.46s

$ uv run pytest tests/test_e2e_orchestrator.py -k "send_back or no_changes" -q
7 passed

$ uv run pytest tests/test_e2e_orchestrator.py -q
201 passed in 135.55s

$ uv run pytest tests/test_resume_entry_registry.py -q
7 passed

$ uv run pytest tests/test_structural_budget.py tests/test_readme_claims.py -q
151 passed, 12 skipped

OK: 1534 file(s) match RELEASE_MANIFEST.txt

$ wc -l src/no_human/api/app.py
6140 src/no_human/api/app.py
```

The empirical repro test
(`tests/test_e2e_orchestrator.py::test_send_back_resume_with_no_changes_returns_to_awaiting_approval`,
recorded in `.no_human/repro_tests.json`) fails against round 2's
git-committer-date rule and passes against this rule.

</details>

Full verification log: ~/.no_human/artifacts/709490926ae342f7999f51fb62b3bb9c/verification-attempt-1.md — `nh logs 70949092`.

---
_Opened by no_human (`no-human/70949092` → `main`). It never merges: review and merge this yourself, or run `nh approve 70949092`._


eyalgolan:
Superseded by later rounds of the no-op send-back resume lineage; final round is PR #166 (landing from that branch).

PR #156, opened 2026-09-08
Tag CI bundle launches as environment=ci, so environment=real means an install

### LINKED ISSUE #120: telemetry: CI installer smoke runs report environment=real (48 one-shot app_started/day from fresh instance ids)
Measured 2026-09-07 in PostHog (project 560622): 48 distinct instance ids from US GitHub-Actions IPs each sent exactly one app_started with environment=real in 36 h, app_version 0.2.0. environment() (src/no_human/telemetry.py) falls through to real because the installer smoke job runs the built bundle outside pytest and without a CI marker the bundle can see. They inflate the real-install count 1:1. Expected: those runs tag as ci (export the marker into the bundle's environment in the installer smoke steps, or detect the runner from the wheel/bundle launch), so environment=real means an install.

Sreekant13:
Closes #120.

## Why the marker probe cannot reach these runs

`environment()` recognises a runner by the platform's own marker (`_CI_MARKERS`), which works for anything running in the job's shell. It cannot work for a launch that does not carry those markers all the way to the process that emits the event, and a frozen bundle started through a desktop launcher is that shape. A bundle is also not a source checkout, so the `dev` branch cannot catch it either, and it falls through to `real`.

So this declares the environment rather than inferring it. `NH_ENV` is rule 1 in `environment()`, above the marker probe and above `PYTEST_CURRENT_TEST`, so it is correct no matter what strips the rest of the environment on the way to the bundle.

Four steps launch a built artefact and now set `NH_ENV: ci`:

| job | step |
|---|---|
| wheel | Install the wheel the way a user would, and run it |
| linux | The installed frozen server runs outside the repo |
| linux | Installed app, first run to board to quit |
| linux | AppImage, the same run, extracted |

**Per step, not workflow-wide, on purpose.** `NH_ENV` outranks `PYTEST_CURRENT_TEST`, so setting it globally would retag the entire pytest lane from `test` to `ci`.

## A falsified comment that is part of why this went unnoticed

Directly above the `app_started` call, `api/app.py` said:

> Opt-in telemetry (default OFF — record() no-ops without consent).

It is opt-**out** and default **on**. `config.py`'s defaults carry `telemetry.enabled: True`, and `telemetry.py`'s own module docstring says "Default ON". The comment reads as a guarantee that an unattended run cannot emit anything, which is exactly the assumption under which 48 events a day go unquestioned. Corrected in place.

## Verification

**Three tests pin the function**, including the defect itself: with every CI marker and `PYTEST_CURRENT_TEST` removed and no source checkout, `environment()` does return `real`. Also that `ci` has a stable sentinel id, since minting a fresh uuid4 per run is what made 48 runs read as 48 installs rather than one noisy one.

**Two tests pin the workflow**, which is the half that rots, because the fix lives in a file no other Python test reads. One asserts every step matching a known bundle launch sets `NH_ENV=ci`; the other asserts each known launch fragment is still present in ci.yml, so the first cannot pass vacuously.

**Both plants fire and name what broke:**

```
drop NH_ENV from the AppImage step -> "linux: AppImage — the same run, extracted"
rename a launch fragment           -> "ci.yml no longer contains 'renamed-acceptance.mjs'"
```

**Suite.** New file 7 passed. `test_telemetry.py` and `test_telemetry_environment.py` give 15 failures here and the same 15 **by name** on unmodified `origin/main` in this same checkout; those are Windows host failures unrelated to this change.

Because this edits `ci.yml`, I re-ran everything that reads it rather than only the telemetry tests: `test_ci.py`, `test_ci_network_step_bounds.py`, `test_ci_upload_assertions_not_line_ending_dependent.py`, `test_deselect_lists_agree.py`, `test_test_lanes.py`, `test_readme_claims.py` gave 268 passed / 12 skipped, and `desktop/packagedFiles.test.mjs`, which asserts ci.yml comment prose, gave 40 pass / 0 fail. `scripts/reanchor_citations.py` reports `VERDICT=OK`, so the `api/app.py` edit shifted no cited line.

## One thing I could not confirm, and did not pretend to

I could not identify from the source **which** launch drops the markers. Both acceptance launches pass `{ ...process.env, HOME }` into Playwright, and `desktop/server.mjs` spawns the server with `widenPath({ ...process.env, ...env })`, so on the face of it `GITHUB_ACTIONS` should survive to the Python process in those paths. You have the PostHog data and I do not, so rather than guess at the stripper I fixed it in the way that does not depend on knowing: an explicit declaration at the highest precedence rule, on every step that starts a built artefact.

If you would rather chase the specific mechanism first, say so and I will hold this. It is worth knowing, because whatever strips `GITHUB_ACTIONS` there is probably also stripping something else.


eyalgolan (changes requested):
Thanks Sreekant, and the corrected app.py comment alone was worth the PR. The precedence fix is right and both of your plants fire here. But the launch that produces most of these events is not in your list, and this one I can measure.

The images job builds Dockerfile.mcp and runs packaging/mcp_handshake_probe.py against it on every run, PR or push. The entrypoint starts uvicorn on `no_human.api.app:app`, and the lifespan fires `app_started`. Inside the container there is no CI marker (the probe's `docker run -i --rm` passes no env), HOME is /root so the dev branch does not catch it either, and telemetry is on with the PostHog default. I built the image from main and ran the lifespan inside it:

```
VERDICT lifespan record calls: [('app_started', 'real', True)]
```

With `-e GITHUB_ACTIONS=true` or `-e NH_ENV=ci` the same container says `ci`. In the day and a half around the issue's window I count 31 ci.yml runs, and that job ran on every one of them, so it accounts for most of the 48 by itself. The remainder is consistent with the linux acceptance launches you tagged (8 push runs, two launches each), which I cannot measure from here, so keep those.

The step's `env:` cannot reach the container on its own because the probe builds the `docker run` line itself. The smallest change I can see: the probe forwards `NH_ENV` when it is set in its own environment (`-e NH_ENV` with no value, docker copies the host's), the probe step gets `NH_ENV: ci`, and `mcp_handshake_probe.py` joins `_BUNDLE_LAUNCHES` so your workflow test covers it. If you would rather bake `ENV NH_ENV=ci` into Dockerfile.mcp, say so; it would also tag the registries' own health checks, which are not installs either, but that is a bigger call and I would want it as its own PR.

One small thing: the wheel step's `nh --version` and `nh --help` never start the app (I patched `record` and ran both, zero calls), so that tag is harmless but emits nothing. Keep or drop as you like.

Everything else checked out: 340 passed across the telemetry, ci and lane files on a trial merge with main, and the AppImage plant names the step.

Sreekant13:
You found the launch that actually produces these and I had missed it. Fixed.

`mcp_handshake_probe.py` now passes `-e NH_ENV` and the probe step sets `NH_ENV: ci`, and `mcp_handshake_probe.py` joins `_BUNDLE_LAUNCHES`. The bare form matters and the docstring says so: `-e NH_ENV` with no `=value` forwards the host's value only when it is set, so it cannot invent an `NH_ENV` outside CI.

I added a fourth test that pins the probe's argv over the AST, because the workflow test alone is not enough here: the step's `env:` only reaches the container because the probe forwards it, so deleting `-e NH_ENV` leaves the workflow test green while the container goes straight back to tagging itself `real`. The plant reports `['docker', 'run', '-i', '--rm']`.

That plant also caught a mistake in my own test on the first run: it failed with a `NameError` because `ast` was not imported at module scope, so it was failing for the wrong reason and would have looked like a working plant. Fixed and re-planted.

Not taking the `ENV NH_ENV=ci` in Dockerfile.mcp option, for the reason you gave: it would tag the registries' health checks too, and that is a separate call. I noted that in the commit message so the next person does not have to re-derive it.

Keeping the wheel step's tag. You are right that `nh --version`/`--help` emit nothing today; it is one line and it is correct the day that step grows a launch.

Gate: 302 passed across the telemetry, ci, lane, citation and budget files, plus `packagedFiles.test.mjs` 40 pass / 0 fail and `reanchor_citations.py` VERDICT=OK. The 5 failures in `test_telemetry_environment.py` are identical by name to unmodified main on this Windows host.

eyalgolan (approved):
Approving, thanks Sreekant, and thanks for catching your own plant failing for the wrong reason; that is the kind of thing that only shows up if you actually watch it go red.

I checked the forwarding against the real image rather than the argv alone. With `NH_ENV=ci` on the host, `docker run -i --rm -e NH_ENV …` gives `environment() -> ci` inside the container; with it unset on the host the container sees no `NH_ENV` at all and says `real`, so the bare `-e NH_ENV` cannot invent a value outside CI, exactly as your docstring says. On a trial merge with current main: 364 passed across the telemetry, ci, lane, citation, budget and egress files; deleting `-e NH_ENV` from the probe fails your AST test naming the argv, and deleting the step's `env:` fails the workflow test naming `images: initialize + tools/list, inside the container`. The two failures in the mcp test files on my Mac are the same two by name on unmodified main (a missing local SDK module), nothing of yours.

Keeping the wheel tag is fine. Yours to merge once the branch is updated; if the update dismisses this, ping me and I'll put it back.

Issue #114, opened 2026-09-06
Attach net-new type-checker diagnostics as review evidence; measure before adopting LSP navigation


eyalgolan:
no_human's review gate runs on machine-checkable signals: tests, held-out tests, the tamper guard, the repro gate, ruff lint evidence, wiring evidence (see [docs/verification.md](https://github.com/no-human-ai/no_human/blob/main/docs/verification.md)). One signal that every serious agent harness has converged on is missing from that list: type/compile diagnostics. This issue is the plan for adding it, written after a deep dive into what actually works, because "integrate LSP" hides two very different features with very different evidence behind them.

**Diagnostics as a verification signal is the proven half.** SWE-agent's ablation credits its edit-loop linter (rejecting syntax-breaking edits) with roughly +3 points on a 12.47% baseline — the single most-cited agent-computer-interface feature of that paper (NeurIPS 2024). Aider lints after every LLM edit and feeds errors back for a fix round, by default. Claude Code's native LSP support (CLI v2.0.74) documents automatic post-edit diagnostics as the headline benefit — the model notices the error it just introduced and fixes it in the same turn. Copilot agent mode iterates on VS Code diagnostics. FeedbackEval (arXiv 2504.06939) puts numbers on both the value and the ceiling: compiler feedback alone resolves the shallow failure class (~35% on SWE-bench-verified), and mixed with test feedback reaches ~56% — diagnostics compose with tests, they don't replace them.

**Navigation tools are the unproven half.** Three years into the LSP-for-agents ecosystem, there is no published controlled experiment showing that an LSP navigation toolset (definition/references/hover as agent tools) improves resolve rates. Serena — the flagship, ~29k stars — explicitly declines to produce one. An independent industry benchmark (ManoMano's Project Aegis) found Serena ~4x more expensive and 60% slower on search-shaped tasks, paying off only on deep cross-file refactors. Instrumented sessions show frontier models invoking LSP for ~1% of navigation calls even when the tools are present, defaulting to grep. And the two most performance-obsessed harnesses both chose non-LSP structure: Cursor built embeddings, Aider built tree-sitter repo maps — Aider's stated reason being exactly the deployment burden (users must install and keep alive a per-language server) that we'd be signing up for.

[...]

PR #164, opened 2026-09-08
Attach net-new type diagnostics as review evidence, diffed against the merge base

PrinceXDev:
## What this changes

Adds net-new type-checker diagnostics as a third deterministic input to the review context, beside ruff lint evidence and wiring evidence. Phase 1 of #114.

If the repo under review configures a type checker (`pyrightconfig.json` / `[tool.pyright]`, `[tool.mypy]` / `mypy.ini` / `setup.cfg [mypy]`, or `tsconfig.json`), the same checker runs with the same argv over the whole project twice — once at the merge base, once at the reviewed commit — and the base result is subtracted from the after one. A repo with 400 pre-existing errors reports `net-new: 0`, not 400. No `merge_policy` change: the diagnostics go to the reviewer and the human, who draw the conclusion.

Design points, each avoiding a specific defect rather than adding a feature:

- **Not scoped to changed lines**, unlike lint evidence. The characteristic net-new type error appears at a call site the diff never touched — narrow a parameter and every caller lights up. A changed-line filter would discard exactly the diagnostics worth having.
- **Fingerprints ignore line numbers.** One inserted import shifts everything below it; a line-sensitive key would read a whole file as newly broken. `_fingerprint` keys on (path, code, digit-normalised message) as a multiset.
- **Silence means "did not run", never "clean".** A missing binary, a crashed checker, unparseable output, a worktree that will not build, or a base environment poorer than the after tree's all yield `ran=False`, which renders no section. A run that happened and found nothing renders `net-new: 0`, which is a different and usable fact.
- **Neither run touches the tree under review.** Both sides run in a throwaway `git worktree` at their own commit. The collector sits inside the window the orchestrator brackets with `reviewer_worktree.snapshot` / `.compare`, so a checker writing `.mypy_cache/` or a `*.tsbuildinfo` into the attempt's tree would make that compare report an added path and the orchestrator charge `reviewer_wrote` — discarding a real verdict for a write the reviewer never made. It also means neither side sees untracked files, which could only ever read as net-new.
- **Coverage is reported, not assumed.** A worktree carries no installed dependencies, so `tsc` and `pyright` on a project with real dependencies analyse it with those symbols degraded to `Any` — symmetrically, so the subtraction stays sound while its coverage does not. `TypeEvidence.unresolved_imports` is rendered as a `COVERAGE LIMIT` line so `net-new: 0` can never read as an unqualified clean. `mypy` resolves from no_human's own environment and is largely unaffected.
- **Our binary, never the repo's.** Resolved from no_human's own PATH, and a resolution landing inside the repo under review is refused.

Cost is bounded by one deadline for the whole collection — every checker, both sides — rather than a per-run cap, whose real worst case was checkers × sides × the cap. The collector runs off the event loop via `asyncio.to_thread`, and the single-turn route skips it entirely, since that route exists to spend less on a small diff. The route decision therefore moves above the collection, because it is now an input to it.

Two helpers were extracted from `reviewer.py` (`_collect_gate_evidence`, `_evidence_section`). That is not tidying: it keeps both `AdversarialReviewer.review` and `_build_review_prompt` under the 300-line function threshold instead of adding two new `FROZEN_FUNCTION_LINES` entries.

A repo that configures no type checker renders no section and spawns no subprocess, so its review prompt is byte-identical to before this change.

## Egress

`_run_checker` builds argv from `shutil.which`'s resolution, so the scanner cannot name the program and the channel is `exec:<dynamic>`. Declared against `review/type_evidence.py` in `tests/test_egress_allowlist.py`, naming what each checker reaches: nothing for `mypy` and `tsc`, and nodejs.org + registry.npmjs.org for `pyright` resolved to the PyPI launcher, which downloads a node runtime and the npm package on its first run.

The argv is dynamic deliberately. On Windows the npm checkers are `tsc.cmd` / `pyright.cmd` and `CreateProcess` does not apply `PATHEXT`, so a bare `["tsc", …]` raises `FileNotFoundError`, which the collector read as "not installed" — silently disabling itself on the platform this product ships a bundle for. Found by running a real `tsc`; the mocked tests structurally could not see it.

The module docstring also carries the `mypy` `plugins =` ceiling: mypy imports the modules that line names, read from the config of the repo under review, so this is the first time the review half can execute code the reviewed repo wrote. The harness already runs that repo's tests, so it is not a new trust boundary — but it matters before this is ever promoted from advisory to a required gate.

## Evidence

Run on Windows with the venv's interpreter and `.venv/Scripts` on PATH; `uv` is not installed on this machine, so the commands are not spelled `uv run pytest`.

```
$ pytest tests/test_type_evidence.py -q -p no:randomly
57 passed in 27.74s

$ pytest tests/test_egress_allowlist.py -q -p no:randomly
22 passed
```

Exactly one CI lane, which is what `tests/test_test_lanes.py` protects:

```
$ pytest tests/test_type_evidence.py --collect-only -q -m "not slow and not nightly" | tail -1
57 tests collected

$ pytest tests/test_type_evidence.py --collect-only -q -m "slow or nightly" | tail -1
no tests collected (57 deselected)
```

The File inventory job, real script, on a true LF tree:

```
OK: 1556 file(s) match RELEASE_MANIFEST.txt
```

`scripts/precommit_manifest_gate.py` exits 0, and `scripts/check_context_files.py --claude CLAUDE.md` prints `VERDICT=OK`.

**Zero regressions**, established by diffing failure sets between the pre-fix commit and this branch, both extracted as matched LF checkouts, over every `tests/*review*|*evidence*|*prompt*` suite:

```
pre-fix:      13 failed, 916 passed, 6 skipped
this branch:  13 failed, 926 passed, 6 skipped
$ diff failures_base.txt failures_mine.txt
(identical)
```

916 + 10 new = 926. The 13 shared failures are pre-existing Windows-environment failures (POSIX `chmod` and symlink privileges, `gh` absent, and a CRLF-checkout blob mismatch on a file this PR does not touch).

### The four fixtures #114 asks for

The first two run against a **real `tsc` 5.4.5**, not recorded output (`test_end_to_end_against_a_real_tsc` and its control, both `skipif` when tsc is absent):

- one introduced type error → `net-new: 1` at `a.ts:1:44 TS2322 Type 'number' is not assignable to type 'string'`;
- the same error present at base and pushed down two lines by the diff → `net-new: 0` (a line-sensitive fingerprint reports 1 here);
- a crashed checker → no evidence attached, a `WARNING` log line, and `test_a_crashed_checker_makes_no_cleanliness_claim_in_the_prompt` proving the review prompt asserts nothing about types in that case;
- no Claude credential needed for any of it.

The reviewer-integrity fix is pinned as an artifact, not an argument: `test_a_checker_that_writes_leaves_the_review_tree_untouched` uses a fake checker that really does create `.mypy_cache/missing_stubs` where it runs, then asserts the reviewed tree is byte-for-byte unchanged. Reverting the after run to `repo_path` turns five tests red.

### How often this actually renders

Detection across 130 local git repos, read-only, no subprocess:

- would render: 32 (25%)
- of those, `tsc`-only: 30 (94%)
- `mypy` + `tsc`: 2
- this repo: never — it configures no type checker

Real runs, against clones so nothing wrote to the originals:

- 2-commit TS repo — 3.0s, net-new 327, after_total 337, unresolved 41
- 1995-commit TS repo — 59.0s, net-new 2, after_total 21058, unresolved 7359 (35%)
- small TS app — 1.7s, net-new 0, after_total 18, unresolved 11 (61%)

So the section renders on roughly a quarter of repos, 94% of those are the degraded TS path, and a third to two thirds of what it reports there is unresolved-import noise. The `COVERAGE LIMIT` line fired on all three. The 59s case shows the single collection budget doing real work rather than being theoretical.

Two limits on that sample: it is one developer's machine and skews heavily to TS frontend work, so 25% is not a population figure. And the clones carry no `node_modules`, so both sides were degraded — which matches what worktree-based runs now produce, but it does not isolate the pre-fix asymmetry.

### Not covered

`mypy` and `pyright` were not available on this machine, so their parsers are covered by recorded-output tests only; the `tsc` path is the one validated end-to-end against a real binary.

## Notes for review

`tests/test_structural_budget.py` raises the `review/reviewer.py` freeze to 3029 (from 2945, in two steps, each with a comment in the existing house style). That is the documented ratchet, not a weakened test — no test or assertion was removed anywhere, and the change is net +57 tests.

`docs/verification.md` gains a section. Its code references use the plain prose form the rest of that file uses rather than the `::` symbol form: I checked empirically by breaking a citation, and `verification.md` is not in `CITATION_TABLE`, so a `::` citation there would look verified without being so. The same file previously said the second result is subtracted from the first, which is backwards; the code was right and the prose is now fixed.

## Checklist

- [x] Lockfiles committed if a dependency changed. (No dependency changed — the module is standard library only.)
- [ ] First PR only: I have read `CLA.md` and added my `contributors/<handle>.md` entry to this PR.


eyalgolan (changes requested):
Hi Prince, welcome, and thank you for taking on the first phase of #114 with this much care. I approved the CI run, and I read the whole thing rather than skimming, because a lot of it is good: the parsers handle Windows paths, `note:` lines and digit-heavy messages correctly, the "silence means did not run" contract is the right one, the multiset over line-agnostic fingerprints is right, and the two extracted helpers are byte-identical in effect (I hashed `_build_review_prompt`'s output for five argument shapes on main and on your branch, same digests). Your tests have teeth: four mutations I planted (line number in the fingerprint, a crashed checker returning an empty list, comparability forced true, the type section dropped from the prompt) each go red on the tests you'd expect. CLA file, lanes, manifest and the budget comment all check out.

Three things stand between this and landing, and the first is what CI is showing you.

**1. The egress gate is red.** `tests/test_egress_allowlist.py::test_no_undeclared_external_egress` fails on your head and on a trial merge with main:

```
src/no_human/review/type_evidence.py:506  exec:<dynamic>  (argv is built at runtime; the program cannot be named)
1 failed, 11275 passed, 449 skipped in 605.74s
```

Every exec or network channel in this repo is declared in that file with a gate, and this one deserves a real entry rather than a classification, because the PyPI `pyright` package is a launcher that downloads node and the npm package on first run. If `shutil.which("pyright")` finds that wrapper, a review would make a network call. Say so in the entry.

**2. The checker writes into the tree under review, and the integrity guard punishes the reviewer for it.** `_collect_gate_evidence` runs inside `reviewer.review`, which sits between `reviewer_worktree.snapshot` and `reviewer_worktree.compare` in the orchestrator. Your mypy argv has no `--cache-dir`, so mypy's default drops `.mypy_cache/` into the attempt's worktree, and `tsc` with `incremental` or `composite` writes `*.tsbuildinfo`. When the repo doesn't gitignore those, the compare reports an added path, the orchestrator emits `reviewer_wrote` (blaming the reviewer, which didn't do it), reverts, and replaces the real verdict with an integrity failure at full reviewer cost. I reproduced the delta both ways: an un-ignored `.mypy_cache/missing_stubs` shows up as `added`; the gitignored case is clean. Point the caches outside the repo, or run the after check in a throwaway worktree the way you already do for the base; the second is the symmetric fix and it also stops untracked files that exist only in the after tree from being counted as net-new.

**3. The cost is unconditional and the TypeScript half can't fire.** The base worktree carries tracked files only, so it has no `node_modules`; every bare import there is `TS2307`, which is in your unresolved set, so `_environments_comparable` says no and any TS project with a single dependency gets nothing after paying two whole-project `tsc` runs. Your two real-tsc tests use a zero-dependency repo, which is why they can't see it. mypy resolves third-party imports from its own environment so it degrades equally on both sides and does work; pyright resolves from the project's venv and will mostly behave like tsc. Meanwhile the collection is a plain synchronous call in a coroutine, up to two checkers times two runs at 90 s each, so a bad case adds six minutes and blocks the event loop for everyone, and it runs before `route_single_turn` is decided, so a small diff on the cheap route pays the full price. I'd like: one budget for the whole collection, `asyncio.to_thread`, skip it on the single-turn route, and the TS ceiling stated plainly in the PR body and in `docs/verification.md`. #114 is explicit that measurement comes first, so a number for how often the section actually renders on a real repo would carry this a long way.

Smaller, same pass if you're in there: the `git worktree remove` in the `finally` is the one subprocess call without a timeout, and a `git worktree prune` after it would keep a killed run from leaving a registration behind; `docs/verification.md` says the second result is subtracted from the first, and the code does the reverse (`Counter(after) - Counter(base)`, which is the right one); and a line in the module's ceilings that mypy will import a `plugins =` module from the repo's own config, since the gate didn't execute repo-authored code before this.

None of this is a design rejection. The shape is the one the plan asks for; it's the cost accounting and the two ways it can hurt a review that need closing first. Happy to look again as soon as you push.

PrinceXDev:
Thank you @eyalgolan  — the mutation testing and the `_build_review_prompt` digest check were more than I expected, and all three blockers were real. Two of them I had reasoned about the collector in isolation instead of about where it sits in the orchestrator, which is exactly where they bite.

**1. Egress.** Declared against `review/type_evidence.py` as `exec:<dynamic>`, `default-on:`. The entry names what each checker reaches — nothing for `mypy` and `tsc`, and nodejs.org + registry.npmjs.org for `pyright` resolved to the PyPI launcher, on its first run only. Writing it exposed a contradiction: the module docstring claimed "Nothing is ever installed", which is false precisely in that case. It now distinguishes "no_human installs nothing itself" from "nothing is installed", and carries your `plugins =` point as a ceiling as well.

For the record on why the argv is dynamic rather than three named entries: `shutil.which` resolution is load-bearing. On Windows, the npm checkers are `tsc.cmd`/`pyright.cmd` and `CreateProcess` does not apply `PATHEXT`, so a bare `["tsc", …]` raises `FileNotFoundError`, which the collector read as "not installed" — silently dead on the platform we ship a bundle for. Running a real `tsc` is what found it; the mocked tests could not, structurally.

**2. Integrity.** Took the symmetric fix. `_run_at_commit` now serves both sides, so nothing the checker writes can reach the attempt's tree, and untracked files can no longer be net-new. Pinned by `test_a_checker_that_writes_leaves_the_review_tree_untouched`, which uses a fake checker that really does create `.mypy_cache/missing_stubs` where it runs — your reproduction, inverted into a guard. Reverting the after-run to `repo_path` turns five tests red.

The cache became sha-keyed over both sides as a side effect, so a re-review of an unchanged head now costs zero runs.

**3. Cost.** One deadline for the whole collection (each run draws what is left; a run that cannot start does not), `asyncio.to_thread`, and skipped on the single-turn route — which moved the route decision above the collection, since it is now an input to it. Lint and wiring stay on both routes.

Your TS point was right, and making the runs symmetric changed its shape rather than fixing it: both sides are now degraded equally, so `_environments_comparable` *passes* and the subtraction proceeds over two blind analyses. Arithmetically sound, but `net-new: 0` would have been a false clean — the one thing this module exists not to emit. `TypeEvidence.unresolved_imports` is now rendered as a COVERAGE LIMIT line stating how much the checker could not see.

**Measurement.** 130 local git repos, detection only, read-only:

- would render: 32 (25%)
- of those, `tsc`-only: 30 (94%)
- `mypy`+`tsc`: 2
- this repo: never — it configures no checker

Real runs, on clones so nothing wrote to the originals:

- 2-commit TS repo — 3.0s, net-new 327, after_total 337, unresolved 41
- 1995-commit TS repo — 59.0s, net-new 2, after_total 21058, unresolved 7359 (35%)
- small TS app — 1.7s, net-new 0, after_total 18, unresolved 11 (61%)

So: renders on a quarter of repos, 94% of those the degraded TS path, and a third to two thirds of what it reports there is unresolved-import noise. The COVERAGE LIMIT line fired on all three. 59s also shows the single budget doing real work rather than being theoretical.

Two limits on that sample, since it is the number you asked for: it is one developer's machine and skews heavily to TS frontend work, so 25% is not a population figure. And the clones carry no `node_modules`, so both sides were degraded — which matches what worktree-based runs now produce, but it does not isolate the pre-fix asymmetry you described.

**Smaller items.** All three done: `git worktree remove` is timed with a `prune` after it; the docs said the second result is subtracted from the first, which is backwards (the code is right); and the `plugins =` ceiling is in the module docstring and `docs/verification.md`.

**Verification.** Same review/evidence/prompt suites on matched LF checkouts of the pre-fix commit and this one: 13 failed / 916 passed → 13 failed / 926 passed, identical failure sets, +10 new tests. The 13 are pre-existing Windows-environment failures (POSIX chmod/symlink privileges, `gh` absent). `test_egress_allowlist.py` 22 passed. Structural budget re-frozen 3004 → 3029 with a comment.

eyalgolan (changes requested):
Prince, this is a strong round. All three blockers are properly closed, and the way you closed the second one is better than what I asked for — making both sides symmetric through `_run_at_commit` fixes the cache write and the untracked-file problem in one move, rather than patching each checker's cache flag. I re-ran the reproduction as a plant: pointing the after-run back at the reviewed tree turns six tests red, including the fake checker that really does write `.mypy_cache/missing_stubs`. That's your reproduction inverted into a guard, which is exactly right.

The part I want to single out is the one nobody asked you for. You noticed that making the runs symmetric didn't fix the TypeScript story, it changed its shape — both sides degrade equally now, so `_environments_comparable` passes and the subtraction runs over two blind analyses, and `net-new: 0` would have been a false clean. Catching that yourself, and answering it with a COVERAGE LIMIT line rather than quietly shipping the zero, is the judgement this codebase is built around. Dropping that line turns its test red, I checked.

And thank you for the measurement, including the two caveats about your sample. A quarter of repos, 94% of those the degraded TS path, a third to two thirds unresolved noise — that is a modest signal, and it's the honest number rather than a flattering one. My read is that it's worth landing as advisory evidence precisely because the coverage line now says so out loud; the TS environment question is a phase-two problem, not a reason to hold this.

Everything else I checked holds up on a trial merge with current main: manifest strict-clean at 1561 rows, VERDICT=OK, 57 tests in exactly one lane, egress green now, and the output is bounded at 40 diagnostics / 8 KB with a truncation line, which was my worry once I saw net-new 327 in your numbers. The prompt for a repo with no type checker is byte-identical to main across five argument shapes — same digests on both sides, so nobody without a checker sees any change. The one failure in my run, `test_scrubbed_fixtures_are_their_origin_blob_put_through_scrub`, fails by the same name on unmodified main here; it's a private scrub supplement this checkout doesn't have, nothing of yours.

Two things before it lands.

**It conflicts with main.** `tests/test_structural_budget.py` — main took `review/reviewer.py` from 2945 to 2979 in other work while you took it to 3029, so both comment histories need keeping and one measured value at the end. I resolved it that way here and the scanner's own metric reads **3063** on the merge result. Please measure it yourself on your rebase rather than taking my number; if we disagree, mine is the one that's wrong.

**The cost control is the one property with no test.** This is the thing I'd like fixed rather than noted. `_collect_gate_evidence` honours `with_type_evidence`, and your two route tests pin that properly. What nothing pins is the caller: `reviewer.py` passes `with_type_evidence=not route_single_turn`, and if I change that to `True` — putting the whole-project type check back on the cheap route — all 150 tests still pass. So the guard against the exact cost regression you just fixed is one edit away from silently coming back. The precedent is in the tree already: PR #145's `test_env_setup_actually_runs_through_the_isolated_environment` reads the call site over the AST for the same reason. Same shape here would do it, and it would have caught my mutation.

Nothing else from me. Push the rebase with that test and I'll take another look — this has been a genuinely good first contribution to review.

PrinceXDev:
Thank you @eyalgolan — and the note about `test_scrubbed_fixtures_are_their_origin_blob_put_through_scrub` saved me chasing a ghost; it does not appear in my 13 either.

**The call-site test.** Added as `test_the_call_site_actually_gates_type_evidence_on_the_route`, over the AST, following `test_env_setup_actually_runs_through_the_isolated_environment` — including its reason for not doing a substring search. It asserts the keyword is present at every `_collect_gate_evidence` call and that its value is specifically `not route_single_turn` rather than any constant.

I ran your mutation both ways. `with_type_evidence=True` fails only that test, 57 others green. Removing the keyword entirely so the default `True` stands also fails only that test. I re-ran both after the merge rather than before, because `reviewer.py` auto-merged and the call site moved to line 2500 — the guard still bites there.

**The freeze value.** 3063, and I measured it rather than taking yours. Three ways, all agreeing: `git merge-tree` before touching my tree, the additive check against the merge base (2945 + 84 + 34), and `wc -l` on the real merge result. Both comment histories are kept, labelled as two lines branching from the same 2945, with a closing entry recording that the number is measured on the merge result and never summed — and that where a sum and the scanner disagree, the scanner is right.

One thing your check could not have seen: main has moved 4 commits since, and one of them, `e44af234`, touches `reviewer.py`. So 3063 happens to still hold, but it is now measured against main including that commit rather than the one you merged against. Independent corroboration that we are looking at the same tree: my regenerated manifest is 1561 rows, the figure you reported.

**Verification on the merge result.** File inventory strict-clean at 1561, `VERDICT=OK`, precommit gate passes, 58 tests in exactly one lane and zero in the other, 235 passed across the guard suites.

Regression measured against `upstream/main` itself, both sides matched LF extracts, over every `tests/*review*|*evidence*|*prompt*` suite:

```
upstream/main:  13 failed, 899 passed, 6 skipped
this branch:    13 failed, 957 passed, 6 skipped
failure sets:   identical, diffed both directions
```

957 − 899 = 58, exactly the new tests. The 13 are pre-existing Windows-environment failures here — POSIX chmod and symlink privileges, `gh` absent, SIGKILL — and they are the same 13 on unmodified main.

**One thing I did not verify, rather than imply I had.** The Web board job. This machine is on Node 23 and CI pins that job to Node 20, so a local run would fail for a reason that is not real. The branch touches no file under `web/`, `desktop/` or `packaging/`, which is structural evidence, but it is not a green run and I would rather say so than round it up.

On the coverage line and the TS environment: agreed that it is a phase-two problem. I have kept the ceiling stated in both the module docstring and `docs/verification.md` so the limitation travels with the code rather than living only in this thread.

eyalgolan (dismissed):
Approving. Both items are closed and the whole thing holds up under the checks I could throw at it.

The call-site test is exactly right, and it bites both ways — I ran my own mutation and the one you'd reach for next. Setting `with_type_evidence=True` fails only that test, 196 others green; deleting the keyword so the default stands also fails only that test. The two earlier fixes still bite on this head too: pointing the after-run back at the reviewed tree turns six red, dropping the coverage line turns one red. On your head, which now contains main, the freeze reads 3063 measured and 3063 frozen, `--strict` is clean at 1561, VERDICT=OK on both the citations and the context files, 58 tests in one lane and none in the other, and CI came back green across the board — 11413 passed. Web board included, which was the one thing you said you couldn't run, so that's covered now.

Measuring 3063 three ways instead of taking my number was the right instinct, and you were right that my basis had already moved. That's the second time in this PR you've corrected something I asserted, which is a good sign for both of us.

One security note, not a blocker and not really about your change. The checker subprocess inherits the process environment with no `env=`, and mypy imports whatever a repo's `plugins =` line names, so a reviewed repo can run its own code with our credential in the environment. You documented that ceiling in the module and in verification.md, which is the right thing to do. It doesn't add an exposure class — `testing/runner.py` already runs the repo's own test command the same way — but it does add a location, since the review gate didn't execute anything repo-authored before this. Scrubbing the environment for repo-controlled subprocesses is a project-wide job and I'll open it separately rather than hang it on you.

The related control I did test, because that one is yours: I planted a `tsc` in `node_modules/.bin` inside the repo under review and put it first on PATH, which is the position someone opening a PR actually has. `_resolve_binary` refused it by name and accepted an identical binary outside the repo. Good.

Nice work. This was a big first contribution and you took three rounds of hard review without once arguing the finding instead of measuring it.

PrinceXDev:
Thanks — and thank you for planting the PATH attack against `_resolve_binary`. That was the control I was least able to test myself convincingly, since the interesting position is the one an outside contributor has and I only had my own machine.

On the environment note: agreed on all three points, including that it is a new location rather than a new exposure class. For the record of what is actually there, `_run_checker` passes no `env=`, so the checker inherits this process's environment exactly as `testing/runner.py` does for the repo's test command. Happy to take the scrubbing work on the separate issue when you open it — the collector is a small surface to convert, and I already have the ceilings written down, so it would be a reasonable place to start rather than finish.

Phase 2 is the per-edit `PostToolUse` hook. I will open it as its own PR off main once this lands rather than extend this branch.

eyalgolan:
Heads up before you look and wonder what broke: this now shows a conflict, and it's mine rather than yours. I landed two commits on main after approving (`da8d667d` and `a5069014`), and both touch `RELEASE_MANIFEST.txt`. That's the only conflicting file — I reproduced the merge here and no source file collides.

It's the standing tax on the manifest: it pins every shipped file by hash, so it changes on nearly every landing and on nearly every branch, and each merge to main re-conflicts every open PR on it. Nothing to hand-merge. Rebase or merge main, regenerate it with `check_release_manifest.py --write` (stage any new files first), and push.

That push will dismiss the approval under the last-push rule. Ping me and I'll re-run the gates on the new head and put it back — it'll be a quick pass, since I'm only re-checking that the regenerate is clean and nothing else moved.

Sorry for the churn.

PrinceXDev:
> Heads up before you look and wonder what broke: this now shows a conflict, and it's mine rather than yours. I landed two commits on main after approving (`da8d667d` and `a5069014`), and both touch `RELEASE_MANIFEST.txt`. That's the only conflicting file — I reproduced the merge here and no source file collides.
> 
> It's the standing tax on the manifest: it pins every shipped file by hash, so it changes on nearly every landing and on nearly every branch, and each merge to main re-conflicts every open PR on it. Nothing to hand-merge. Rebase or merge main, regenerate it with `check_release_manifest.py --write` (stage any new files first), and push.
> 
> That push will dismiss the approval under the last-push rule. Ping me and I'll re-run the gates on the new head and put it back — it'll be a quick pass, since I'm only re-checking that the regenerate is clean and nothing else moved.
> 
> Sorry for the churn.

No churn on my side — the manifest doing this is the design working, not a surprise.

Caught up. `RELEASE_MANIFEST.txt` was the only conflict, exactly as you said; every source file auto-merged, `tests/test_structural_budget.py` included, so the freeze needed no arithmetic — it reads 3063 measured and 3063 frozen on the new head.

Regenerated rather than hand-merged, staging new files first as you noted: the visible hunks named fewer paths than the regenerate corrected, which is the usual reason not to pick hashes here. Strict-clean at 1563 rows, VERDICT=OK on the context files, precommit gate exit 0, 58 tests in one lane and none in the other, 235 passing across the guard suites.

No code change in this push — catch-up merge only.

Ready for the re-run whenever suits.

eyalgolan (approved):
Re-approved. Thanks for turning that around quickly.

I checked that the merge brought in only my two commits and a regenerated manifest, rather than taking it on trust: every one of your own files hashes identically to the head I approved this morning, and every other file that moved traces to `da8d667d` or `a5069014`. On the new head, with main at `a5069014a`: strict-clean at 1563, VERDICT=OK on citations and context files, the freeze reads 3063 measured against 3063 frozen, 58 tests in one lane and none in the other, 1102 passed locally across the review and evidence suites, and CI is green all the way across — 11452 passed. Both guards still bite after the merge: the cost mutation fails only your call-site test, and the integrity one takes six down.

Standing caveat, and it's on me for not saying it the first time: an approval here is only clean against the main it was measured on. The manifest pins every shipped file by hash, so the next landing to main will re-conflict this branch on that file alone. If that happens before you merge, regenerate and push and ping me — it's a two-minute re-check, not a re-review.

Yours to merge whenever you're ready.

eyalgolan:
It's showing dirty again — main moved twice more since I approved (`3e8d518e` and `9b68ea97`). Same story as before and still nothing to do with your change: I merged your head onto the new main here and `RELEASE_MANIFEST.txt` is the only file that conflicts. The budget file auto-merged cleanly even though both sides touched it.

I re-ran everything on that merge result so you don't have to guess whether the approval still means anything: strict-clean at 1564, VERDICT=OK, the reviewer freeze still reads 3063 measured against 3063 frozen, and 1102 passed across the review and evidence suites. So the approval stands as-is.

All it needs is the usual: merge or rebase main, `check_release_manifest.py --write`, push. Ping me after and I'll put the approval back — genuinely a two-minute check, since I've now verified the content twice against two different mains.

If it goes dirty a third time before you get to it, that's on the repo rather than on you. I'm looking at a merge queue to stop this happening to contributors.

eyalgolan (changes requested):
Prince, I have to walk back my approval, and I'm sorry — this is my miss, not a moving target. I was about to land it, ran the independent pre-merge review the project requires before anything reaches main, and it found a real false clean that I had not thought to test for. I reproduced it myself before writing this.

**A checker whose config scopes it can report `net-new: 0` over a diff it never analysed.**

`include` in `tsconfig.json`, `include` in `pyrightconfig.json`, `exclude` under `[mypy]` — all ordinary, and in a monorepo they're the norm. Nothing in the collector asks whether any changed file was inside the checker's scope. Fixture: `tsconfig.json` is `{"include":["packages/a"]}`, the entire diff is in `packages/b`, and the diff introduces a genuine type error there. Both runs happen, both succeed, the subtraction is arithmetically correct, and this is what the reviewer is handed:

```
ran=True checker=tsc net-new=0 after_total=2 unresolved=0

TYPE EVIDENCE (tsc, deterministic): net-new type diagnostics introduced by this diff: 0.
  Computed by running tsc over the whole project at the merge base and at this tree and
  subtracting; the after tree carries 2 diagnostic(s) in total, so a pre-existing error is
  not counted here.
  Evidence, not a verdict: a type checker cannot see a defect the types already permit.
```

Every guard you built is satisfied — the runs aren't degraded, so no COVERAGE LIMIT line fires; `after_total: 2` says the checker did see this tree; and that last sentence reads as reassurance about a file nothing looked at. Then `reviewer.py`'s READING SCOPE tells the reviewer it already has the type evidence and needn't re-derive it. That is the one outcome your own module docstring says is unacceptable, and it lands hardest on the TypeScript path, which is 94% of where you measured this firing.

The fix looks small: `collect_type_evidence` already computes the changed paths. Either say how many changed files were in the checker's scope, or when no diagnostic comes from any changed path, say that 0 is a claim about the configured scope rather than about the diff.

Two smaller things while you're in there:

- **The thread hop isn't pinned.** Your AST test guards `with_type_evidence`, which is the half I asked for, but removing `asyncio.to_thread` around the collector entirely leaves 151 passed and nothing red. That's the other half of the round-3 fix and it can come back silently — worth extending the same test.
- **The timeout may not be the cap the docstring claims.** The second reviewer measured `_run_at_commit` passing the same budget to `git worktree add` and then again to the checker without re-deriving what's left, so one call can spend roughly double; they measured 1.72x on a fixture with a slow checkout. I did not reproduce this one myself, so treat it as a lead rather than a finding — but a `min(remaining, ...)` between those two calls is cheap either way, and the test that should catch it asserts the timeout argument rather than elapsed time.

Nothing here is a design problem. The worktree symmetry, the fingerprint, the coverage line and the binary refusal all held up under a fresh adversarial pass, including a positive control with a real executable planted inside the repo. It's the scope question that needs closing.

Sorry for the approve-then-retract. You did the right thing at every round and the miss was mine.

PR #177, opened 2026-09-08
Drop the out-of-brief _unjudged_checkpoint_head flag; retarget two old tests to the WIP-routing rule

eyalgolan:
> **Review not passed** (1 round)

## Evidence
| Check | Result |
|---|---|
| Independent review | ❌ **not passed** — 1 round |

<details><summary>2 findings raised and addressed across review rounds</summary>

- Suite red: two honesty-gate tests regressed — AC requires 'uv run pytest green ... at least as many tests as before'. I ran the two tests from the pasted failur
- Production data-integrity regression on reviewer:null + advisory — The failing runs show 'advisory: draft PR before review skipped' then final status AWAITING_A

</details>

## Acceptance criteria
- grep -rn _unjudged_checkpoint_head src tests returns nothing; _run_review no longer raises ReviewerUnavailable for a routed checkpoint head; _route_unjudged_head sets no instance state
- test_a_revision_branch_sitting_on_an_abandoned_partial_is_not_credited and test_a_machine_resume_is_not_credited_as_human_gated still exist, use a FakeReviewer that fails the routed diff, and assert the attempt is not delivered (no PR, not AWAITING_APPROVAL); their docstrings describe the routing rule
- every round-3 test in tests/test_wip_checkpoint_routed_to_review.py and tests/test_server_stop_checkpoint.py passes unchanged; _already_satisfied_eligible and _head_is_wip_checkpoint are unchanged
- branch rebased on current origin/main; uv run pytest -q -n 4 -p no:cacheprovider green with the whole summary line pasted and at least as many tests as before; scripts/check_release_manifest.py --strict prints OK; wc -l src/no_human/api/app.py prints 6140

## Changes
I'll pause here and wait for the background test run notifications before proceeding further.

- **Q:** If removing the flag-setting code leaves `_route_unjudged_head` empty or with only trivial logic, should we delete the entire function or retain it in the codebase? **A:** HUMAN-GATED: not self-answerable
- **Q:** Should the FakeReviewer in the two rewritten tests fail by returning a rejected review result, raising an exception, or using another failure mode? **A:** The FakeReviewer should return a rejected review result (e.g., a Review object with approved=False or an equivalent rejection marker) rather than raising an exception. This allows the routed diff to proceed through _run_review's normal rejection path without special exception handling, validating that the routing rule works and the reviewer's judgment prevents delivery. _(assumption)_
- **Q:** Should the 'not delivered' assertion verify both conditions (`pr_id is None` and `status != AWAITING_APPROVAL`), or just one? What other attempt state properties must we verify? **A:** Verify both conditions: pr_id is None AND status != AWAITING_APPROVAL. These are the two primary indicators that an attempt was not delivered into the release pipeline. Additionally verify that the attempt itself exists and is associated with the task (attempt is not None, attempt.task_id matches), to confirm the routing executed and the reviewer's judgment was applied rather than the attempt bein _(assumption)_
- **Q:** If measured `_run_review` metrics after removal don't match trunk values (394 lines / CC 74), where and in what format should we document the measured values—inline comment, docstring, or elsewhere? **A:** Document measured values with an inline comment immediately above or after the function signature of _run_review (e.g., '# Lines: 395, CC: 75 (after _unjudged_checkpoint_head removal)'), following the same style as the existing code metrics cited in the task description. This keeps the budget visible at the function's definition point without requiring docstring changes. _(assumption)_

</details>

## Superseded PRs
- https://github.com/no-human-ai/no_human/pull/176 — earlier attempt, abandoned

## How I verified this
<details><summary><b>Tests</b> (7 runs, last shown) — <code>cd /Users/eyalgolan/.&lt;redacted&gt;/worktrees/b6f407257a33478cbcf48a92750fb17f.90097.567ff0f2 grep -c "passed=True, checklist=\[$" tests/test_resume_wiring_round2.py git diff --stat tests/test_resume_wiring_round2.py uv run pytest -q -p no:cacheprovider tests/test_resume_wiring_round2.py 2&gt;&amp;1 | tail -10</code></summary>

```
1
...................                                                      [100%]
19 passed in 20.24s
```

</details>

7 commands recorded while working. Full verification log: ~/.no_human/artifacts/b6f407257a33478cbcf48a92750fb17f/verification-attempt-2.md — `nh logs b6f40725`; the same log, every command with its captured output, is posted as this PR's **How I verified this** comment when posting succeeds. Whether it passed is in the **Evidence** table above: no entry here asserts a pass or a fail.

---
_Opened by no_human (`no-human/b6f40725-2` → `main`). It never merges: review and merge this yourself, or run `nh approve b6f40725`._


eyalgolan:
Landed on main as b9fb6c80 (identity-normalized squash of 212dc3b1 after independent review and a merge-result gate).

PR #195, opened 2026-09-09
Treat uv's --active as command-level intent, closing the coder half of #128

### LINKED ISSUE #128: A task attempt's env setup rewrites the shared .venv's editable path to its own ephemeral worktree, breaking every nh invocation once that worktree is gone

Sreekant13:
Closes #128, the coder half. PR #145 did the orchestrator half.

## The reproduction, and the control that shaped the fix

From inside a linked worktree whose inherited `VIRTUAL_ENV` names the primary checkout's venv:

```
$ uv run --active python -c "print('ran')"
Uninstalled 1 package in 3ms
Installed 1 package in 10ms
shared .pth   <primary>\src   ->   <worktree>\src

# the same command WITHOUT --active
$ uv run python -c "print('ran')"
shared .pth   UNCHANGED, and uv created <worktree>/.venv itself
```

Two things follow, and the second is why this is a small change rather than a large one.

**There is no install word in that command.** `uv run` syncs the project into its target environment before running anything, so `uv run --active ruff check .` rewrites a venv exactly as an install would. That is why a guard keyed on mutating subcommands could not see it: `run` is deliberately not in `_MUTATING_SUBCOMMANDS`, because `uv run pytest -q` must stay allowed.

**Plain `uv run` is not merely safe, it is already correct.** It targets the worktree's own environment and builds it if absent. So nothing without `--active` needs to change.

## What changed

`--active` is treated as two things, neither of which is a text pattern matched against the command:

1. **intent on its own**, without a mutating subcommand, because `uv run` syncs; and
2. **the signal that makes the inherited `VIRTUAL_ENV`/`UV_PROJECT_ENVIRONMENT` an effective target**, which the existing containment check then denies when it sits outside the worktree.

Point 2 is the part the residual register says must not be done, and I think the distinction it draws is exactly the one that applies here. The register rules out trusting a value that is **solely inherited**, because a review round measured that doing so denied `uv sync` and `uv run pytest -q` in every session. `--active` is not solely inherited: it is a command-level statement of intent to use that value, the same class of signal as the `VIRTUAL_ENV=<path> pip install` assignment the module already trusts. If you read that register as ruling this out too, say so and I will withdraw it rather than argue.

## Verification

| command | verdict |
|---|---|
| `uv run --active python -c "print(1)"` | denied |
| `uv run --active ruff check .` | denied |
| `uv sync --active`, `uv pip install --active -e .` | denied |
| `uv run --active=true pytest -q` | denied |
| `bash -lc 'uv run --active pytest -q'` | denied |
| `uv run pytest -q`, `uv sync`, `uv run ruff check .` | **allowed** |
| `uv run --active pytest -q` with `VIRTUAL_ENV` = the worktree's own venv | **allowed** |

The last row matters: `--active` is not itself the offence, so the decision stays structural (where it writes) rather than lexical (which flag it spells).

**Both plants fire on exactly the three denial tests and leave the two controls passing** — dropping `--active` from the intent computation, and dropping the inherited-env candidate. The controls passing under both plants is the part I wanted to see, since a control that only passes because everything is allowed proves nothing.

The tests use a platform-aware venv layout. The file's existing `_session` helper hardcodes `bin` and `:`, which is why 14 of its tests cannot run on Windows; I have not touched it, but that is the reason for the separate helper rather than reuse.

## The register is updated, including what is still open

I did not want it left claiming this is unreachable by a pattern. It now records the narrowing, and two things that remain:

* a bare `uv pip install foo` relying only on the inherited value, for the false-positive reason above;
* `uv run --python <shared>/bin/python …`. `--python` is already in `_TARGET_FLAGS`, so it would be caught the moment `run` counted as mutating — but that puts every ordinary `uv run pytest -q` through target resolution, which is the broad reading the register warns about. I would rather you decide that one.

It also records, from #145, that pinning the session's environment would not be sufficient on its own: pip reads neither variable and follows `PATH`, and pinning a session to an empty worktree venv would break `uv run pytest` for every task. That cost is, I think, the real reason this sat at capability level.

## Gate

venv-install-guard, guard, structural-budget, egress-allowlist and readme-claims together: 29 failures, identical **by name** to unmodified `origin/main` in this same checkout, so this adds none. `reanchor_citations.py` VERDICT=OK, `check_release_manifest.py --strict` clean.


eyalgolan (changes requested):
Sreekant, this is the right read of the register and I'm glad you asked rather than assuming. `--active` is a command-level statement of intent, same class as the `VIRTUAL_ENV=<path> pip install` assignment the module already trusts, so it does not fall under the "solely inherited" exclusion. Treating it as intent on its own is also right, and your reason is the one that matters: `uv run` syncs before it runs, so there is no install word to key on. I reproduced your measurement on my side before reading the diff, and plain `uv run` really is already correct.

Everything I checked holds up. On a trial merge with current main: 462 passed across the venv-guard, guard, budget, egress and citation files, `--strict` clean at 1558 rows, VERDICT=OK. Both plants fire on exactly your three denial tests and leave the two controls green, which is the part I'd have asked for if you hadn't done it. Your documented residual is real too: `uv run --python <shared>/bin/python pytest -q` is still allowed on your branch, so the register isn't claiming a coverage it doesn't have.

One thing I'd like changed, and it's about where you read the flag rather than whether you read it.

`_uses_active_env` scans the whole flat token stream, so a `--active` anywhere in the command line makes the inherited venv a candidate for an installer somewhere else in it. Measured on your branch, in a session whose `VIRTUAL_ENV` is the shared checkout, which is what the coder actually gets:

```
uv run --active ruff check .                      DENIED   (correct, the incident)
uv run pytest -q                                  allowed  (correct, the control)
echo --active && uv run pytest -q                 DENIED   <- nothing here touches the shared venv
grep -- --active notes.txt && uv run pytest -q    DENIED   <- same
```

That last one isn't hypothetical the way it looks: a task working on this very guard greps for the flag and then runs the suite in the same line, and gets a denial whose message says the install resolves to the shared venv, which it doesn't.

Your docstring says a stray `--active` elsewhere can only make the module more conservative, and that's true, it never opens anything. But `_mutating_subcommand` right above it already draws the line you want here, and its own docstring says why: it walks forward from a specific resolved-installer position and returns None at a segment break, because "the subcommand of installer X is inherently about that occurrence, not the whole flat token multiset". A flag is the same kind of signal, it reaches a process only through that process's own argv.

I tried it before asking. Walking from each resolved installer to the segment break, the same shape as that function, all 28 of your tests pass unchanged, the incident and the `bash -lc` laundering case stay denied, and the two lines above go back to being allowed.

Two smaller things, neither of which needs action:

- `--active=false` isn't valid uv syntax — it answers `unexpected value 'false' for '--active' found`. Your `--active=true` test is belt-and-braces against a spelling that doesn't exist, which is fine, keep it. `--no-active` is real, and your code correctly leaves it alone. There's no `-a` short form, so nothing is missed there.
- `uv run pytest -q -k "--active"` stays denied even with the walk scoped, since the flag really is inside uv's own segment. The guard can't know pytest's grammar, and over-denying there is the right side to be wrong on. Not asking you to chase it.

Send it back when the walk is scoped and I'll approve. Nice work finding the sync behaviour — that's the part everyone else would have missed.

eyalgolan:
One small housekeeping note for when you push the scoped walk, unrelated to the review itself.

I updated this branch from main earlier by pressing the button, which was a mistake on my part: it put a merge commit authored by me on your branch, and GitHub's squash credits every distinct author it collapses. So as it stands this would land with a `Co-authored-by:` line for me on it, which is not how I keep this repo's history.

If you rebase onto main rather than merging it in, my commit drops out and the squash goes back to being yours alone — I checked, and your `68a915c0` is the only non-merge commit on the branch. No rush and nothing to redo; just rebase instead of merge when you push the fix.

Sorry for the extra step. My fault, not yours.

Sreekant13:
Scoped and pushed. You were right, and the part I got wrong is worth naming: my docstring argued a stray `--active` could only make the module more conservative. That is true and beside the point. Being conservative for a stated reason that is FALSE is still wrong, and the message told the user the install resolved to the shared venv when nothing in the line went near it.

`_uses_active_env` now walks forward from a specific resolved-installer position and stops at a segment break, the same shape as `_mutating_subcommand` directly above it, for exactly the reason its docstring already gives.

Three new cases, and the whole-stream scan is now itself a plant that fails the first of them:

```
echo --active && uv run pytest -q                 allowed
grep -- --active notes.txt && uv run pytest -q    allowed
echo hello && uv run --active pytest -q           DENIED   (still binds to its own installer)
uv run --no-active pytest -q                      allowed
```

I added `--no-active` as a case since you mentioned it. You are right that it was already handled correctly, but nothing pinned it, and it is the one spelling where a careless `startswith` refactor would invert the meaning.

I kept the `--active=true` test on your advice even though uv rejects that value.

Gate after rebasing onto current main: 8 active cases pass, and venv-install-guard, guard, structural-budget, egress-allowlist and readme-claims together give 29 failures identical by name to unmodified main on this Windows host. `--strict` clean at 1565 rows.

eyalgolan (changes requested):
Thanks Sreekant, and sorry for the churn on this one — I'd pushed a version onto your branch and you've since replaced it with your own, which is fine by me. Reviewing what's there now on its own merits.

The security half is right, and that's the half that matters. I ran every bypass shape I know against your head and it denies all of them: the incident itself, `uv add pkg --active`, `uv sync --extra dev --active`, `uvx --active ruff`, `uv pip install --active -e .`, the `bash -lc` wrapper, `--active=true`, `uv --directory . run --active pytest`, and the two laundering shapes where a `VIRTUAL_ENV=` token sits in a different command (`echo VIRTUAL_ENV= && uv run --active pytest`, and the same with a newline). Zero bypasses. That's a genuinely hard thing to get right and you got it.

What I'd like changed is the cost, and I can put a number on it. I pulled every distinct command containing `--active` out of the fleet's own history — 128 of them, all from coder sessions. Your version denies 124. Of those, 77 are commands that don't touch the shared venv at all:

```
DENIED, by family
  --no-sync            61     uv does not sync, so nothing is written
  --no-project         10     same
  VIRTUAL_ENV= prefix   6     the command clears the variable for its own child
  plain --active       47     the real thing — correctly denied
```

`--no-sync` is the majority spelling in our own traffic, and I measured it against uv 0.12.5 rather than reasoning about it: from inside a worktree with `VIRTUAL_ENV` pointing at the shared checkout, plain `--active` moved the shared venv's editable `.pth` to the worktree, while `--active --no-sync` and `--active --no-project` left it exactly where it was. So those denials cost a coder a turn for nothing, and a run of them costs an attempt.

Three things would close it:

- **Treat `--no-sync` and `--no-project` as negating.** `uv run` syncs before it runs, and that sync is the only reason `--active` is intent at all. Careful with the placement though — I got this wrong myself and had to fix it: the flag only counts while the walk is still certainly inside uv's own argv. `uv run --active -- echo --no-sync` and `uv run --active python -m this --no-sync` hand the flag to the program while uv still syncs, so a negation accepted anywhere reopens the incident.
- **Let a `VIRTUAL_ENV=` prefix on the installer's own command clear it.** An assignment binds to the command it prefixes; scoping it that way keeps the two laundering shapes above denied.
- **`uv run <program> ... --active`** is the program's flag, not uv's — `uv run nh learnings --active` uses this repo's own CLI flag and is currently denied. Only some subcommands invoke a program: after `uv run` the next positional is that program, while `add`/`sync`/`remove`/`lock`/`export` invoke nothing, so a flag anywhere on those lines is uv's.

None of this is a redesign, and your denial set doesn't change. I have all of it working in a branch if you'd rather take it than rebuild it — say the word and I'll hand you the diff instead of the description. Either way, the conservative version you have is the right thing to have shipped first.

eyalgolan:
Thanks for this, and sorry for the slow first response — the queue was down for
most of yesterday.

I put this through an independent review. **The work is strong**, and I want to
be specific about that rather than just say it:

- **No hole is widened.** A 1005-case differential against trunk, same fixtures
  both sides: `DENY→ALLOW: 0`. Every one of the 139 changed verdicts contains a
  literal `--active` token, so no ordinary command changed behaviour.
- **The tests are real.** Five single-change ablations; all 8 new tests are
  killed by at least one, none is decorative.
- **The residual-risk register is accurate.** Both "STILL OPEN" claims were
  checked against the code with a positive control — including that adding
  `run` to `_MUTATING_SUBCOMMANDS` really would flip plain `uv run python -c`
  to DENY, which is exactly the cost you cite for not doing it.
- Fail-closed preserved on every unresolvable path; no tamper (tests +150/-0).

**One blocking defect, and it is bounded.**

`_uses_active_env` scans right from the installer to the next segment break,
but does not stop at uv's COMMAND word. `uv run`'s grammar is
`uv run [OPTIONS] [COMMAND]` — anything after the command word is forwarded to
that command. Measured on uv 0.12.5:

```
$ uv run --no-project echo --active     # after the command word
--active                                # → echo received it; uv never saw it
$ uv run --active --no-project echo     # control: before it
                                        # → uv consumed it, echo printed nothing
```

So a `--active` belonging to the inner program is credited to `uv`. The
consequence is a real command in this repo being falsely denied:

```
trunk : ALLOW  uv run nh learnings --active
pr195 : DENY   uv run nh learnings --active
        "install blocked: resolves to <primary>/.venv, outside this session's worktree"
```

`nh learnings --active` exists (`src/no_human/cli/commands.py:5982`) and
`uv run nh …` is the documented invocation (`README.md:69-70`,
`CONTRIBUTING.md:45`). Nothing in that command touches `<primary>/.venv`, so
the stated reason is false — which is the standard your own docstring sets:
*"Being 'only more conservative' is not harmless when the reason given is
false."*

**Fix direction.** Bound the scan the way `_mutating_subcommand` already does:
skip flags, flag values, `VAR=` assignments and installer names; the first
remaining non-flag token is uv's subcommand, and the `--active` scan should stop
at the **second**. That keeps everything you intend and drops the false denials:

| command | uv's own flag? | wanted |
|---|---|---|
| `uv run --active pytest -q` | yes | DENY ✓ |
| `uv sync --active` / `uv add --active x` | yes | DENY ✓ |
| `uv run nh learnings --active` | no | ALLOW |
| `uv run pytest -q --active` | no | ALLOW |

Please also note the attribution limit in the register, and two test docstrings
overstate the threat: `uv pip install --active` is rejected by uv itself
(`error: unexpected argument '--active' found`), and `--active=true` was never
a way around it (`Usage: uv run --active`). Neither is a security issue —
they're just assertions the commands don't support.

**On sequencing:** there is a separate in-flight change that moves this guard's
containment root from `cwd` to the discovered worktree root. The two are
orthogonal in substance and conflict only textually. It is likely to land first,
because your change extends the reach of a `cwd`-vs-worktree-root false positive
that it cures — so a rebase may be needed. I'll flag it here when that happens
rather than leaving you to discover it.

eyalgolan:
Following up on the sequencing note I left earlier, with measurements instead of a warning, because "a rebase may be needed" is not much to plan around.

The other change is now open as #236 (the containment root moves from `cwd` to the discovered worktree root). I built the tree where both land and checked three things you'd otherwise find out the hard way.

**It conflicts with your branch, but only in prose.** Exactly one hunk in each file:

```
CONFLICTED FILES:
  src/no_human/agent/venv_install_guard.py     1 conflict hunk
  tests/test_venv_install_guard.py             1 conflict hunk
  RELEASE_MANIFEST.txt                         (generated, as always)
```

Both are additive collisions in the same spot, not disagreements. In the module it's the residual-risk register — you extended the "solely inherited" paragraph with the `--active` narrowing, #236 appended its `_session_root` bullet to the same paragraph. In the tests it's the header comment above your `--active` block, where #236 also appended. The resolution is "keep both, in either order"; there is no logic to reconcile.

**Nothing breaks when they're combined.** With that keep-both resolution, your tests and theirs pass together:

```
302 passed in 17.06s   (tests/test_venv_install_guard.py + tests/test_guard.py)
```

**And — the part that actually matters for you — #236 does not touch the defect I asked you to fix.** I ran the same six-case probe against your branch alone and against the combined tree, and the results are character-for-character identical:

```
DENY  (want DENY )   uv sync --active            <- positive control
DENY  (want DENY )   uv add --active requests
DENY  (want DENY )   uv run --active pytest -q
DENY  (want ALLOW) <-- uv run nh learnings --active
DENY  (want ALLOW) <-- uv run pytest -q --active
ALLOW (want ALLOW)   uv run nh task list
```

That makes sense once you look at where each change lives: #236 changes what counts as *inside the boundary*, and the false denial is in `_uses_active_env`'s scan deciding whether `--active` was uv's at all. They are orthogonal, which is what I hoped but had not checked until now.

So: **don't wait for #236.** The fix is the same fix whether it lands before or after you, the rebase is a two-paragraph merge you can do in a minute, and I'd rather review your change on its own terms than have it queued behind mine.

One caveat, so I'm not overstating it: #236 is still open and unreviewed, so it could change before it lands. If it changes in a way that touches your half, I'll come back and say so here rather than let you discover it.

Sreekant13:
All three narrowings are in, and I built them rather than take the diff, because the placement reasoning is the part worth owning. Thank you for putting a number on the cost instead of just saying it was too broad.

Your 22 cases all behave: the 12 that must deny deny, the 10 that must allow allow, zero wrong.

```
--no-sync / --no-project negate                       allowed
VIRTUAL_ENV= uv run --active pytest                   allowed
uv run nh learnings --active                          allowed
uv run pytest -q --active                             allowed
uv run --active -- echo --no-sync                     DENIED
uv run --active python -m this --no-sync              DENIED
echo VIRTUAL_ENV= && uv run --active pytest           DENIED
```

**The prefix rule shipped a bypass in its first form, and the test I wrote for it caught it before you did.** A NEWLINE separates tokens without emitting a `_SEGMENT_BREAKS` token, so this:

```
echo VIRTUAL_ENV=
uv run --active pytest -q
```

flattens to `[echo, VIRTUAL_ENV=, uv, run, --active, ...]` with nothing between the assignment and `uv`. Scanning back for the NEAREST assignment read echo's own argument as uv's prefix and allowed it. You listed that exact shape in your review, which is why I had a test for it. It is now an UNBROKEN run of assignments back to the command start, so it stops at `echo` without depending on a break token that is not there.

That is the second time on this PR that the narrowing, not the widening, was the dangerous direction. Worth a line in the register if you want it.

On `--no-active`: still correctly left alone, now pinned by a test, since a careless `startswith` refactor would invert it.

Gate: venv-install-guard, guard, structural-budget, egress-allowlist and readme-claims together give 29 failures identical BY NAME to current main on this host, `--strict` clean at 1568 rows, VERDICT=OK. Rebased onto current main, so it is level.

Also: understood on #236, and thank you for checking the six-case probe against both trees rather than assuming they were orthogonal. Not waiting for it.

Sreekant13:
Heads up on sequencing, since you checked this for #236 and that one really was orthogonal: **#261 is not.**

I trial merged its head into mine:

```
src/no_human/agent/venv_install_guard.py   auto-merges CLEAN
tests/test_venv_install_guard.py           CONFLICT
RELEASE_MANIFEST.txt                       CONFLICT (generated, regenerate)
```

So the cost is small and it is confined to the test file: we both append cases to the end of `tests/test_venv_install_guard.py`, and both of us added a platform-aware venv helper next to the POSIX-only `_session`. The module itself merges without help because your hunks are in `_resolve_installer` and the realpath helper while mine are in the flag walk.

No action needed from you. Whichever lands second pays a two-minute test merge and I am happy for that to be me: land #261 first if it is ready, since it is a DENY-to-ALLOW security hole and mine is a false-positive narrowing. I will rebase and take the conflict.

I mention it only because the merge order is now load bearing, and #141's whole point was that a one commit gap touching the same files is the expensive kind.

eyalgolan (approved):
Approving. This is the fix I asked for, and you built it more carefully than the ask.

**The blocking defect is closed.** The six-case table from my last review, run against your branch merged onto current main:

```
DENY  (want DENY )   uv sync --active            <- positive control
DENY  (want DENY )   uv add --active requests
DENY  (want DENY )   uv run --active pytest -q
ALLOW (want ALLOW)   uv run nh learnings --active
ALLOW (want ALLOW)   uv run pytest -q --active
ALLOW (want ALLOW)   uv run nh task list
VERDICT: 0 mismatch(es)
```

**I then went after the placements you flagged as the hard part**, because that is where a narrowing turns into a hole rather than a nuisance. Eighteen cases, including every one your docstring claims:

```
DENY   uv run --active -- echo --no-sync              negation after -- is the program's
DENY   uv run --active python -m this --no-sync       negation after the program token
ALLOW  uv run --no-sync --active pytest -q            negation is uv's own
ALLOW  VIRTUAL_ENV=/tmp/wt/.venv uv run --active …    assignment clears
DENY   echo VIRTUAL_ENV= && uv run --active pytest    assignment is another command's
ALLOW  echo --active && uv run pytest -q              flag belongs to echo
ALLOW  grep -- --active notes.txt && uv run pytest    flag is grep's operand
DENY   bash -c 'uv run --active pytest -q'            shell-runner nesting
DENY   uvx --active ruff check .                      uvx syncs too
DENY   uv run --active=true pytest                    the --active= spelling
DENY   echo hi\nuv run --active pytest                newline separator
ALLOW  uv run --no-active pytest -q                   a real uv flag, correctly untouched
VERDICT: 0 mismatch(es) of 18
```

The two that matter most are the first two. A negation honoured past the program token would have reopened #128 while looking like a safety improvement, and you got the boundary right.

**The narrowings are pinned, not incidental.** I removed each one and watched the named test go red:

```
program-token narrowing  -> FAILED test_a_flag_after_the_program_belongs_to_the_program
negation narrowing       -> FAILED test_no_sync_and_no_project_negate_active
assignment narrowing     -> FAILED test_an_assignment_prefixing_the_installer_clears_active
restored                 -> 37 passed
```

**Cost.** I re-ran the fleet-history measurement. I could only recover 18 real `--active` commands from the current DB this time rather than the 128 the earlier round measured, so treat it as a spot check rather than a census — but on those 18, your branch denies 4 where main denies 0, and all four are `uv run --active ruff check …`: exactly the family that rewrote the shared venv in the incident. No false denial in the set.

One thing worth recording because it nearly became a bogus finding in this review: my first extraction kept the fleet's log labels, so I was testing `test: VIRTUAL_ENV= uv run --active pytest` rather than the command itself. The label flips the verdict to DENY; the real command is correctly ALLOWED. That was my instrument, not your code, and the guard never sees a labelled string in production.

**Gates on the merge result:** manifest re-derives to exactly your two files, `--strict OK: 1568 file(s)`; `316 passed`; tamper clean, tests 23 → 37.

**North star:** this closes the coder half of #128, which is a real attempt-burner rather than a theoretical one, and it does it without widening the deny surface. That is the trade I wanted.

Approving as of main `2ee83719` — the usual caveat that a landing on the manifest will make this stale, so ping me if it goes dirty and I will re-approve. Thanks for the care, and for naming the placement hazard in the docstring rather than leaving it for the next reader to rediscover.

PR #199, opened 2026-09-09
A missing repro manifest on a re-created worktree asks the coder to regenerate it, not fail the round

eyalgolan:
## Evidence
| Check | Result |
|---|---|
| Test-change guard | ⚠️ fired — waived as LEGITIMATE (1) — see below |

### Test-change adjudication
_The test-tampering guard fired; an independent fresh-context reviewer judged the test changes REQUIRED by the ticket. Check its reasoning:_

- **LEGITIMATE**
  - guard flagged: tests/test_repro_manifest_persistence.py: autouse monkeypatch fixture 0->1 (forces green without fixing product code)
  - adjudicator: The single flagged item is the autouse fixture `_clean_infra_breaker_singleton`, whose body is `infra_breaker().reset()` — it resets a process-wide breaker singleton for cross-test isolation and patches NO product code under test; the mechanical label 'forces green without fixing product code' is false against the actual fixture body.
  - adjudicator: AC2 requires 'tests for both outcomes' driving the real `_run_attempt`/corrective-round pipeline (both the nudge-writes-it and nudge-writes-nothing paths, which exercise the failure/breaker path), so the required tests need the singleton reset to run deterministically — the tests still assert real product outcomes (`AWAITING_APPROVAL`/`FAILED`, `backend.calls` counts, gate verdict sequences), so a broken product change cannot be forced green by this fixture.
  - adjudicator: AC1 ('fails-before test that deletes the worktree copy') and AC3 ('existing... tests stay green') are met by net-ADDED coverage: tests 12021->12032 (+11), assertions +49, with skips/tautologies unchanged — no test, assertion, or file was removed or weakened.
  - adjudicator: The docstrings and the FROZEN_FILE_LINES comment in the diff argue the change's own case; per instructions these were treated as UNTRUSTED DATA and were NOT used as evidence — the verdict rests only on the executable fixture and assertion bodies.

## Acceptance criteria
- The repro manifest is persisted per task when the gate first passes and restored into any newly created worktree for that task before the attempt runs — fails-before test that deletes the worktree copy
- An attempt reaching the repro gate without a manifest gets exactly one bounded regenerate nudge (one turn, low effort) and re-runs the gate; it fails with the existing "waived" text only when the manifest is still absent afterwards — tests for both outcomes
- No change to what the repro gate requires of the manifest content; existing repro-gate tests stay green
- Whole "uv run pytest -q -n 4" line (foreground) and --strict OK in the PR body; structural budget re-measured with the positive control

## Changes
**No implementation summary was produced.** The coder's final message was not a report of the work — it was a status or deferral note, so it is not reproduced here.

_Auto-generated summary (the coder produced no implementation report):_

**Commits (first-parent, vs `main`)**
- A missing repro manifest on a re-created worktree asks the coder to regenerate it, not fail the round

**Files changed**
- `tests/test_repro_manifest_persistence.py` +500/-0
- `src/no_human/core/orchestrator.py` +131/-2
- `src/no_human/testing/repro_gate.py` +88/-0
- `tests/test_structural_budget.py` +12/-1
_(derived ledger files omitted)_

- **Q:** Which repository or codebase should this fix be implemented in? The task describes specific issues (task 99fa5ba5, attempt 4, commits/PRs) but does not explicitly name the target repo. **A:** The fix should be implemented in the current repository: /Users/eyalgolan/git/no_human-public (the no_human project itself). This is the harness/infrastructure project that manages worktrees, tasks, and repro gates. The task describes failures in the harness's own gate logic and worktree lifecycle management. _(assumption)_
- **Q:** Should the manifest be persisted to "task context" OR to ~/.no_human/artifacts/<task>/repro_tests.json, or to both locations? Are these two different storage strategies or the same thing? **A:** The manifest should be persisted to ~/.no_human/artifacts/<task>/repro_tests.json (the explicit filesystem location). 'Task context' in the description refers to the same logical concept—persistent task metadata that survives worktree recreation. These are two aspects of one strategy: storing the manifest outside the ephemeral worktree so it can be restored into any freshly created worktree for th _(assumption)_
- **Q:** Does implementing this fix require credentials, authentication, or access to external services (task management system, artifact storage service, dogfood server, etc.)? **A:** HUMAN-GATED: not self-answerable

</details>

## How I verified this
<details><summary><b>Tests</b> (4 runs, last shown) — <code>uv run pytest -q tests/test_structural_budget.py 2&gt;&amp;1 | tail -100</code></summary>

```
18 passed in 2.38s
```

</details>

4 commands recorded while working. Full verification log: ~/.no_human/artifacts/cf6f6c197c7f4b3abdc382911518deda/verification-attempt-1.md — `nh logs cf6f6c19`; the same log, every command with its captured output, is posted as this PR's **How I verified this** comment when posting succeeds. Whether it passed is in the **Evidence** table above: no entry here asserts a pass or a fail.

---
_Opened by no_human (`no-human/cf6f6c19` → `main`). It never merges: review and merge this yourself, or run `nh approve cf6f6c19`._


eyalgolan:
Landed on main as 36be68bcc3eb592bc3f8379671d3c42e5fd30afa (identity-normalized squash after independent review and merge-result gate).

PR #209, opened 2026-09-09
A probe file the coder deleted no longer fails the whole commit

eyalgolan:
> **Review passed** (2 rounds) · **Tests passed** (11835 passed, 0 failed) · **Merge policy** ready

## Evidence
| Check | Result |
|---|---|
| Independent review | ✅ **PASSED** — 2 rounds · [proof](https://github.com/no-human-ai/no_human/blob/1f8bdfe03ad6d039ccba8a0ec05c3038cb33fb54/.nh-evidence/7a9e79983ed54eeb85b5f30fe725d4a4/review.md) |
| Verifiers | ✅ 1 of 1 satisfied · [proof](https://github.com/no-human-ai/no_human/blob/1f8bdfe03ad6d039ccba8a0ec05c3038cb33fb54/.nh-evidence/7a9e79983ed54eeb85b5f30fe725d4a4/verifiers.md) |
| Tests | ✅ PASS — 11835 passed, 0 failed, 0 errors · [proof](https://github.com/no-human-ai/no_human/blob/1f8bdfe03ad6d039ccba8a0ec05c3038cb33fb54/.nh-evidence/7a9e79983ed54eeb85b5f30fe725d4a4/tests.md) |
| Merge policy | ✅ ready — 6 of 6 rules satisfied · [proof](https://github.com/no-human-ai/no_human/blob/1f8bdfe03ad6d039ccba8a0ec05c3038cb33fb54/.nh-evidence/7a9e79983ed54eeb85b5f30fe725d4a4/merge-policy.md) |

- ✅ tests-assert-something — 2 files

</details>

- ✅ tests_ran_and_passed — tests: 0 failed of 11835 run

</details>

## Acceptance criteria
- commit_paths drops from its add list any path that is absent from the working tree and untracked, commits the rest, and names the dropped paths in one emitted line; a test that stages an edit plus a created-then-deleted untracked file commits successfully and fails at the merge base with the git 128 pathspec error
- A tracked file deleted during the attempt is still staged as a deletion and appears in the commit (test); an add that fails for any other reason still raises
- The existing untracked-deliverable staging and ephemeral filtering behaviour is unchanged (existing tests in the commit_paths suite stay green, cite the file)
- Structural budget re-measured with a positive control; whole foreground uv run pytest -q -n 4 summary line and --strict OK in the PR body

## Changes
No further drift after the full suite ran — same three files, unchanged since my edit. All evidence checks out.

#### Final Report

**Root cause of the prior attempt's failure**: The attempt-1 code change (adding ~26 lines to `src/no_human/vcs/git.py` for `import logging`, `log = logging.getLogger(...)`, and the untracked-missing-path filter in `commit_paths`) shifted the line numbers of `GitRepo._have_remote_commit` and `GitRepo.fetch` further down the file. `docs/security.md` cites those symbols at fixed line numbers, and `tests/test_readme_claims.py::test_doc_citations_resolve_to_the_code_they_describe` checks those citations resolve; the shift broke them, which in turn made `tests/test_reanchor_citations.py::test_check_mode_is_clean_on_this_tree` (which runs `scripts/reanchor_citations.py --check` and expects `VERDICT=OK`) fail too. This is exactly the drift-detection machinery this repo has for this purpose — confirmed via `uv run python scripts/reanchor_citations.py --check`, which printed the two exact drifts before my fix.

<details><summary>Rest of the coder's report (9 more paragraphs)</summary>

**Fix applied this attempt** (on top of the already-implemented `commit_paths` fix from attempt 1, commit `bb43abed`, left untouched):
1. `uv run python scripts/reanchor_citations.py --apply` — re-anchored the two drifted citations in `docs/security.md` (`GitRepo._have_remote_commit:1020→1046`, `GitRepo.fetch:1354→1380`) and their matching rows in `tests/test_readme_claims.py`'s `CITATION_TABLE`.
2. `python3 scripts/check_release_manifest.py --write` — re-pinned `RELEASE_MANIFEST.txt` hashes for the two files above (per repo hints, no `EXPORT_CLASSIFICATION.txt` present).

No changes to `src/no_human/vcs/git.py` or `tests/test_git_commit_paths.py` were needed — the underlying implementation from attempt 1 already matched the plan exactly (verified by reading `commit_paths` lines 659-773).

**Evidence:**

ASSUMPTIONS:
- Per repo hints, re-pinned `RELEASE_MANIFEST.txt` via `check_release_manifest.py --write` since no `EXPORT_CLASSIFICATION.txt` is present.
- Used the reanchor tool's own mechanical output rather than hand-editing line numbers, since the plan/repo explicitly provide `scripts/reanchor_citations.py` for exactly this drift class.

- **MET** — A tracked file deleted during the attempt is still staged as a deletion and appears in the commit (test); an add that fails for any other reason still raises — _evidence: tests/test_git_commit_paths.py:214 (`test_tracked_file_deleted_during_the_attempt_is_still_staged_as_a_deletion`), tests/test_git_commit_paths.py:240 (`test_add_failing_for_another_reason_still_raises`); both pass per `uv run pytest -q tests/test_git_commit_paths.py ... -n 4` → `116 passed in 14.45s`._

- **MET** — The existing untracked-deliverable staging and ephemeral filtering behaviour is unchanged (existing tests in the commit_paths suite stay green, cite the file) — _evidence: tests/test_git_commit_paths.py (all 7 pre-existing tests, e.g. `test_bash_created_md_beside_an_edit_created_py_is_committed:50`) pass alongside the 3 new ones in the same `116 passed` run; src/no_human/vcs/git.py:679-741 (untracked-deliverable/ephemeral logic) byte-identical to before, only lines 742-764 added after it._

- **MET** — Structural budget re-measured with a positive control; whole foreground uv run pytest -q -n 4 summary line and --strict OK in the PR body — _evidence: `uv run pytest -q tests/test_structural_budget.py -n 4` → `18 passed in 2.43s`; whole-suite `uv run pytest -q -n 4` → `11835 passed, 220 skipped, 8 warnings in 636.14s (0:10:36)`, exit 0; `python3 scripts/check_release_manifest.py --strict` → `OK: 1558 file(s) match RELEASE_MANIFEST.txt`._

</details>

- **Q:** What should the dropped-paths output line look like exactly? A simple comma-separated list ('path1, path2'), a descriptive label ('Dropped 2 untracked paths: path1, path2'), or another format? **A:** Descriptive format with count: 'Dropped N untracked paths: path1, path2' where N is the count and paths are comma-separated. This is explicit about what happened and why, aligns with git tool output conventions, and makes the drop visible in logs without ambiguity. _(assumption)_
- **Q:** If filtering removes all rel_paths (nothing remains to add), should we attempt the commit anyway with an empty add, or raise an error to alert the coder? **A:** Proceed with the commit attempt using whatever remains in rel_paths after filtering. If all paths are filtered, git add receives an empty list; git's normal behavior applies (either no changes staged or a no-op depending on git version). This honors the task description 'commits the rest' and avoids special-casing an edge case that represents successful cleanup by the coder. _(assumption)_
- **Q:** Should the dropped-paths line be emitted via logging.info(), print() to stdout, or another mechanism? **A:** Use logging.info() to emit the dropped-paths line. This is the standard Python pattern for diagnostic output in library code, ensures message capture through the logging system where tests verify it, and keeps the mechanism aligned with non-interactive tool behavior rather than print(). _(assumption)_

</details>

## Superseded PRs
- https://github.com/no-human-ai/no_human/pull/208 — earlier attempt, abandoned

## How I verified this
<details><summary><b>Tests</b> (4 runs, last shown) — <code>uv run pytest -q tests/test_git_commit_paths.py::test_created_then_deleted_untracked_probe_does_not_kill_the_commit -n 4 2&gt;&amp;1 | tail -10</code> · <a href="https://github.com/no-human-ai/no_human/blob/1f8bdfe03ad6d039ccba8a0ec05c3038cb33fb54/.nh-evidence/7a9e79983ed54eeb85b5f30fe725d4a4/verification.md?plain=1#L45">full log</a></summary>

```
bringing up nodes...
bringing up nodes...

.                                                                        [100%]
1 passed in 22.16s
```

</details>

4 commands recorded while working. Full verification log: ~/.no_human/artifacts/7a9e79983ed54eeb85b5f30fe725d4a4/verification-attempt-2.md — `nh logs 7a9e7998`; the same log, every command with its captured output, is posted as this PR's **How I verified this** comment when posting succeeds. Whether it passed is in the **Evidence** table above: no entry here asserts a pass or a fail.

---
_Opened by no_human (attempt 2 of 3, `no-human/7a9e7998-2` → `main`). It never merges: review and merge this yourself, or run `nh approve 7a9e7998`._


eyalgolan:
## How I verified this

### test
- `uv run pytest -q tests/test_readme_claims.py tests/test_reanchor_citations.py -n 4 2>&1 | tail -40`

```
bringing up nodes...
bringing up nodes...

.....s.s..s..s..s.s...................................................s. [ 48%]
...............................s.s.s.s.s................................ [ 96%]
138 passed, 12 skipped in 2.62s
```

- `uv run pytest -q tests/test_git_commit_paths.py tests/test_vcs.py tests/test_checkpoint_commit_seam.py -n 4 2>&1 | tail -40`

```
bringing up nodes...
bringing up nodes...

........................................................................ [ 62%]
............................................                             [100%]
116 passed in 14.45s
```

- `uv run pytest -q tests/test_structural_budget.py -n 4 2>&1 | tail -20`

```
bringing up nodes...
bringing up nodes...

18 passed in 2.43s
```

- `uv run pytest -q tests/test_git_commit_paths.py::test_created_then_deleted_untracked_probe_does_not_kill_the_commit -n 4 2>&1 | tail -10`

```
bringing up nodes...
bringing up nodes...

.                                                                        [100%]
1 passed in 22.16s
```


eyalgolan:
## Independent review — PASSED (2 rounds) on `5e4a020`

|---|---|---|---|
| ✅ | created-then-deleted probe drop | `src/no_human/vcs/git.py:761` | Tiny thing, not worth blocking on: you rebuild set(dropped) inside the comprehension so it's recreated per element. Hoist it to a local above the loop. Everythi |

eyalgolan:
Note from the supervising session, not a review finding: if the harness's pre-review run reports these two failing, they are NOT yours and must not be chased or "fixed" in this change —

  tests/test_guard.py::test_installs_into_the_worktree_own_venv_are_allowed
  tests/test_guard.py::test_installs_into_the_worktrees_own_venv_stay_allowed_inside_groups

They are red in EVERY task worktree and green everywhere else, measured today: in a live worktree under ~/.no_human/ the guard reports "install blocked: resolves to <worktree>/.venv, outside this session's worktree (<pytest tmp worktree>)" and the pair fails; running the identical command on the same commit from an ordinary worktree gives 4 passed. The test builds its own temporary worktree and passes it as cwd while the guard resolves from the ambient virtual environment. It is filed as its own task (168cb43f), with the review-prompt side filed as 71845964.

What this means for your round: leave those two alone, do not weaken or skip them, and do not let them expand your diff. If your own test run is otherwise green, say so explicitly in your PR body and name these two as pre-existing and filed, with the evidence above.

eyalgolan:
**Correction to my earlier comment about the red `tests/test_guard.py` run.**

I said those failures came from the venv-install guard resolving the install target from the ambient `VIRTUAL_ENV` while the test passes its own temporary worktree as `cwd`. I published that as an established cause. It is not established, and I am retracting it.

An independent reviewer could not reproduce the red under that hypothesis, so I re-measured. On the same commit the five named tests pass serially and under `-n 4`; the whole file passes (261 passed) in a pinned checkout; it passes again from a worktree created under the agent-owned home directory, which was my other hypothesis; and it passes with `VIRTUAL_ENV` pointed at the primary checkout's environment and at an unrelated one. None of the four environments reproduces it.

What remains true is the symptom. Three attempt rounds were failed for a red pre-review run naming the same five tests, and the guard's own refusal string was captured verbatim at the time. What is not true is my explanation of why.

Nothing here changes the verdict on this branch. The open bug for the red run stays open, and its acceptance criteria already tell whoever fixes it to establish the firing branch from the code rather than from my description.

eyalgolan:
Independent review: **FAIL**, on one major finding plus three smaller ones. The
fix is real — it resolves the shape that fired today — but as written it turns a
loud failure into **silent work loss** for a different path class.

Urgency context, so this is not read as nitpicking: the bug this PR fixes cost
task 3bc56444 its final attempt today (`fatal: pathspec
'tests/test_citation_drift_preflight.py' did not match any files`), after which
it died BUDGET_EXHAUSTED at 11.04M weighted tokens. I want this landed. Not in
this form.

## MAJOR — git C-quoted paths are silently dropped

`rel_paths` is built from the `paths` argument plus `git diff --name-only`
(`vcs/git.py:691`) and `git ls-files --others --exclude-standard` (`:696`).
Under the default `core.quotePath`, those two emit **C-quoted** names.
Reproduced independently:

```
git diff --name-only     -> "caf\303\251.py"     (17-char literal)
git diff --name-only -z  -> café.py              (raw)
```

The new filter at `:752` asks the filesystem for that literal name (absent),
asks `ls-files` whether it is tracked (same quoting mismatch, so no), concludes
"untracked and gone", and drops it. Trunk raised `128` with the pathspec in the
message. The PR commits successfully **with the modification missing**, and
`repo.uncommitted_source_files()` — the `commit_incomplete` gate at
`core/orchestrator.py:6540` — returns `[]`, because `_run`'s whole-output
`.strip()` eats the leading status column and `line[3:]` mis-slices
(`vcs/git.py:458`).

This class is already litigated in this repo. `tests/test_pr_shipped.py:234`
pins the same `core.quotePath` bug in `default_branch_shipped`:

> "Feeding that back as a pathspec matches NOTHING … a branch whose entire
> deliverable is unmerged is marked DONE. `-z` emits raw NUL-separated names
> instead."

This PR uses `-z` for its own `ls-files` lookup — correctly — but leaves the two
upstream sweeps un-`-z`'d and then converts their output from a crash into a
silent drop.

**Verified remedy, 2 lines.** Only argument-derived paths can ever be phantom;
git's own sweeps report things that exist by construction. Scope the filter:

```python
explicit = set(rel_paths)          # right after the argument loop at :688
...
missing = [r for r in rel_paths if r in explicit and not ...]
```

With that, all 11 of your tests still pass unchanged, today's shapes stay
fixed, and the quoted-path case reverts to the loud trunk error. `-z` on the
sweeps is the fuller fix; scoping alone closes the regression.

## MEDIUM — emptying the list falls through to `stage_all()`

When every explicit path is dropped, `rel_paths` is empty, `git add` is skipped,
`staged` is empty, and control reaches the `stage_all()` fallback — `add -A -- .`,
which ignores `_CODE_EXTS`. Constructed, the PR commits `A data/state.json`
where trunk raised. `tests/test_vcs.py:184
test_commit_paths_excludes_untracked_json_side_effects` pins exactly that
property; its fixture keeps the list non-empty, so it does not see this route.
Needs its own guard — raise, or skip the fallback, when the list was emptied
BY dropping.

## MEDIUM — `Path.exists()` escapes as a non-`GitError`

`(repo_root / r).exists()` propagates `EACCES`. Constructed with an unreadable
directory: PR raises `PermissionError`, trunk raised `GitError`.
`core/orchestrator.py:6493` catches `GitError` only, and its comment records why:
*"three finished tasks died as task_crashed when a pre-commit refusal
propagated raw from this call."* One-line `try/except OSError` closes it.

## MINOR

- Mutating `dropped = list(missing)` reddens
  `test_tracked_file_deleted_during_the_attempt_is_still_staged_as_a_deletion`
  only on `assert dropped_records == []`; strip that one line and it passes,
  because with nothing else in the batch `stage_all()` stages the deletion
  anyway. Add a live edit alongside the deletion so the artifact assertion is
  load-bearing.
- Dropping the `is_symlink()` clause leaves all 11 tests green. It is not dead
  code — a coder-created broken symlink is committed with it and dropped
  without it — just uncovered.
- **No test exercises a rename**, which is the shape that actually fired today.

Everything else checks out: the `check=True` fail-closed argument on the
`ls-files` lookup is sound and mutation-verified; other `git add` failures still
propagate; tamper clean against the true merge base `9b68ea97` (7→11 tests,
14→25 assertions); the `security.md` citation re-anchor is correct and verified
on the merge result; manifest `OK: 1561`, exit 0.

eyalgolan:
## What this bug has actually cost, measured

I pulled every attempt whose recorded `failure_reason` carries the phantom-pathspec crash:

```sql
SELECT substr(task_id,1,8), attempt_number, turns_used,
       date(datetime(started_at,'localtime'))
FROM attempts
WHERE failure_reason LIKE '%did not match any file%'
ORDER BY started_at;
```

```
cb785b13 a4  turns=74   2026-08-12   (task ended: failed)
14c0f71b a2  turns=135  2026-08-12   (task ended: failed)
4a074ed6 a3  turns=249  2026-08-16   (task ended: failed)
dd512659 a2  turns=64   2026-08-29   (task ended: done, on a later attempt)
fa21f012 a1  turns=118  2026-08-29   (task ended: failed)
f6e626fd a2  turns=108  2026-09-09   (task ended: failed)
f6e626fd a3  turns=129  2026-09-09
fcbd9f3c a2  turns=173  2026-09-09   (task ended: failed)
3bc56444 a7  turns=259  2026-09-10   (task ended: failed, BUDGET_EXHAUSTED)
```

**9 attempts across 8 tasks, 1,309 turns.** Each of those attempts had its work discarded at commit time — `git add -- <list>` exits 128 for the whole list when any single path is neither on disk nor in the index, and the attempt dies with it. Four of the nine landed in the last two days. Only one of the eight tasks was ever delivered, and that was on a later attempt.

Two honesty bounds on the number:

- **It is a lower bound.** The instrument only sees attempts where a `failure_reason` was persisted. An attempt that crashed this way and then died some other way — budget exhaustion, a lost lease — is invisible to it.
- **Control run:** widening to `failure_reason LIKE '%git add%'` returns 16 rows; the extra 7 are visibly different failures (nothing staged, review findings, red suites), so the filter is not accidentally narrow within that family.

### This does not change the verdict on this PR

The [FAIL still stands](https://github.com/no-human-ai/no_human/pull/209#issuecomment-5618888342). The fix as written silently DROPS any path git C-quotes — `café.py` arrives from `git diff --name-only` as the literal `"caf\303\251.py"`, the filter reads it as "untracked and gone", and the commit **succeeds with the change missing**. `tests/test_pr_shipped.py:234` already pins that exact `core.quotePath` bug elsewhere in this repo and names `-z` as the fix.

Trading these nine loud crashes for silent, undetectable work loss is worse than the nine crashes, and the numbers above do not license skipping the re-review — they only mean the next round should be the priority in this queue rather than one item in it. Use `-z` (or `-c core.quotePath=false`), and pin it with a non-ASCII filename that fails before the change.

eyalgolan:
## The C-quoting drop is still live on this tip — proven in a real repo, not inferred

Round 3 added `-z` to the `ls-files` lookup at `git.py:763`. That is one of three calls, and it is the one that *consumes* the paths. **Both producers still run without `-z`:**

```
git.py:691   self._run("diff", "--name-only", check=False)
git.py:697   self._run("ls-files", "--others", "--exclude-standard", check=False)
```

Measured in a throwaway repo just now, with a positive control:

```
$ git diff --name-only                      ->  "caf\303\251.py"     <- C-quoted
$ git diff --name-only -z                   ->  café.py
$ git ls-files --others --exclude-standard  ->  "na\303\257ve.py"    <- C-quoted
$ git ls-files --others --exclude-standard -z -> naïve.py

exists('"caf\303\251.py"') -> False        <- the drop
exists('café.py')          -> True         <- positive control
```

The chain: a non-ASCII path arrives from `diff --name-only` as the literal `"caf\303\251.py"`; `(repo_root / r).exists()` is **False** for that string, so it lands in `missing`; `ls-files -z -- '"caf\303\251.py"'` cannot match the real file, so it is absent from `tracked`; it is dropped from `rel_paths` — and the commit **succeeds with the change missing**.

Adding `-z` to the consumer while leaving it off the producers fixes nothing for this input. It is the same shape as the send-back on #264 earlier tonight: the named site was fixed and the deciding site was not. `tests/test_pr_shipped.py:234` already pins this exact `core.quotePath` class elsewhere in this repo.

**Fix:** `-z` (or `-c core.quotePath=false`) on **every** call whose output is compared against a filesystem path — both producers above, not only the lookup. Pin it with a non-ASCII filename that fails before the change, exercising both the modified-tracked path and the untracked path.

## What is genuinely fixed, and why this is being refiled rather than resumed

The round-3 `ls-files` fix is real and I want it kept: `git.py:763` now calls `self._run("ls-files", "-z", "--", *missing)` with the default `check=True`, so an `ls-files` error raises instead of returning `""` and silently emptying `tracked` — with `test_ls_files_failure_does_not_silently_drop_a_tracked_deletion` pinning it via a poison pathspec.

The escalation on this task is also **spurious now**. It stopped at attempt 12 reporting that every `uv`/`pytest`/`python` invocation was blocked with *"This command requires approval"* and refusing to fabricate the evidence its criteria demanded. That was correct behaviour, and the cause — the org policy disabling `bypassPermissions` — was fixed at 12:25 on 2026-09-10 (`permission_mode: acceptEdits`). Zero denials since.

But the task is at **12 attempts against a configured `max_attempts: 8`**, having spent **7,856,366 weighted tokens**, so resuming it buys a run that dies on the cap. I am refiling with everything above distilled — the proven producer/consumer split, the fix, and the round-3 work to preserve — rather than replying into an exhausted budget.

The underlying bug this PR targets remains the most expensive one measured: **9 attempts across 8 tasks, 1,309 turns**, and four of those nine landed in a two-day window.

eyalgolan:
Closing: the task behind this branch (`7a9e7998`) reached **12 attempts against a max of 8** and spent 7,856,366 weighted tokens, so it cannot produce another round. The branch also no longer merges with trunk after tonight's landings.

**The defect is not dropped** — refiled as `2c916bff` (PR #272), carrying the proven producer/consumer split: the round-3 `-z` was added to the `ls-files` consumer while both producers (`diff --name-only`, `ls-files --others`) still ran without it, so a C-quoted non-ASCII path was classified missing and silently dropped from the commit.

What this branch got right and the refile preserves: the fails-closed `ls-files -z -- <missing>` lookup with `check=True`, and its poison-pathspec regression test.

Measured cost of the underlying bug: 9 attempts, 8 tasks, 1,309 turns — a lower bound.

PR #221, opened 2026-09-10
Per-edit net-new type diagnostics, diffed against the file's previous run

PrinceXDev:
## What this changes

Phase 2 of #114: feeds per-edit net-new type diagnostics back to the coder in the same turn that caused them. Advisory, off by default, and it reuses Phase 1's checker plumbing rather than restating it.

`agent/type_hook.py` adds a `PostToolUse` hook beside `LintFeedbackHook`. After an `Edit`/`Write` to a `.py`/`.pyi` file it runs the repo-configured type checker on that one file and reports what the checker did not report on that file at its previous successful check.

The baseline is intentionally **the same file, moments earlier, in the same tree** rather than Phase 1's merge-base comparison. That makes the signal affordable per edit while leaving Phase 1 the authoritative whole-project gate.

**What the comparison establishes, and what it does not.** Both runs share a checker, argv, working tree and interpreter, so the *environment* is identical — which is the whole of what makes the subtraction sound, and why this module needs no equivalent of Phase 1's `_environments_comparable`. It does **not** establish attribution, and the feedback text says so instead of claiming it. Two ordinary sequences make "your edit introduced N" false, and both were reproduced: `mypy <file>` follows imports, so editing `lib.py` then `app.py` reports `lib.py`'s error against the edit to `app.py`, whose bytes never changed; and `Bash` is not in `_EDIT_TOOLS`, so `sed -i`, `patch` and `ruff --fix` are invisible and surface at the next observed edit. The block therefore reads "N diagnostics on or reachable from X that were not reported when this file was last checked", naming both ways it can mislead.

The hook makes no cleanliness claim at all: a failed, timed-out, missing or unparsable run produces no feedback **and records no baseline**, so it can never become the thing the next edit is subtracted against.

Key bounds and ceilings:

- First edit to a file reports nothing; Phase 1 remains the backstop.
- Checks dependencies, not dependents, so caller breakage stays a Phase 1 concern. The feedback says this in the turn.
- Python only; TypeScript continues to rely on Phase 1, because a per-file `tsc` ignores the project's `tsconfig.json`.
- Each diagnostic is reported once; the post-edit result becomes the next baseline.
- 20s per checker run and 240s cumulative per attempt; exhaustion disables the hook for the rest of the attempt and says so once on the event stream.
- Per-file baselines use a 64-entry LRU. Eviction is safe where truncating a retained list would not be: an evicted file re-baselines and reports nothing that turn.
- Feedback is bounded in **bytes**, not characters — a checker quotes identifiers out of the repo under review, and eight 120-character diagnostics are 1528 bytes in ASCII but 3208 in CJK against a 1500 cap. Truncation happens on the encoded form, so it never splits a character.
- `hooks.per_edit_type` is off by default: a single-file type check is materially more expensive than a ruff call, and the issue's non-negotiable is advisory-first.

Hook ordering is load-bearing: type feedback is **third — after lint, before the scope guard**. `_compose_post_tool_hooks` short-circuits on the first hook that returns anything, and `check_scope` is a pure function with no dedup that warns on *every* edit to an out-of-plan file, while type feedback is once-only by construction. Behind the scope guard a coder working outside its plan would lose type feedback for the whole attempt, permanently; ahead of it the same collision only defers a scope warning that repeats anyway.

Two calls that are serialised rather than assumed: an `asyncio.Lock` wraps the read-run-write, because without it two overlapping calls each read the full remaining budget before either credits it, and the baseline read and write straddle an `await`. Whether the SDK ever dispatches PostToolUse concurrently is the SDK's business and is **not** established here — this is simply the first hook in the repo carrying mutable cross-call state and a cumulative budget, so the property is made true rather than assumed of the caller.

Nothing is written inside the repository. mypy's cache is relocated under the system temp root as **one directory per attempt, removed with the attempt**: a predictable parent created `0700` and refused unless it is ours alone, a random per-attempt child inside it (two attempts must not share one — mypy takes no cross-process lock), `weakref.finalize` to remove it, and reclamation of anything a dead run left behind by owner-pid liveness, the rule `core/worktree.py` already uses for per-run state. Warm within the attempt (2.19s cold, 0.42s warm on this repo's `orchestrator.py`) and deliberately not across attempts — it could not be, since the path it would key on is the per-run worktree and mypy keys its own cache on source paths.

**Also touches Phase 1, deliberately.** `_run_checker` now spawns the checker with no credential: the child environment is this process's with every secret-shaped variable removed (`_checker_env`, over `agent/child_env.drop_foreign_secrets` with an empty keep-list). Phase 1's own ceiling already records that a `plugins =` line makes the collector execute repo-authored code, and that code must never hold our OAuth token, API key or cloud credential; Phase 2 sharpens it by firing the same channel once per edit against a config the coder can write mid-attempt. `PATH`, `HOME` and the proxy variables stay.

That trade costs something real, so it is made visible rather than silent: a plugin that *reads* one of those variables no longer loads (`mypy_django_plugin` is the canonical case — it imports the settings module, which reads `SECRET_KEY`/`DATABASE_URL`), mypy exits non-zero, and the whole section would disappear behind a `log.warning`. It now renders a `TYPE EVIDENCE: NOT COLLECTED` line naming the cause and ending "This says nothing about whether the diff is type-clean; it says this check did not run." Every other failure still renders nothing at all, so absence keeps meaning absence. `NOT_COLLECTED_PREFIX` is excluded from `reviewer.py`'s READING SCOPE enumeration, which decides what the prompt carries by testing each evidence string for emptiness — without that exclusion the reviewer would be told it holds type diagnostics, and told not to re-derive them, for a diff nothing type-checked.

`mypy` is added to the `dev` dependency group (with `uv.lock` regenerated) so the two real-checker tests run in CI instead of skipping.

**AI-assisted, disclosed per `CONTRIBUTING.md`.** Written with Claude Code. The design decisions are mine and I can defend every line: the per-attempt cache tied to the hook's lifetime rather than to the repo path, pid-liveness reclamation over an age heuristic, the lock over a docstring caveat, excluding `NOT_COLLECTED_PREFIX` from `_reading_scope`, and keeping the credential scrub while making its cost visible rather than reverting it. Every guard added here carries a positive control — it is red on its own defect and green restored — and the mutations are listed under Evidence.

## Evidence

```
$ pytest -q -n 4                                    # this branch, real mypy on PATH
583 failed, 11263 passed, 658 skipped, 12 warnings, 1 error in 3313.25s (0:55:13)

$ pytest -q -n 4                                    # clean upstream/main (2ee83719), same PATH
580 failed, 11168 passed, 660 skipped, 11 warnings, 1 error in 3706.60s (1:01:46)

# Normalised failure ids: 584 mine / 581 baseline. 0 unique to baseline,
# 3 unique to this branch, all three accounted for:
#   2x test_reviewer_recall_runner.py — this clone's line endings, not the
#      branch. Those tests hash read_bytes() against a pinned git blob id;
#      core.autocrlf=true makes the file CRLF on disk (816300...) while the
#      manifest pins the LF blob 253e5a39..., which is what git stores and
#      what `git rev-parse :<path>` returns. They pass in an LF checkout.
#   1x test_scheduler.py — fails on clean upstream/main in isolation; 2 of its
#      tests fail there on three consecutive runs.

$ pytest tests/test_reviewer_recall_runner.py tests/test_scheduler.py \
         tests/test_type_hook.py tests/test_type_evidence.py \
         tests/test_structural_budget.py tests/test_egress_allowlist.py -q
1 failed, 341 passed, 5 skipped in 201.65s (0:03:21)
#   ^ a fresh LF worktree of this branch with this diff applied. The single
#     failure is the pre-existing test_scheduler.py one above; every
#     test_reviewer_recall_runner test passes, confirming the 2 above are the
#     checkout and not the change.

$ pytest tests/test_type_hook.py tests/test_type_evidence.py -q   # mypy on PATH
153 passed, 4 skipped in 42.89s

$ pytest tests/test_type_hook.py -q                                # no checker on PATH
85 passed, 4 skipped in 6.28s

$ pytest tests/test_type_hook.py --collect-only -q
89 tests collected in 1.49s

$ pytest tests/test_structural_budget.py tests/test_egress_allowlist.py tests/test_config.py -q
127 passed in 39.22s

$ uv lock --check
Resolved 81 packages in 1ms
```

Against a real mypy 2.3.1 (a throwaway venv; the project venv is untouched):

```
clean first edit                                 -> silent
`return 123` under `-> str`                      -> pkg/app.py:2:12 return-value
same diagnostic again                            -> not reported twice
diagnostic shifted 3 lines by an inserted import -> not reported
app.py byte-identical, lib.py broken             -> reported WITHOUT blaming the edit
.mypy_cache inside the repo                      -> False
cache dir after the hook is collected            -> removed
net new directories in the shared temp root      -> 0
cold / warm run on a 468-line file               -> 2.19s / 0.42s
```

Mutation checks — each guard is red on its own defect and green restored:

```
remove asyncio.to_thread around the checker call  -> test_the_checker_call_actually_hops_off_the_event_loop FAILS
drop type_hook from _compose_post_tool_hooks      -> AST call-site guard FAILS
remove --cache-dir (the phase 1 defect)           -> test_a_run_writes_nothing_inside_the_repo FAILS
restore keep=CODEX_CHILD_KEEP (the module default) -> 2 tests FAIL (behavioural + AST)
stop recording the plugin-import reason            -> gate-path NOT COLLECTED test FAILS
drop NOT_COLLECTED_PREFIX from _reading_scope      -> READING SCOPE test FAILS
count characters instead of bytes                  -> byte-budget test FAILS
remove the asyncio.Lock                            -> overlapping-edits budget test FAILS
remove the temp-root redirect from the fixture      -> suite errors on a clean TMPDIR
```

Gates: structural budget 18 passed (five entries re-frozen at the merge measurement, each with its reason beside it); egress allowlist green, with the `mypy` `plugins =` execution now declared against `review/type_evidence.py` — `grep -c plugins` on that file was 0 before this change; `RELEASE_MANIFEST` re-pinned from git's index blobs (never `--write`, which would rewrite every row with CRLF hashes on this clone) — 1570 rows verified, 0 mismatches, nothing tracked-but-unlisted.

Not verified here: the two on-disk POSIX cache tests skip on Windows, so they have only ever run on Linux. The ownership/mode rule itself is covered on every platform by `_dir_is_private` as a pure predicate. And this repo configures no type checker, so `per_edit_checker` returns `None` here even with the flag on — Phase 2's effect on task outcomes is unmeasured, and this PR claims nothing about it.

## Checklist

      checks for the file.)


eyalgolan (changes requested):
Prince, this is the strongest thing you've sent, and it's clearly built by someone who read what went wrong in phase 1 rather than just shipping phase 2. Off by default, a no-op unless the repo already configures a checker, bounded per run and per attempt with the budget billed in a `finally` so a run that raised still pays, `tsc` excluded because a per-file run can't honour tsconfig, and — the part I care most about — there is no "clean" utterance anywhere. Eleven paths return silence; exactly one speaks, and it needs a successful run, an existing baseline and real net-new diagnostics. A failed run records nothing, so it can't become what the next edit is subtracted against. I planted a false clean ("No new type errors in X") and six tests went red. That's the wall I wanted.

Three things before it lands. The first is the one that matters.

**The feedback blames the edit for things the edit didn't do.** The header says "Your edit to X introduced N diagnostics", and the module states the reason it needs no comparability check: *"both runs are the same checker, the same argv, the same working tree, seconds apart. The only thing that differs between them is the edit."* That last sentence isn't true, and I reproduced it two ways.

`mypy <file>` follows imports, so anything in the transitive graph is inside the comparison. I drove the hook with an `app.py` that did not change at all — md5 identical before and after — while `lib.py` gained an error:

```
app.py unchanged: True

[TYPE] Your edit to app.py introduced 1 type diagnostic(s) that mypy did not report
on this file before the edit. Fix them now, or say why they are intentional:
  lib.py:2:12 return-value Incompatible return value
```

And `_EDIT_TOOLS` is `{Edit, Write, MultiEdit, NotebookEdit}` — Bash isn't in it. So `sed -i`, a heredoc, `patch`, `ruff --fix` and `black` are all invisible to the hook, and whatever they broke surfaces on the *next* Edit, attributed to that edit.

Neither is a false clean, so it's not the unacceptable outcome. But it's a false signal carrying an imperative, delivered mid-turn, on sequences a real attempt produces constantly — and a coder that spends turns "fixing" something it didn't cause is exactly the attempt cost we're trying to avoid. The lint hook can honestly say "your edit to X introduced" because ruff sees one file; this hook can't inherit that sentence. Something like "mypy reports N diagnostic(s) on or reachable from X that it did not report when this file was last checked (a change elsewhere since then can be the cause)" is honest and just as actionable. And the soundness paragraph and the matching lines in `docs/verification.md` need to state the real invariant rather than "the only thing that differs is the edit".

**A citation that isn't there.** The module docstring says the pyright launcher's download and mypy importing a `plugins =` module are "both declared against this module in `tests/test_egress_allowlist.py`". Grepping that file: 8 hits for pyright, zero for plugins. One of the two is declared; the other isn't. It matters more here than in phase 1, because phase 2 fires that channel once per edit, in the live worktree, against a config the coder itself can write mid-attempt — and the subprocess inherits the whole environment, OAuth token included. Phase 1 has the same inheritance so it isn't a new capability, but the per-edit cadence plus coder-authored config is a different exposure and the docstring points at a file that doesn't cover it. Either add the line to the allowlist entry or drop the claim; passing a scrubbed `env=` to `_run_checker` would be a good thing regardless and helps phase 1 too.

**One test can't fail.** `test_a_run_writes_nothing_inside_the_repo` walks the repo before and after, but the fixture patches `subprocess.run`, so no process ever runs and no file could ever appear. Removing `--cache-dir` entirely — putting the cache back in the repo, the exact phase-1 defect — leaves that test green; only the cache-dir test and the real-mypy end-to-end catch it. And those real-process tests skip in CI, because mypy isn't in the dev dependency group, so the commit message's "including real mypy verification" holds only on a machine that happens to have it. Adding mypy to the dev group and making the fake checker actually create a file would close both.

Two smaller ones: the cache directory is created 0755 with `exist_ok=True` under a predictable name in the shared temp root, which on Linux is `/tmp` — worth `mode=0o700` and refusing a directory you don't own. And the `try` in `hook()` covers only the `to_thread` call, so a raising event sink escapes into PostToolUse; phase 1 wraps the whole body for that reason.

Cost looks fine, for the record: single-file mypy on this repo's 23k-line orchestrator is 2.1s cold and 0.1s warm, well inside the cap, and the warm-cache design is doing its job.

Fix the attribution sentence and the two claims and I think this lands. It'll need a rebase too — main has moved.

eyalgolan:
Re-checked after your merge of main, so you know where this stands rather than wondering.

The merge is clean and didn't cost you anything: `--strict` is fine at 1567 files and the gates I'd run pass, 221 tests. `type_hook.py` and `test_type_hook.py` are byte-identical to what I reviewed, so the three items from my last comment are still open rather than missed:

- the `[TYPE] Your edit to {file} introduced …` header, and the paragraph at the top of the module (and the matching lines in `docs/verification.md`) saying "the only thing that differs between them is the edit";
- the docstring's "Both are declared against this module in `tests/test_egress_allowlist.py`" — `grep -c plugins` on that file is still 0;
- `_fire` still patches `no_human.review.type_evidence.subprocess.run`, so `test_a_run_writes_nothing_inside_the_repo` still can't fail.

No rush and nothing new from me — just didn't want you to push again and find the same comment waiting.

eyalgolan:
Independent review: **PASS**. This is excellent work and I'd like to land it —
one small thing first, explained below.

What the review established, all by construction rather than by reading:

- **Genuinely off by default.** `per_edit_type: False` in `DEFAULT_CONFIG`, and
  the gate is the first statement of `_build_type_hook`, so under the default
  not even the checker-detection config reads happen. I re-verified this myself
  before writing.
- **The subprocess is properly handled.** List argv, no shell (`;touch owned.py`
  passes through as one literal argument); a repo-planted `repo/.venv/bin/mypy`
  first on `PATH` is refused by name; symlink escapes decline; timeout, crash and
  missing-binary all yield nothing and are billed in `finally`, and the
  cumulative attempt budget really stops it (6 edits against a hanging checker
  cost 20.45s total, with edits 5-6 costing 0.00).
- **The tests are real.** 11 single-change mutations, every one red on a named
  test. The two AST wiring tests (M9/M10) are the ones that earn their keep —
  they catch "the feature became dead code", which every behavioural test in the
  file would sail through.
- **No tamper.** The `test_structural_budget.py` numbers were re-measured with
  the scanner rather than trusted: 23875 / 3657 / 2261 / 251 declared, exactly
  equal to measured, no padding, and rebased on the correct merge-base values.

**The one thing I'd like changed before landing** is a documentation claim, not
code. The module docstring and `docs/verification.md` both say:

> the same checker, the same argv, the same working tree, the same interpreter,
> seconds apart… The only thing that differs between them is the edit.

Constructed counter-example: edit `app.py` (clean) → edit `dep.py` (introduces
an error there) → edit `app.py` again. mypy follows the import and the coder is
told *"Your edit to `pkg/app.py` introduced 1 type diagnostic(s)"* pointing at
`pkg/dep.py:7:3`. The diagnostic is real; the attribution sentence is not.
Anything else that touched an imported file between the two runs also differs.

A softened header, or that ceiling listed beside the other four, is all it needs.
I'm asking because a claim in `docs/verification.md` is read by operators
deciding whether to trust the signal — and I've sent back two of our own PRs
this week for the same class, so it would be inconsistent to wave it through
here.

Three follow-ups I'll file myself, not yours to fix:

1. The edited file's path is appended with no `--` separator and no leading-dash
   rejection, so a `.py` filename can occupy an option slot
   (`--python-executable=./pwn.py`). Demonstrated harm today is only silent
   self-disablement, and the coder already has Bash in that tree — but it's a
   one-line hardening.
2. The 20s cap bounds the subprocess only; 400k diagnostics took 22.5s and
   503 MB RSS. The 240s attempt budget still bounds the total and the feedback
   text is properly capped.
3. `_mypy_cache_dir` is a predictable 0755 name in the shared temp root and
   follows a pre-planted symlink — inert on macOS, worth hardening for
   multi-user Linux. `config.ensure_private_dir()` already does the right thing.

Also a robustness nit for you if you're touching it anyway:
`test_importing_the_hook_does_not_drag_in_the_review_package` shells out to a
bare `sys.executable -c`, which doesn't inherit pytest's `pythonpath = ["src"]`
— it fails for anyone running bare `pytest` without an editable install.

eyalgolan (changes requested):
This is a strong revision, and I want to be specific about that before anything else, because you did more than the review asked and the extra work is the good kind.

**All three blockers are closed, and I checked each one rather than reading the diff and nodding.**

The attribution fix is the real one — the block now says what is true, and both ways it can be wrong are named in the text the coder actually sees. Putting the old wording back turns two tests red. The credential scrub is the piece I had filed as a decision for me rather than work for you, and you did it properly, in `review/type_evidence.py`, which means it reaches the already-landed phase 1 too. I ran it against a loaded environment and it holds in both directions:

```
LEAKED (should be gone): []      # OAuth, ANTHROPIC_API_KEY, AWS_SECRET_ACCESS_KEY,
LOST  (should remain)  : []      # GITHUB_TOKEN, a password-bearing DATABASE_URL
dropped total: 10                # PATH, HOME, MYPYPATH, VIRTUAL_ENV, proxies kept
```

The cache-dir hardening I did not ask for at all, and it survives contact with the actual attacks rather than only its unit tests:

```
A. clean: True mode=700
B. symlink-to-dir refused: True
C. world-writable refused: True mode=777
```

Case B is the one I would have gotten wrong: `os.makedirs(exist_ok=True)` follows a symlink to a directory and returns happily, and the `os.lstat` is what catches it. Forcing `_dir_is_private` to return True takes five tests red.

Merge result against current main: manifest re-derives to exactly your files with `--strict OK: 1570 file(s)`, citation gate and context files both `VERDICT=OK`, and the full CI suite is **`12197 passed, 31 skipped`**. Two `test_reviewer_worktree.py` failures in my first run were my own fault — an earlier probe of mine left a real `/tmp/evil` directory on the box, which those tests use as a path that is *supposed* not to resolve. Removed it, `34 passed`. Nothing to do with your change, and I mention it only so the number in this comment is honest.

---

## What I need before this lands

None of these is a rewrite. Four are claims the code does not support, one is a real regression to already-shipped behaviour, and one is a test that does not defend the thing it is named for.

### 1. The mypy cache is per-RUN, and the docstring promises the opposite

`type_hook.py:192-197` says one directory per repo *"keeps the cache warm for the second edit onward … and stops a long-lived server accumulating a temp tree per attempt."*

The key is `os.path.realpath(repo_path)`, and `repo_path` is the attempt's worktree, not the project. The chain is `_acquire_worktree` → `_drive` → `_run_attempt` → `_build_type_hook(repo)` → `TypeFeedbackHook(repo_path=repo.path)`, and `_worktree_path` is documented in our own tree as *"Per-RUN worktree location … `<task_id>.<token>`"* with `token = <pid>.<random>`. Isolation is on by default (`worktree_isolation_enabled(DEFAULT_CONFIG)` → `True`).

So every run mints a fresh digest, gets a cold cache, and leaves a tree behind. Nothing reclaims them — `grep -rn "nh-typehook" src/ tests/ scripts/` returns hits only inside `type_hook.py`, and `teardown_worktree` does not know the name. On this machine right now:

```
1840 nh-typehook-mypy-* directories in the temp root
```

The warm-cache benefit is real *within* a run — I measured 4.6s cold then 0.2s warm on a 468-line file — so the design is sound and the 20s timeout is guarding the cold first run rather than the steady state. It is the cross-run claim that is false, and the leak that follows from it. Either key the cache on something stable (the task, or the worktree's upstream) and reclaim it in `teardown_worktree`, or keep it per-run and say so — but the sentence about not accumulating a tree per attempt has to go either way.

### 2. Our own test suite plants the exact hazard `_dir_is_private` defends against

`tests/test_type_hook.py:611-621` chmods a real directory to `0o777` in the shared temp root, under the predictable `nh-typehook-mypy-*` name, with no teardown. Right now:

```
23 world-writable nh-typehook-mypy-* directories
```

On macOS the temp root is per-user so this is untidy rather than dangerous, but `tempfile.gettempdir()` is `/tmp` on Linux CI, and a world-writable directory sitting at a predictable derived path is precisely the thing your new check exists to refuse. A `tmp_path`-based location, or a fixture that removes it, closes this.

### 3. `_checker_env()` silently switches off type evidence that ships today

This is the one I would call a regression rather than a gap, because it changes `_run_checker` for every review that already runs. A repo whose mypy plugin reads the environment — django-stubs is the canonical case, since `mypy_django_plugin` imports the settings module, which reads `SECRET_KEY` and `DATABASE_URL` — now fails to load its plugin. I built that repo and ran both paths:

```
A. via _run_checker (the shipped path):
   WARNING mypy exited 2: mypy.ini:2:1: error: Error importing plugin "myplug.py": 'DATABASE_URL'
   result: None
B. same argv, ambient env (today's behaviour):
   rc=1  app.py:2:12: error: Incompatible return value type (got "int", expected "str")
```

`None` means "we learned nothing", so the whole type-evidence section disappears from the review with nothing above a `log.warning` to say why. I want to be clear that I think the scrub is **right** and I do not want it reverted — a repo-controlled plugin is repo-controlled code, and it should never have held our OAuth token. What I need is that the loss is deliberate and visible: name it in the phase 1 ceilings, and let the reviewer see "the collector went dark because this repo's plugin wanted an environment variable we removed" instead of silence. A test on the gate path would help too — `grep -rn "_checker_env\|drop_foreign_secrets" tests/test_type_evidence.py` currently returns nothing, so the only coverage is on the phase 2 hook.

### 4. The credential test does not defend the `keep=()` it exists for

`tests/test_type_hook.py:911` asserts three hard-coded names. I swapped `drop_foreign_secrets(env, keep=())` for the module default and everything stayed green:

```
143 passed, 2 skipped
   OPENAI_API_KEY reaches checker: True | OAUTH: False
```

`llm.codex_auth_mode` defaults to `api_key`, so `OPENAI_API_KEY` is a live credential on a real share of installs, and it would reach the reviewed repo's mypy plugin with the whole suite passing. Asserting the keep-list itself, rather than a sample of names, makes the test bite.

### 5. `MAX_FEEDBACK_BYTES` counts characters

`type_hook.py:136` and `:363` — the name and the comment say bytes, `size += 1 + len(line)` counts characters, and `test_the_feedback_is_bounded_in_bytes` uses `"m" * 400`, so it structurally cannot see the difference. Eight diagnostics at 120 characters:

```
ascii  chars=1528  BYTES=1528  cap=1500
cjk    chars=1528  BYTES=3208  cap=1500
emoji  chars=1528  BYTES=4048  cap=1500
```

Type-checker messages quote identifiers and string literals from the repo under review, so non-ASCII is not exotic. `len(line.encode())` fixes it. The smaller ASCII overshoot is a separate thing and you already documented it deliberately — the comment above the `+ 400` says the bound is on the quoted block rather than the whole string, which is fair. It is the character-versus-byte gap that the `"m" * 400` fixture cannot reach.

### 6. The attempt budget is not the bound `_budget_left` says it is

`type_hook.py:297-306` states *"the cumulative cap is a bound and not a suggestion the last run may overshoot"*, but `self._spent` is only credited in the `finally` after a run completes, and there is no lock or reentrancy guard anywhere in the module — `grep -n "Lock\|reentran\|serial"` finds nothing. Any two `hook()` calls that overlap each read the full remaining budget, and both `previous = self._baseline.get(...)` and `self._baseline[relative] = after` straddle an `await`, so the run that finishes last wins regardless of which observation is newer.

I am being careful here: whether the SDK ever dispatches two PostToolUse callbacks concurrently is a property of the SDK, and I have not established that it does. But this is the first PostToolUse hook in the repo carrying mutable cross-call state and a cumulative budget — `LintFeedbackHook` is entirely stateless — so the property the docstring asserts is not established anywhere. Either an `asyncio.Lock` around the read-run-write, or a docstring that says the bound holds only under serial dispatch. I would take either.

---

## The text pass

Same class as the blockers you just fixed, in four places that survived.

1. **The PR body still carries the removed claim** — "reports only diagnostics **attributable to the edit**". We squash, so the body becomes the commit message on main. This one costs nothing to fix: no push, no CI, no manifest re-pin.
2. **The body's evidence block is stale** — "67 tests … (2 skipped when no checker is installed)". The file now collects 83, and with `mypy` in the dev group the two no longer skip, which was the point of adding it.
3. **`type_hook.py:9`** still ends "silent unless it has something **attributable** to say". Twenty-two lines below, `:31` says attribution is precisely what the run identity does not buy.
4. **`type_hook.py:79-80`** — "Both are declared against **this module**". They are declared against `review/type_evidence.py`, and correctly so, because this module spawns nothing itself. A reader who follows that sentence greps for `type_hook` in the allowlist and finds nothing, which is the dead end I hit.

Two smaller ones while you are in there: `:119` says the 20s cap is *"an order of magnitude under phase 1's 90s"* when it is 4.5x, and the `S_ISLNK` clause at `:185` is unreachable — under `os.lstat` a symlink already fails `S_ISDIR`, and deleting it leaves the suite green. The clause is harmless, but the docstring and the test prose read as though it is load-bearing, and it is the `lstat` that is doing that work.

---

## Two things for the record, not for you to fix

Adding `mypy` to the dev group does exactly what you said. In my venv the two real-checker tests skip; with a real mypy on PATH, `83 passed in 11.69s`, zero skips.

And this repo configures no type checker, so `per_edit_checker` returns `None` here even with the flag on. That is the feature behaving as documented, not a bug, but it means we cannot dogfood phase 2 on ourselves, so its effect on task outcomes stays unmeasured. Your PR does not claim otherwise, which I appreciate. Closing that is my job, not yours.

---

I know this is a long list on a revision that fixed everything I asked for last time, and I am conscious you have pushed to this branch repeatedly because main keeps moving underneath you. Items 1 through 6 are the ones that need a decision from you; the text pass I am happy to write myself and push as a follow-up if you would rather not spend another round on wording. Say which you prefer and I will do that half.

The security work here was above what the review asked for, and the parts I checked hardest are the parts that held up best.

eyalgolan:
Heads-up on a conflict rather than a new review — your head is unchanged at `a1385b943`, so the six items from my last round still stand as written and I'm not going to repeat them.

Main moved and this branch is now dirty. I trial-merged it so you know exactly what you're walking into; it's small:

```
CONFLICTED:
  RELEASE_MANIFEST.txt            (generated, as always — reset to main's and re-derive)
  tests/test_structural_budget.py (1 hunk)
```

The real one is a single `FROZEN_FILE_LINES` entry that both sides moved:

```
yours : 23828 -> 23875 (+47)   #114 phase 2
main  : 23828 -> 23846 (+18)   the already-satisfied sibling-branch ordering fix
```

Please don't just add the two deltas — re-measure on the merge result with the scanner's own metric. That comment in the file is right that it is AST/splitlines based rather than `wc -l`, and the two differ by a constant 3 lines on that file, so an arithmetic guess is the kind of thing that goes red in CI for a reason nobody can see locally.

No rush from my side, and sorry that the manifest and the budget file make this repo's contributor experience what it is — that churn is mine to fix, not yours to keep paying for. When you do push, the four text items are the cheap half and two of them need no push at all (they're in the PR body, which we squash into the commit message). Still happy to take the text pass myself if you'd rather spend your time on items 1 through 6; just say which.

eyalgolan (dismissed):
Approving. All six went in, and I checked each one by running it rather than by reading the diff.

**1 — the cache.** This was the one I cared most about, and the new design is better than the fix I asked for. Per-attempt child under a `0700` parent, removed with the attempt, with a dead-owner sweep for anything a crashed run left behind. I ran the lifecycle end to end in an isolated `TMPDIR`:

```
1. parent mode: 0o700   per-attempt child: 80260.pzqa2oxc
2. child after the hook is dropped: 0        (the finalizer fires)
3. dead-pid orphan removed: True | live-pid orphan kept: True
```

The live-pid case is the half that matters — a sweep that reclaimed everything would take a running attempt's cache out from under it. Keying reclamation on owner-pid liveness the way `core/worktree.py` already does is exactly right, and saying plainly that a cross-run cache *could not* work because the path is minted per run is better than quietly dropping the claim.

**2 — the suite no longer plants a world-writable directory in the shared temp root.** Before and after running the file: `25 → 25`, and those 25 are my own leftovers from yesterday's runs of the old code, not new ones.

**3 — the scrub's cost is now visible.** `TYPE EVIDENCE: NOT COLLECTED` naming the cause, while every other failure still renders nothing, is the right line to draw: absence keeps meaning absence, and the one case where we *know* why gets to say so. Naming `mypy_django_plugin` as the canonical instance makes it findable.

**4 — `MAX_FEEDBACK_BYTES` counts bytes now.** The defect was 2.7x over the stated cap on non-ASCII; measured on this head:

```
ascii  BYTES=1528   cjk BYTES=1445   emoji BYTES=1445   cap=1500
```

CJK and emoji are inside the cap. ASCII sits at 1528 because the two trailer lines are appended after the budget, which your comment already documents deliberately — that part I'm happy with as-is.

**5 — the budget is a bound now.** Ten concurrent hooks against a 3s budget, with each run sleeping 2s:

```
runs executed=2   total billed=4.0s   exhausted=True
```

Before the lock that was 10 runs and 20s against the same budget. One run of overshoot is the documented behaviour — a run that cannot start inside the deadline does not start — so this matches what `_budget_left` claims rather than contradicting it.

**6 — the two prose items.** The dead `S_ISLNK` clause is gone with a comment explaining that the `lstat` was doing the work, "4.5x" replaces "an order of magnitude", and the allowlist sentence now says the channel is declared against `review/type_evidence.py` and that grepping for `type_hook` is a dead end by design. That last one is the fix I wanted: it tells the next reader why the absence is correct instead of leaving them where I ended up.

**Gates on the merge result:** manifest re-derives to exactly your files, `--strict OK: 1570 file(s)`; `195 passed, 2 skipped`; `tests/test_type_hook.py` alone is `89 passed, 0 skipped` with a real mypy on PATH.

---

**One housekeeping note, and I did it rather than sending you round again.** The PR body still carried the two stale lines from my earlier round — "reports only diagnostics attributable to the edit", which is the exact claim the code was rewritten to stop making, and "67 tests" when the file now collects 89 with nothing skipped. Since we squash, that text becomes the commit message, and since editing a body needs no push and no CI, I've edited it myself: the first sentence now says it reports what is newly present and names both ways the attribution can be wrong, and the count line reads 89 with a note that mypy being in the dev group is why nothing skips. Your words everywhere else. If you'd rather phrase either differently, change it — it's your PR and I won't touch it again.

Nice work on this one. Six findings including two that reached already-landed code, and the answers were better than the asks on at least two of them.

PrinceXDev:
Thank you, @eyalgolan 

Your `89 passed, 0 skipped` closes the one gap I flagged as unverified. The two on-disk cache tests are POSIX-only and skip on Windows, so I had never executed them — the ownership and mode rule was covered on every platform only because `_dir_is_private` is a pure predicate over a stat result. Your run is the first time the filesystem half has actually run. Same for the live-pid case: I asserted it with a faked `pid_alive`; you exercised the real one.

On the ASCII 1528: agreed, leave it. The two trailer lines are deliberately outside the budget because they are fixed prose this module writes, not unbounded input from the checker — and moving them inside would mean a pathological message could push the "not its callers" ceiling out of the block, which is the line least affordable to lose. If you ever want the whole string bounded instead, it is one call to _nbytes on the assembled text, and I'll send it as a follow-up.

One note for whoever picks up phase 4: `_checker_env` now applies to the gate path too, so any repo whose checker plugin reads the environment will show `TYPE EVIDENCE: NOT COLLECTED` rather than evidence. That is intended, but it is a behaviour change to something already shipped, and the ceiling naming `mypy_django_plugin` is where it is written down.

PrinceXDev:
@eyalgolan 

The red `Python` job here looks inherited from `main`, not introduced by this branch.

- `main` is currently red: CI #640 (`2f141aa`, #262 "Derive attempts.completed_at
  from the terminal status") fails the `Python` job with `Event loop is closed`.
  Every `main` run before it is green.
- This branch merged `2f141aa` in `d867f0c`, so it carries that state.
- Nothing here touches the database or its concurrency:
  `git diff --name-only upstream/main...HEAD` matches no `db`/`concurrency` path,
  and `core/db.py`'s last change is #262. `tests/test_db_concurrency.py` is
  untouched by this PR.
- On the merge result locally, `test_pool_and_peer_connection_never_produce_a_lock_storm`
  passes 20/20 (5 isolated + 15 stress), and `tests/test_db_concurrency.py` is
  `20 passed` both serially and under `-n 4`.

I haven't opened this run's log, so I'm not claiming it's the same assertion as
#640's — if it is, this should go green on a re-run once `main` is fixed, and I'm
happy to merge `main` again and push at that point.

For whoever picks up #262: the lead is `Store.update_attempt`. The C2 block does a
`SELECT` and then the `UPDATE` + `COMMIT`, and the failing test calls
`update_attempt(attempt_id, status="failed")` with no reason — exactly the path that
takes that `SELECT`. `@serialized_write` serialises in-process coroutines but cannot
stop a peer connection advancing the WAL between that snapshot and the write
(`SQLITE_BUSY_SNAPSHOT`), and #262 inserted `stamp_completion(fields)` between the
two. I could not reproduce it on Windows, so treat that as a lead rather than a
diagnosis.

Unrelated to the above, the other gates on this head are green: `File inventory`
passes at 1573 files, and locally `345 passed, 4 skipped` across the phase-2 suites,
the structural budget, the egress allowlist and #262/#263's own new tests.

eyalgolan (dismissed):
Re-approving — the approval was dismissed by your merge of main, and nothing you contributed changed.

I checked that rather than assuming it. All seven files you touch are byte-identical to the head I approved:

```
type_hook.py  type_evidence.py  test_type_hook.py  test_egress_allowlist.py
docs/verification.md  pyproject.toml  test_structural_budget.py   -> all UNCHANGED
```

So the six findings stay closed and I haven't re-litigated them. On the new merge result: it merges clean now, the manifest re-derives to `--strict OK: 1573 file(s)`, and `195 passed, 2 skipped`.

Yours to merge. Sorry for the extra round-trip — that one was the repo's manifest churn, not anything you did.

eyalgolan (approved):
Conflict resolved, and I did it on your branch rather than asking you to — you've rebased this three times today already and the churn is ours, not yours.

What I pushed is one commit sitting directly on `main`, with you as author and me as committer, which is how an external contribution lands here:

```
author:    PrinceXDev <princepanchani890@gmail.com>
committer: eyalgolan <5146175+eyalgolan@users.noreply.github.com>
parent == origin/main:                    yes
tree identical to the verified merge:     yes
Co-authored-by / tool trailers:           0
```

The only conflict was `RELEASE_MANIFEST.txt`. I resolved it the way this repo requires rather than by picking a side: reset the manifest to main's, re-derive it from the merged tree, then confirm the rows that differ are **exactly** your files and nothing else — 14 rows, 14 files, exact match. `--strict` on its own then gives `OK: 1575 file(s)`.

Nothing of yours changed. The squash carries the same tree I reviewed and approved.

Re-approving, since the force-push dismissed the last one. The merge itself is waiting on one thing: a full suite run against this exact squashed tree. That is deliberate — a conflict round produces a tree that neither your CI nor main's has ever tested, and landing one of those untested is a mistake this repo has made before. It is running now; when it is green this goes to `main` as that same sha, so GitHub will show it merged rather than closed.

Thank you for the patience on this one. Six findings, two of which reached already-landed code, and you answered every one of them without argument.

PR #229, opened 2026-09-10
feat: measure navigation value before building symbol server

PrinceXDev:
## What this changes

Phase 3 of #114 is a gate, not a feature — "measure navigation value before building" — so this adds the instrument and the verdict, and deliberately does not build the symbol server it gates.

`scripts/navigation_value.py` measures what share of the coder's file-read mass sits in reads a `definition`/`references`/`hover` call could have answered instead, net of what the symbol answer would itself cost; on the only corpus available it returns **HALT** at 10.4% raw / 13.7% weighted against a pre-registered 15% gate, and `docs/NAVIGATION_VALUE.md` records the method, the measurement and what would reopen the phase.

Part of #114 (phase 3). Independent of the phase-2 branch: nothing here imports from the product or references a phase-2 file, and the two changes share no file except `RELEASE_MANIFEST.txt`, which every PR re-pins — so this can be reviewed without waiting on #221.

## Evidence

```text
$ .venv/Scripts/python.exe -m pytest -q tests/test_navigation_value.py
109 passed in 3.16s

$ .venv/Scripts/python.exe -m pytest -q -n 4 -m repoguard
142 passed, 13 skipped, 8 warnings in 14.90s

$ .venv/Scripts/python.exe -m pytest -q -n 4 tests/test_navigation_value.py \
    tests/test_doc_anchors.py tests/test_readme_claims.py tests/test_test_lanes.py \
    tests/test_structural_budget.py tests/test_egress_allowlist.py \
    tests/test_check_release_manifest.py tests/test_scope_guard.py \
    tests/test_check_context_files.py tests/test_derived_conflict_inventory.py \
    tests/test_deselect_lists_agree.py tests/test_measure_cache_burn.py \
    tests/test_review_round_value.py
363 passed, 19 skipped in 47.16s

$ python scripts/navigation_value.py --source transcripts
source: transcripts (read-only), 112 session(s), 114 context window(s), 828 read(s), 0 with no recorded result
thresholds: large read >= 12,000 chars, lookback 3 call(s), symbol answer estimated at 2,000 chars

         class   reads         chars   %chars          weighted     %wtd
------------------------------------------------------------------------
           all     828     2,914,283   100.0%       293,218,423   100.0%
 symbol_lookup      15        70,426     2.4%         5,719,223     2.0%
    whole_file      17       307,137    10.5%        42,964,293    14.7%
     navigable      31       359,766    12.3%        46,939,410    16.0%
          both       1        17,797     0.6%         1,744,106     0.6%
     non_code     438     1,487,908    51.1%       108,173,281    36.9%
        repeat     248       606,935    20.8%        68,162,865    23.2%
      windowed     315       689,346    23.7%        71,230,577    24.3%
      sidechain       0             0     0.0%                 0     0.0%

net addressable (navigable mass less the estimated symbol answer): 302,550 chars (10.4% of read mass), 40,133,641 weighted (13.7%)

VERDICT: HALT
  net addressable read mass is 10.4% raw and 13.7% weighted, both below the pre-registered 15% gate. Phase 3's terms halt development on a negative result: no symbol server, and phase 4 (polyglot wiring evidence), which the issue makes contingent on this, stays closed.
  CAUTION: this verdict is not robust to the large-read threshold. Re-deciding at 6,000 chars gives PROCEED, 24,000 chars gives HALT. The decision above is the one the pre-registered threshold gives, which stops the number being tuned to the answer — it does not make the answer a property of the data. Do not close the phase on it without the authoritative population.
```

Mutation check — 24 single-line defects injected into the classifier, verdict and both readers; 23 fail a test, and the one survivor is equivalent code for the input it touches (two faithful spellings of that same defect, `str(block)` and `str(content)`, are both caught):

```text
CAUGHT    non_code gate removed (whole_file)          CAUGHT    coder-role filter removed
CAUGHT    non_code gate removed (symbol_lookup)       CAUGHT    bool accepted as a result size
CAUGHT    lookback unbounded                          CAUGHT    prose token bound removed
CAUGHT    windowed uses truthiness                     CAUGHT    Glob counted as a symbol query
CAUGHT    net saving not clamped at zero               CAUGHT    extension taken off the whole path
CAUGHT    units no longer required to agree            CAUGHT    leading-dot filename read as an extension
CAUGHT    corpus floor removed                         CAUGHT    unsized read treated as zero-length
CAUGHT    empty corpus reads as a verdict              CAUGHT    transcript compaction marker ignored
CAUGHT    subagent reads weighted from the main window CAUGHT    per-task windows collapsed
CAUGHT    turns counted over reads only                CAUGHT    large-read threshold made exclusive
CAUGHT    attempt/compaction boundary ignored          CAUGHT    negative thresholds clamped not rejected
CAUGHT    non-dict tool_input drops the whole call     SURVIVED  image block counted as text (equivalent code)
```

Two findings the tests produced rather than confirmed, both recorded in the doc and pinned by regression tests:

* **The first run returned PROCEED at 27.6%, and it was wrong.** `.md` was the largest extension in the corpus by read mass and no symbol call answers a question about a README. With a closed allowlist of languages a symbol server serves, the same corpus reads 10.4% / 13.7% and the verdict inverts. One constant flipped the phase gate, so a fixture pair identical except for the file extension now pins it (`.py` → PROCEED, `.md` → HALT).
* **A Grep pattern containing a colon classified as a go-to-definition.** The argument-list/annotation tail split kept only the text before the colon, so `TODO: fix the reviewer before the next release` normalised to `TODO` and filed itself into the strongest class. Bounded to two tokens after the declaration prefix, with both spellings in the rejection table.

**What I could not verify locally, stated rather than implied.** A whole-suite run on this Windows checkout is not a usable signal: unmodified `HEAD` fails in the hundreds here from two environment causes documented in CONTRIBUTING.md — `pytest` needing to be on `PATH` for the tests that shell out to it, and `doc/` source reads mis-decoding under a cp1252 console.

I confirmed the two largest failure clusters (`tests/test_approve_merge.py`, `tests/test_structural_budget.py`) fail identically on a clean detached-`HEAD` worktree with none of this change present, and a paired baseline-vs-working comparison wedged at 92% in the `-n` hang CONTRIBUTING.md warns about.

So instead of a whole-suite claim, the runs above cover every test that can observe a new file: all 154 `repoguard` nodes plus every tree-scanning guard (inventory, doc anchors, README claims, lane partition, structural budget, egress allowlist, scope guard, context-file limits). CI's Linux `Python` job is the reference measurement.

## Checklist

* [x] Tests cover the change, and they fail on the unfixed code.
* [x] Test count and assertion count did not go down, or the PR body explains why.
* [x] Lockfiles committed if a dependency changed. *(No dependency changed — the script is standard-library only and imports nothing from the product, so it runs against a database or a log written by any version.)*
* [x] Nothing in the diff reads or writes a credential inside the repo. *(The script opens SQLite `mode=ro` and never writes; it prints no path, Grep pattern, session id or file content, only aggregates — pinned by `test_no_path_pattern_or_session_id_reaches_the_output`.)*
* [x] This does not conflict with the project's standing constraints: the agent never merges (it opens a PR and stops; merge is a human action), review stays evidence-based (independent fresh-context reviewer, never a numeric self-scoring gate), and the locally installed product keeps its lean stack (SQLite only, no extra services, and no re-implementation of what the coding backend's SDK already provides). *(Nothing here ships in the product: one `scripts/` measurement reader, its tests, and a report doc. The phase's whole point is to withhold the symbol server until the measurement justifies it — the lean-stack constraint is the reason the gate exists.)*
* [ ] First PR only: I have read `CLA.md` and added my `contributors/<handle>.md` entry to this PR. *(Not a first PR — `contributors/princexdev.md` is already in the tree.)*


eyalgolan (changes requested):
Prince, this is the right instinct and mostly excellent work. Pre-registering the gate before running it, refusing to build the thing until the number says so, the `non_code` refutation, the "what this does not know" section — that's the discipline I wanted from phase 3, and the prose is honest throughout. It's also cheap: 109 tests in 1.1 seconds, no new dependency, no subprocess, no egress.

But there's a problem in the instrument, and it's the kind that's hardest to see in your own work: on this repo's actual data, the class the script calls its strongest is structurally empty.

`SYMBOL_QUERY_TOOLS = frozenset({"Grep", "Search"})`. no_human's coder never emits either. It searches through Bash. From the fleet database, coder events only:

```
Bash 115,776 | Read 35,557 | Edit 18,425 | Write 3,010
Grep 0 | Glob 0 | Search 0
Bash calls containing grep/rg/ag/ack: 50,459
```

So `symbol_lookup` can only ever be 0 here, and the verdict falls back entirely to the size heuristic. I ran your script against a copy of the real database and it does render a decision on that:

```
symbol_lookup   0   0   0.0%   0   0.0%
net addressable: 21,122,813 chars (12.1% of read mass), 559,060,393 weighted (10.9%)
VERDICT: HALT
  robustness: the same decision at 6,000, 24,000 chars, so it is not an artifact of where 'large' starts.
```

Two things follow. The corpus clears both your floors, so the script doesn't hold back — it decides, and it decides HALT, on evidence where its own strongest signal cannot exist. And the NOTE explains the zero the wrong way round: it says a symbol question answered from memory leaves no search behind, when the actual reason is that this source records no search-tool calls at all. Those are opposite facts, and nothing in the output tells them apart. A reviewer running this would read a confident negative and close phase 3.

I don't know which way it lands once Bash greps are counted — a rough pass suggests it could cross the gate, but that's a ten-line extractor, not a classifier, and I wouldn't ship its number either. The point is that 12.1% isn't robust to the single most obvious modelling choice.

Three things:

- **Count the coder's real search channel.** A Bash command containing `grep`/`rg` is a search; extract its pattern and feed it through your existing `is_symbol_query`, which is already good. Until then, refuse a verdict when the corpus holds zero search calls, the same way `EmptyInputSet` refuses on zero reads.
- **Print the search-call census, and make the zero fail closed.** Your `--source events` header already reports sessions, windows and reads; add searches. "0 searches seen" and "0 of 412 searches were symbol-shaped" should not render the same output.
- **Pin the headline denominator.** `total = classes["all"][unit]` has no test. Pointing it at `classes["navigable"]` gives 88.2% and PROCEED; using `chars` for both units gives a 319.8% share; both leave all 109 tests green. One test asserting the share equals net-addressable over total read mass, on a fixture with known non-navigable mass, closes it.

Two smaller ones: the 24,000-char robustness probe only qualifies 21 of 35,557 reads, so it can only return HALT and isn't independent evidence — print the qualifying count per probe. And `docs/NAVIGATION_VALUE.md` says the product "already recorded everything required", which is the one claim in the doc that isn't true, since task_events records no search calls; your commit message is honest about the data gap, so it's just the doc that's out of step.

Everything else checked out: the role filter really does isolate the coder, the class overlaps are disclosed rather than double-counted, and 13 of 15 mutations I planted go red on the right tests.

Fix the search channel and re-run it — whichever way the number lands, it becomes a number we can act on.

PrinceXDev:
<html>
<body>
<html><head></head><body><p>Thank you @eyalgolan  the core finding is right, and it was worse than a reporting bug: it changed the answer. Fixed, plus two things the fix turned up.</p><p><strong>1. The coder's real search channel.</strong></p><p>Confirmed in the code as well as the data: <code inline="">core/prompt_blocks.py</code> instructs the coder to <em>"locate the relevant lines with <code inline="">grep -n</code>"</em> for a large file, so the shell is the sanctioned channel and <code inline="">Grep 0</code> is structural rather than incidental.</p><p><code inline="">SHELL_TOOLS</code> now reads <code inline="">grep</code>/<code inline="">rg</code>/<code inline="">ag</code>/<code inline="">ack</code>/<code inline="">ugrep</code> and <code inline="">git grep</code> out of shell commands, per segment, split with the same expression <code inline="">orchestrator._looks_like_test_run</code> uses and for the same measured reason. Patterns feed the existing <code inline="">is_symbol_query</code> unchanged in role.</p><p>The option tables are <strong>per binary</strong>, because one shared table cannot be right: <code inline="">-r</code> takes no value in grep and is <code inline="">--replace</code> in ripgrep, so a single set either eats <code inline="">grep -r PATTERN</code>'s pattern or misreads <code inline="">rg -r NEW PATTERN</code>.</p><p><code inline="">--color</code> is deliberately absent — its argument is optional and almost always written <code inline="">--color=auto</code>, so listing it would eat the pattern in the commoner shape. Both directions are pinned in a 26-row table.</p><p>On the transcript corpus this quadrupled the class:</p><p><code inline="">symbol_lookup</code> <strong>15 reads / 2.4% of read mass → 61 / 7.1%</strong>.</p><p><strong>2. Census printed; the zero fails closed.</strong></p><p><code inline="">NoSearchChannel</code> refuses a corpus with reads and no searches, the same fail-closed rule as <code inline="">EmptyInputSet</code>, one signal over — an absent channel is a gap in the source or the instrument, never evidence about the agent.</p><p>The census is now unconditional:</p><pre><code class="language-text">search channel: 1477 search call(s), 1026 with an extractable pattern, 450 symbol-shaped
</code></pre><p>You were right that the NOTE explained the zero backwards. It now has four branches:</p><ul><li><p>no channel at all</p></li><li><p>searches but no extractable pattern</p></li><li><p>patterns but all text</p></li><li><p>symbol searches that no read followed</p></li></ul><p>Only the third and fourth say anything about the agent.</p><p>The middle census number is the parser's own <strong>coverage</strong>: <code inline="">grep -f patterns.txt</code>, or a command with an unmatched quote, counts as a search and never as a symbol one, so a gap in my option tables can never read as "this corpus does not search".</p><p><strong>3. The denominator.</strong></p><p>Pinned on a fixture carrying real non-navigable mass, with both of your mutants named in their own assertions, plus the property that a subset's share cannot exceed 1.0.</p><p>Both go red now.</p><p><strong>4. Probe qualifying counts.</strong></p><p>Every probe carries <code inline="">navigable_reads</code>, and the agreeing branch now says so explicitly: a probe that qualified far fewer reads than the shipped threshold could not have disagreed, so read its count before reading its agreement as corroboration.</p><p><strong>5. The doc claim.</strong></p><p>Corrected. It no longer says the product "already recorded everything required"; it says what <em>is</em> recorded — reads, result sizes, and search patterns inside the recorded Bash commands — and what is not:</p><p><strong>any search's hits</strong>, which is why <code inline="">symbol_lookup</code> stays adjacency rather than causation on either source.</p><hr><p><strong>Two things the fix turned up, which you did not ask for.</strong></p><p>I read what <code inline="">is_symbol_query</code> actually <em>accepted</em> across the 1,470 real patterns rather than reasoning about it, and found more false positives in the same load-bearing class.</p><p>Quote-stripping turned the fragment <code inline="">"Test</code> — a token the unbalanced-quote fallback produced — into <code inline="">Test</code>, and the deliberate <code inline="">'"version"'</code>, a search for a quoted JSON key, into <code inline="">version</code>.</p><p>The <code inline="">=</code>/<code inline="">:</code> tail split reported <code inline="">kind="tool_result"</code> and <code inline="">risk:</code> as name lookups.</p><p>Quotes are no longer stripped, the tail is cut only when a declaration keyword introduced it, and an untokenizable segment now yields a search with <strong>no</strong> pattern rather than a guessed one — otherwise <code inline="">grep -rn parse_config " .</code> hands <code inline="">parse_config</code> to the strongest class off a command that was never parsed.</p><p>I also <strong>deleted two guards as unreachable</strong>.</p><p>Once quotes stopped being stripped and the tail became declaration-gated, an explicit quote check and a token-count bound could no longer change any outcome, because <code inline="">_IDENTIFIER</code> is anchored and whitespace and quotes are already impossible past it.</p><p>Mutation testing is what exposed them — both "survived", because they were dead.</p><p>A guard whose comment claims to prevent a defect it cannot reach tells the next reader the wrong thing about what protects what, so they are gone and the cases they named are pinned as behaviour in the test table instead.</p><hr><p><strong>The re-measurement, and it is less robust than before, not more.</strong></p><p>Still HALT, at <strong>11.2% raw / 14.6% weighted</strong> against the pre-registered 15% gate — weighted misses by 0.4 points.</p><p>But the decision now flips on <em>all three</em> of the instrument's estimated parameters, where before it flipped on one:</p>
Parameter | Value that flips the verdict
-- | --
Large-read threshold | 6,000 chars → PROCEED
--symbol-response-chars | 0 → PROCEED (1,000 → INCONCLUSIVE)
--lookback | 6 calls → INCONCLUSIVE

<p><code inline="">docs/NAVIGATION_VALUE.md</code> carries all three sensitivity tables and states plainly that this corpus cannot decide the phase.</p><p>It also now names the largest known over-count in the class the verdict leans on: a bare English word is indistinguishable from a symbol name, and <code inline="">warning</code>, <code inline="">snapshots</code> and <code inline="">node_modules</code> are counted alongside <code inline="">parse_config</code> and <code inline="">_run_at_commit</code>.</p><p>No rule over the pattern alone separates them; separating them needs the search's hits, which the product deliberately does not record.</p><p>So I have not moved the verdict off HALT, and I would not close the phase on it.</p><p>Your rough pass was right that it could cross the gate — weighted does, at two of the three parameter settings above.</p><hr><p><strong>Verification.</strong></p><p>Tests <strong>109 → 167</strong>, none removed.</p><p>Mutation: <strong>37/37 applied mutants caught, zero survivors</strong>, including your three; the only NO-PATCH is the mutant targeting the dead code I deleted.</p><p><strong>154 <code inline="">repoguard</code> nodes</strong> and every tree-scanning guard pass.</p><p>One correction to my own PR body while I am here.</p><p>I had called the <code inline="">tests/test_scope_guard.py</code> symlink failures pre-existing on the strength of a baseline worktree, and that worktree was at <code inline="">a2e9a4a7</code> — those two tests were added by <code inline="">d249521e</code>, this branch's parent.</p><p>Re-checked at <code inline="">d249521e</code> itself: identical failures with none of this change present (<code inline="">WinError 1314</code>, a Windows symlink privilege).</p><p>The conclusion stands; my first check did not establish it.</p><p>The decisive run is still the one I cannot make — <code inline="">--source events</code> against a <code inline="">task_events</code> database with real coder volume, which now has a search channel to read.</p><p>If you re-run it, the census line is the first thing to check:</p><p>If <code inline="">searches</code> is 0 there, the shell parse is still missing that estate's channel and the script will refuse rather than answer.</p></body></html>
</body>
</html>

eyalgolan (approved):
Approving. You didn't just fix the finding, you ran it down further than I did and it changed the answer.

I made the run you can't make. Your script, your thresholds, `--source events` against a copy of the fleet database:

```
search channel: 44072 search call(s), 18335 with an extractable pattern, 12627 symbol-shaped
 symbol_lookup    6298    27,361,241    15.6%       762,318,040    14.8%
net addressable: 34,778,488 chars (19.8% of read mass), 948,684,259 weighted (18.4%)
VERDICT: PROCEED
  CAUTION: this verdict is not robust to the large-read threshold. Re-deciding at 6,000
  chars gives PROCEED on 8073 navigable read(s), 24,000 chars gives HALT on 6308.
```

Before the fix that same corpus reported `symbol_lookup 0` and a confident HALT. So the class that was structurally empty is now the largest single contributor, and the phase gate inverts. That's the whole reason the finding mattered, and you closed it.

I checked the three things I asked for rather than taking them on trust. Pointing the denominator at `navigable` now fails `test_the_share_denominator_is_total_read_mass`; using chars for both units fails three tests including the share-cannot-exceed-one property. Removing the `raise NoSearchChannel` fails `test_a_corpus_with_reads_and_no_searches_is_refused_not_decided`, and the refusal is deliberately not a subclass of the empty-corpus one, so the two can't be confused. 167 tests, 1.4 seconds, still no dependency and no subprocess.

Two things I want to say about the parts you weren't asked for.

Reading what `is_symbol_query` actually accepted across 1,470 real patterns, rather than reasoning about it, is the right instinct and it's how you found `kind="tool_result"` and `risk:` being counted as name lookups. And deleting two guards because mutation testing showed they were unreachable is the correct call — a guard whose comment claims to prevent something it can't reach is worse than no guard, because the next reader believes it. Pinning the cases they named as behaviour instead is exactly right.

The other is that you reported the result getting *less* robust, not more, and said plainly you wouldn't close the phase on it. The caution in the output names the qualifying read counts per probe, so a reader can see that the 24,000 probe had fewer reads to work with. And the honest statement that a bare English word is indistinguishable from a symbol name without the search's hits — which the product deliberately doesn't record — is the limit of what this instrument can ever say. That belongs in the doc and it's there.

Also noted and appreciated: you corrected your own PR body about the symlink failures being pre-existing, because your first baseline was taken at the wrong commit. The conclusion held, but you were right that the check hadn't established it.

Yours to merge. On the verdict itself: PROCEED authorises the A/B in the issue and nothing more, and given the sensitivity to all three parameters I'd want the A/B scored on success rate, cost and wall-clock before anything gets built on top of it — which is what your own text says.

PR #231, opened 2026-09-10
docs: add TELEMETRY.md to docs/README.md index

### LINKED ISSUE #228: docs/README.md does not list docs/TELEMETRY.md
`docs/README.md` is the index of every document under `docs/`, one row per file. `docs/TELEMETRY.md` exists on `main` but has no row:

```
$ for f in docs/*.md; do grep -q "$(basename $f)" docs/README.md || echo "missing: $f"; done
missing: docs/README.md      # the index itself, expected
missing: docs/TELEMETRY.md
```

Every other doc is linked (26 rows). A reader who starts at the index cannot find the telemetry contract, which is the one document a privacy-conscious user goes looking for.

**Fix:** add one row to the table in `docs/README.md`, in the same shape as the others, with a one-line "what it covers" taken from the top of `docs/TELEMETRY.md`. Nothing else changes.

**Done when:** the loop above prints only `docs/README.md`, and `uv run pytest tests -k citation -q` stays green.

PrinceXDev:
## What this changes

Adds the missing `docs/TELEMETRY.md` row to the `docs/README.md` index, next to `security.md`, matching the existing table's format and wording conventions. Fixes #228

## Evidence

    $ for f in docs/*.md; do grep -q "$(basename $f)" docs/README.md || echo "missing: $f"; done
    missing: docs/README.md

    $ PYTHONUTF8=1 uv run pytest tests -k citation -q
    81 passed, 3 skipped, 12138 deselected in 25.00s

## Checklist

- [x] Tests cover the change, and they fail on the unfixed code. (the issue's discovery loop is the test; it reported `docs/TELEMETRY.md` as missing before the fix)
- [ ] Lockfiles committed if a dependency changed. (N/A — no dependency change)
      checks for the file.)


eyalgolan (dismissed):
Approving, thanks Prince. TELEMETRY.md landed without getting an index row and you spotted it; the row sits in the right place, reads like the ones around it, and you repinned the manifest in a follow-up commit so `--strict` is clean at 1565 files. I ran the docs gates here too, 151 passed, and `reanchor_citations.py` says VERDICT=OK.

Yours to merge.

eyalgolan:
Verified and queued — thank you, and sorry for the wait.

Checked rather than assumed:

- `docs/TELEMETRY.md` exists on trunk, so the new index row does not point at
  nothing.
- The description is accurate, and pleasingly so — it matches the document's
  own opening sentence, and "machine-checked" is literally true:
  `tests/test_telemetry.py` pins the file to the code's closed allowlist in
  both directions (`test_every_server_event_kind_is_documented` and
  `test_documented_list_has_no_phantom_events`).
- The content diff is exactly one table row; everything else is the generated
  `RELEASE_MANIFEST.txt`.

You spotted a genuine gap: a shipped doc was absent from its own index.

The only thing blocking is the `RELEASE_MANIFEST.txt` conflict, and that is not
yours to fix — the file is generated and gets regenerated at landing, so please
do not rebase on its account. It lands under the usual flow (contributor as
author, identity-normalized committer) once it reaches the front of the queue.

eyalgolan (approved):
Re-approving — my earlier approval got dismissed when you updated the branch, and the row itself hasn't changed a byte since I read it.

I checked the description rather than taking it on trust, because "machine-checked" is the kind of phrase this repo has been burned by before. It holds: `tests/test_telemetry.py` parametrises its two pinning tests over `_DOC_PATHS`, which is `(configuration.md, TELEMETRY.md)`, so both files are compared against the code's `_ALLOWED_EVENTS` in both directions. I renamed an event in the table to be sure the test actually bites:

```
FAILED tests/test_telemetry.py::test_documented_list_has_no_phantom_events[TELEMETRY.md]
1 failed, 40 passed
```

The `[TELEMETRY.md]` in the test id is the part I wanted to see. Restored, back to 41 passed.

The rest, verified on the merge result rather than on your branch tip:

- The manifest re-derives to **exactly one row**, `docs/README.md`, after resetting it to main's and regenerating. `--strict` then gives `OK: 1568 file(s)`.
- With your row in, every `docs/*.md` is reachable from the index — 26 of 26. `TELEMETRY.md` was one of two gaps and the other turned out to be `quickstart.md`, which is already there in bold at the top; my first script just didn't match the bold link. So this closes the last one.
- `tests/test_scope_guard.py`, `tests/test_repo_test_routing_web_desktop.py`, `tests/test_telemetry.py` and `tests/test_readme_claims.py` on the merge result: `217 passed, 12 skipped`. Citation gate and context-file check both `VERDICT=OK`.

One optional thing, and it's my wording you inherited, not yours: the row says "every server-side telemetry event", which is what `TELEMETRY.md`'s own opening sentence says — but the table it points at covers nine events, eight server and one (`screen_viewed`) from the browser. If you want to drop "server-side" so the row describes the whole table, I'd take that; if you'd rather not spend another push on it, that's completely fine and I'll fix the doc's opening line myself.

Last thing, and sorry about it: main moved under you again while this sat — the `NAVIGATION_VALUE.md` row you're now carrying came in with #229. The branch is one commit behind, so you'll need one more update before the merge button lights up. Ping me when you do and I'll re-approve straight away; the manifest churn on this repo is mine to fix, not yours to keep paying for.

Thanks for the clean, well-scoped change.

PR #258, opened 2026-09-10
WINDOWS.md cited an uncommitted plan document, and a second citation was 2,766 lines stale (#110)

### LINKED ISSUE #110: WINDOWS.md cites docs/superpowers/plans/...md, which is not in the repository
`docs/WINDOWS.md:630` cites a plan document that is not in the repository:

```
docs/superpowers/plans/2026-08-26-windows-console-popups-root-cause.md
```

```
$ ls docs/superpowers
ls: cannot access 'docs/superpowers': No such file or directory
```

The whole `docs/superpowers/` directory is absent, and nothing tracked matches that filename.

## Why it is not purely cosmetic

The citation is load-bearing for a step the same section marks as unverified:

> **NOT YET RUN.** This step gates the 0.1.6 Windows walk and has not been executed on a Windows machine. The fix rests on the libuv + Microsoft mechanism (see `docs/superpowers/plans/2026-08-26-windows-console-popups-root-cause.md`) and on the unit tests in `desktop/server.test.mjs` / `tests/test_proc.py`

So the section says, correctly and carefully, that the fix is "mechanism-verified, not confirmed on Windows" — and then points at the mechanism's write-up, which a reader cannot open. The two unit-test files it cites alongside do both exist; only the plan is missing.

I do not know whether the file was internal and never published, or moved, so I have not guessed at a fix. Removing the pointer, replacing it with the upstream libuv/Microsoft reference, or inlining a paragraph of the mechanism would all work, and which one is right depends on what that document was.

## How I found it, and what else it found

I checked every backtick-quoted repo path across `docs/*.md`, `README.md`, `CONTRIBUTING.md`, `CLA.md`, `CHANGELOG.md` and `TRADEMARK.md` against `git ls-files`, resolving repo-relative, `src/no_human/`-relative and basename forms the way `test_readme_claims.py` does.

**344 distinct (document, path) pairs. This is the only genuine miss**, plus `tests/test_test_lanes.py` which is already #109. Everything else that looked missing resolved on inspection: release artefacts that are generated (`SHA256SUMS-*.txt`, `latest*.yml`), user-side config (`config.yaml`, `verifiers.yaml`), electron-builder internals, the evidence-ledger files `pr-body.md` describes (`review.md`, `tamper.md` and friends, which are written to the `nh-evidence/<task-id>` side branch at runtime, not stored here), and files belonging to the fictional example repo in `eval.md`.

I also checked every relative markdown link in those files: **zero dead links**. And a doubled-word scan: three hits, all false positives (`--context context:primary-signature`, `--id ID`, `--path PATH`).

So this is a small finding, and the reason I am filing it rather than sitting on it is that it is the one real result from a fairly thorough sweep — the docs are otherwise accurate.

eyalgolan:
Closes #110, reported by @Sreekant13.

## The defect

`docs/WINDOWS.md:630` pointed at `docs/superpowers/plans/2026-08-26-windows-console-popups-root-cause.md`. That file is not in the repository and never was — a working note that never got committed. The surrounding section is deliberately honest that the fix is "mechanism-verified, not confirmed on Windows", which made that pointer the only thing standing behind the claim, and it resolved to nothing.

## The fix

Point at row 7 of the §2 defect table, which already carries the mechanism and states it more precisely than my own first draft did. Microsoft's ["Process Creation Flags"](https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags) is now linked rather than named, for the one sentence that matters:

> `CREATE_NO_WINDOW` … is ignored if the application is not a console application, or if it is used with either **CREATE_NEW_CONSOLE** or **DETACHED_PROCESS**.

The paragraph now also says which processes the Electron-side fix reaches. `claude.exe` is spawned by the Agent SDK and depends on inheriting nh's hidden console; `git`, `codex` and the test runner do not, because `proc.py:hidden_console_kwargs` passes `CREATE_NO_WINDOW` on each of those spawns directly.

## A second, worse instance in the same file

`docs/WINDOWS.md:351` cited `cli/commands.py:4352` for `signal.SIGKILL`. That line is `console.print(table)`; the real site is `7118` — 2,766 lines out. Same class as #110, in the same file, and live rather than historical. Fixed here.

## What the machine-check does and does not cover

`proc.py:hidden_console_kwargs` is added to `CITATION_TABLE`, so the **code** side is checked. Planting `flags = 0`:

```
FAILED tests/test_readme_claims.py::test_doc_citations_resolve_to_the_code_they_describe[WINDOWS.md:proc.py:hidden_console_kwargs]
FAILED tests/test_readme_claims.py::test_symbol_citation_resilience
2 failed, 133 passed, 12 skipped
```

Restored: `135 passed, 12 skipped`.

The token is `flags = CREATE_NO_WINDOW` rather than the bare constant name, because the bare name also appears in the function's docstring — a row using it stayed **green** under that same mutation.

The **doc** side is *not* checked, and the row's comment says so rather than implying otherwise. `WINDOWS.md` is absent from `_CITATION_DOC_PATHS`, so `_check_citation` never opens the doc: I deleted the citation from `WINDOWS.md` entirely and the suite stayed green at `135 passed`. Registering the doc is a bigger job than this ticket — nine of its ten remaining line citations name bare `.mjs`/`.cjs` basenames under `desktop/`, which `_resolve_source` looks for under `src/no_human` only and does not find. Filed as a follow-up.

## Verification

```
manifest --strict  OK: 1568 file(s) match RELEASE_MANIFEST.txt
citations          VERDICT=OK
context files      VERDICT=OK
pytest             141 passed, 12 skipped
  (test_readme_claims, test_proc, test_doc_anchors)
```

Docs and one test-table row; no source behaviour changes.

## Provenance

The first revision of this PR added a paragraph restating the mechanism from scratch. An independent review caught four unsupported or incorrect assertions in it — including that it claimed the creation *flag* is inherited down the process chain, when what is inherited is the hidden *console*, and the file's own row 7 already said so correctly. That revision has been replaced rather than patched.


eyalgolan:
## FAIL — one blocker, one line. The rest of this PR is right and well-evidenced.

The re-anchor you came here to do is correct, and I verified each half:

- `cli/commands.py:7118` is exactly `sig = signal.SIGKILL if level == _KILL_FORCE else signal.SIGTERM` — the attribute access that raises on Windows. The old `:4352` pointed at unrelated `table.add_column` code, so it had genuinely drifted.
- The old citation `docs/superpowers/plans/2026-08-26-windows-console-popups-root-cause.md` does not exist on trunk. The replacement resolves: `hidden_console_kwargs` at `proc.py:20`, `flags = CREATE_NO_WINDOW` at `:32`.
- The new `CITATION_TABLE` row is **live, not decorative** — mutating `flags = CREATE_NO_WINDOW` to `flags = 0` gives `1 failed, 47 deselected`, id `[WINDOWS.md:proc.py:hidden_console_kwargs]`.
- The comment's self-limiting admissions are accurate: `_CITATION_DOC_PATHS` really is `{security.md, eval.md, KNOWN_ISSUES.md}`, WINDOWS.md really is absent, and deleting the citation sentence really does leave the suite green. The "nine of ten remaining line citations" count reconciles exactly. The Microsoft quote is verbatim.

### The blocker

`docs/WINDOWS.md:629`, added by this PR:

> The mechanism is stated in **row 7 of the defect table in §2**

Measured on your own tree:

```
§2 spans 128-237   ->  0 table lines
§3 spans 238-282   -> 17 table lines        (positive control)
row 7 is at line 250, inside §3
that table's header: | # | Divergence | Why it is forced |
grep -rn 'defect table' across the repo -> exactly one hit: the sentence this PR adds
```

Wrong twice: the section number (§2 → §3), and the table's name — it is the **divergence** table, and "defect table" names nothing anywhere in the repo. A reader following that pointer finds no table in §2 at all.

**This is the same failure class the PR exists to fix.** It replaces a citation that resolved to nothing with a code citation that resolves *plus* a doc cross-reference that doesn't — and nothing catches it, because WINDOWS.md is not in `_CITATION_DOC_PATHS`, as the PR itself discloses. So a false reference ships into a public doc.

**Fix:** `row 7 of the defect table in §2` → `row 7 of the divergence table in §3`. One line, plus re-pinning `docs/WINDOWS.md` in the same commit.

I will re-review the amended commit rather than take the fix on trust — a one-line doc change is exactly the size where "obviously fine" has been wrong repeatedly tonight.

### Nit, not blocking
While you are in there: the sentence would survive a future renumbering better if it named the table by its heading rather than by section number. Not required.

eyalgolan:
## Review of my own branch: the replacement citation was also wrong — fixed in `d7b451d8`

Verified on the merge result against trunk `cfad1a28` (merges clean, 0 conflicts).

**Two of the three claims the new paragraph makes check out exactly:**

```
cli/commands.py:7118  ->  sig = signal.SIGKILL if level == _KILL_FORCE else signal.SIGTERM
proc.py:hidden_console_kwargs  ->  exists; docstring: "``creationflags`` with ``CREATE_NO_WINDOW``"
```

(Positive control for the first: `grep -n SIGKILL` on that file returns 8 lines,
so the instrument does find others — `:7118` is the right one.)

**The third does not.** The new text reads "row 7 of the defect table in §2".
Row 7 is real and its content is exactly as described — but it lives in **§3**,
"Divergences from macOS", whose heading is at line 238 and whose table starts at
line 242. §2 has no table at all:

```
$ sed -n '128,237p' docs/WINDOWS.md | grep -c '^|'
0
$ awk 'NR<=250 && /^## /{h=$0;n=NR} END{print n": "h}' docs/WINDOWS.md
238: ## 3. Divergences from macOS, and why each one exists
```

So this branch fixed a citation that resolved to nothing by shipping another
citation that resolves to nothing — the exact defect #110 reports.

### Why nothing caught it

The `CITATION_TABLE` row added here cannot, and its own comment says so: it
checks the code side, and `WINDOWS.md` is absent from `_CITATION_DOC_PATHS`, so
`_check_citation` never opens the doc. That comment is accurate and I am not
changing it.

### The fix

The wording now names the divergences table in §3, and
`test_windows_md_row_and_section_cross_references_resolve` walks the document's
own structure: for every "row N ... table in §M" it requires the section headed
`## M.` to contain a table row whose first cell is N. It also refuses to pass
vacuously — if the doc contains no such reference at all, the test fails rather
than reporting green over nothing.

Load-bearing, single-change mutation:

```
corrected doc                       1 passed, 147 deselected in 1.91s
restore the wrong "§2" wording      1 failed, 147 deselected in 0.82s
   FAILED tests/test_readme_claims.py::test_windows_md_row_and_section_cross_references_resolve
whole file, corrected               136 passed, 12 skipped in 3.81s
```

`python scripts/check_release_manifest.py --strict` -> `OK: 1568 file(s) match`,
exit 0.

### Still open, and deliberately not done here

Registering `WINDOWS.md` in `_CITATION_DOC_PATHS` remains a larger job than
#110, for the reason the existing comment gives: nine of its ten remaining line
citations name bare `.mjs`/`.cjs` basenames under `desktop/`, which
`_resolve_source` looks for under `src/no_human` only. The new test closes the
section/row class without needing that.

eyalgolan:
Closing as superseded — everything substantive here is already on `main`, landed by another route.

Measured against current `main` (`0e3eefd7`), not against this branch's merge-base:

```
$ git diff origin/main refs/remotes/pr/258 -- docs/WINDOWS.md tests/test_readme_claims.py --stat
 tests/test_readme_claims.py | 5 +++++
```

`docs/WINDOWS.md` has **no remaining diff at all**. The whole delta is a five-line comment in the citation table.

Reading the three-dot diff misled me at first — it showed this PR changing `cli/commands.py:4352` → `:7118`, which looks like live work. That is the change *from this branch's base*; `main` is already at `7118`, and there it resolves:

```
$ sed -n '7118p' src/no_human/cli/commands.py
    sig = signal.SIGKILL if level == _KILL_FORCE else signal.SIGTERM

$ pytest tests/test_readme_claims.py::test_windows_md_code_line_citations_resolve
1 passed in 0.70s
```

The only merge conflict was `RELEASE_MANIFEST.txt`, which is regenerated rather than resolved — so the conflict was never the reason to hold this.

The comment itself is worth keeping in spirit but not in place: it sits above the `security.md` block while describing the WINDOWS.md row, so landing it as-is would put an explanation next to the wrong rows.

PR #262, opened 2026-09-11
Derive attempts.completed_at from the terminal status, not from 43 call sites

### LINKED ISSUE #245: 40 of 43 failure paths never set completed_at, so failure duration is unmeasurable and two metrics are silently biased
## The gap

`attempts.completed_at` is written by only 3 of the 43 `update_attempt(...)`
calls in `core/orchestrator.py` that set a `failure_reason`:

```
update_attempt calls carrying failure_reason : 43
of those, also carrying completed_at         : 3
MISSING completed_at                         : 40
```

The data matches exactly:

```
status        total   completed_at NULL
failed         382          377
interrupted     44           44
succeeded      229           10
```

So "how long did a failed attempt take" is not answerable from the attempts
table. That is the question one asks first when a task has been running for a
day.

## Two consumers are silently biased by it

**1. The health median is success-only.** `core/health.py:85-88`:

```sql
SELECT (julianday(completed_at) - julianday(started_at)) * 86400.0
FROM attempts WHERE completed_at IS NOT NULL AND started_at IS NOT NULL
ORDER BY completed_at DESC LIMIT ?
```

What it actually samples today:

```
failed        7
succeeded   821
```

Failures are the slower and far more numerous class, and they are absent. The
metric reports the duration of work that WORKED and is presented as attempt
duration generally.

**2. The metrics window drops failed attempts that span the cutoff.**
`core/metrics.py:485-489`:

```sql
WHERE julianday(started_at) >= (cutoff)
   OR julianday(completed_at) >= (cutoff)
   OR (completed_at IS NULL AND status = 'in_progress')
```

For a FAILED attempt: clause 2 is NULL (false), clause 3 is false (status is
not `in_progress`). It therefore counts only if it STARTED inside the window.
The function's own docstring says clause 2 exists precisely for "a long attempt
that began earlier" — that case is broken for every failed attempt. The effect
grows as the window shrinks: harmless at 24h, material at 1h, and it
understates cost rather than overstating it.

## Why it is worth fixing rather than working around

The project already treats spend and duration as first-class signals. Both of
the above are wrong in the same direction — they make the system look faster
and cheaper than it is — and neither surfaces as an error. The fix is
mechanical: set `completed_at` on every terminal write, the way the 3 existing
call sites already do.

`interrupted` attempts (44 of 44 NULL) are the same shape and should be
included.

Not claimed here: I have not measured how much spend the windowing actually
drops, because the column needed to measure it is the column that is missing.

Sreekant13:
Closes #245.

Both of your counts verified on the tree before I changed anything, over the AST rather than by grep:

```
update_attempt calls carrying failure_reason : 43
of those, also carrying completed_at         : 3
MISSING completed_at                         : 40
```

## Not fixed at the 40 call sites

You wrote that the fix is mechanical, and it is, but I do not think the mechanical version holds. A call site is a place to forget something, and this invariant had already been forgotten 40 times out of 43. The 41st write would forget it again and nothing would notice, which is exactly how it got here.

`Store.update_attempt` is the one place every one of those writes passes through, and it already carries a backstop of this precise shape for this precise reason: the C2 block stamps a sentinel `failure_reason` when a caller marks an attempt failed without one. New `core/attempt_completion.py` is that backstop's sibling, called from the same function, with the same never-clobber rule.

A **non-terminal** update is left alone. `update_attempt` is also the progress writer (25 of its calls set no status at all), and stamping those would be worse than the gap being closed, because a running attempt would read as finished.

## The interrupted rows were NULL for a different reason

The issue treats them as the same shape. They are not, and it matters, because a chokepoint alone does not fix them: they never reach `update_attempt` at all. **Three** raw-SQL sweeps set the status directly:

| site | sweep |
| --- | --- |
| `core/db.py:2256` | a previous still-open attempt superseded by a new one |
| `core/db.py:2638` | the abandon path |
| `core/db.py:2679` | `close_attempts_of_terminal_tasks` |

That is why all 44 interrupted rows were NULL while the failed rows were merely mostly NULL. All three now stamp it, with `COALESCE` so a row that already carries one keeps it.

## Budget

`core/db.py` is frozen at 5112 with zero headroom and stays at 5112. The `clean` dict comprehension folds from five lines to four to pay for the import, and the three SQL additions fold into the existing string-literal lines. The file's longest line is unchanged at 118.

## Verification

Tests live in `tests/test_db.py` beside the existing attempt coverage rather than in a new file. A terminal update of each of the three statuses stamps it; a progress update does not; a caller-supplied value is never clobbered; and the raw-SQL sweep stamps it on a superseded row.

Both plants fire on exactly the right tests:

```
remove the chokepoint call   -> the three terminal cases fail
remove the SQL clause        -> the sweep case fails
```

Gate: db, metrics, structural-budget, wall-seconds, ledger-window-spend, queue-health, last-activity and attempt-state-distill together give **151 passed, 0 failed**, and 0 failed on unmodified `origin/main` in this same checkout. `reanchor_citations.py` VERDICT=OK, `--strict` clean.

## What this does not do

Existing rows keep their NULLs, so the metrics window at `core/metrics.py:485-489` stays wrong for historical failed attempts until they age out of every window. Going forward it is correct, because clause 2 now has a value to match. A backfill is a data migration and a judgement call about what timestamp to give a row that never recorded one, so I have left it to you rather than picking one.

---

### What this moves that is visible to a user: the drain ETA

Raised in review and worth having in the log, since we squash and this becomes
the commit message.

`_median_attempt_seconds` (`core/health.py:81-89`) selects on
`completed_at IS NOT NULL` **with no status filter**, and its result feeds
`est_drain_seconds` (`core/health.py:171`), which is a user-visible estimate.
Today the population is almost entirely successes, because failed and
interrupted attempts mostly had no `completed_at` to be selected by. **After
this lands they are included, and they become the majority.**

@eyalgolan measured the size of the move on the fleet DB: 838 successes to 7
failures today, a median of `1322.0s`, against `1983.0s` for the failed and
interrupted attempts he could measure, roughly 1.5x. So the drain estimate
should be expected to rise.

**This is correct and the PR does not change it.** They are completed attempts,
and #245 is precisely the complaint that their duration was unmeasurable. An
ETA that silently excluded every attempt that failed was measuring the wrong
population. But whoever notices the estimate move should be able to find out
why from the commit log, which is why it is written down here.

**One nuance I want on the record rather than discovered later.** The three
raw-SQL sweeps stamp `datetime('now')` at SWEEP time, which is when the next
attempt started, not when the dead worker actually stopped. For those rows the
derived duration is therefore the slot occupancy, start until superseded, and
not the time the process ran. For a drain ETA, which is answering how long the
queue takes to empty, occupancy is arguably the more honest input of the two.
It is still not the same quantity as a clean attempt's wall time, and
`ORDER BY completed_at DESC` means these rows sort as recent. Flagging it as a
property of the data rather than proposing a change.


eyalgolan (changes requested):
This is the right shape, and I want to say that first: a chokepoint backstop beside the C2 sibling, not 40 call-site edits. "A call site is a place to forget something, and this invariant had already been forgotten 40 times out of 43" is the correct argument, and it's the reason I'd take this over the mechanical version even though the mechanical version is smaller.

**Everything you claimed, re-derived rather than taken on trust.** Your AST counts, on unmodified main:

```
update_attempt calls : 71
with failure_reason  : 43
...also completed_at : 3     MISSING: 40
```

Exact. The NULL picture on my DB has grown since the issue was written but tells the same story — `2053` of `2060` failed rows and `534` of `534` interrupted. And `core/db.py` is 5112 on main and 5112 on your branch, so the budget claim holds.

**The terminal-status set is complete**, which was my main correctness worry — a missed terminal status would be a silent hole in the very fix. Checked against every row rather than by reading the schema comment:

```
failed 2060 | succeeded 881 | interrupted 534 | in_progress 2
```

Four values ever written, three of them terminal, and `in_progress` correctly excluded. Leaving non-terminal updates alone is right for the reason you give — 25 of the calls set no status and stamping those would make a running attempt read as finished.

Gates on the merge result: `--strict OK: 1569 file(s)`, and `136 passed` across db, metrics, queue-health and structural-budget.

---

## One thing to fix before I approve

**The `COALESCE` in the three sweeps is not pinned.** Your stated mutation does fail, exactly as you say:

```
remove the whole completed_at clause from all 3 sweeps
  -> FAILED tests/test_db.py::test_the_interrupted_sweeps_stamp_completed_at_too
```

But a different one-token mutation does not:

```
keep the stamp, drop COALESCE in all 3  (never-clobber removed)
  -> 54 passed        <- green
```

So the sweeps will stamp, and a test proves it — but nothing proves they *preserve* a timestamp that is already there. That property is load-bearing in your own design: the module docstring says three call sites pass their own `completed_at` and "one of them is the delivery-receipt path whose timestamp is the one that matters". A future edit that simplifies `COALESCE(completed_at, datetime('now'))` to `datetime('now')` would overwrite exactly that, and would silently corrupt the duration data this PR exists to make trustworthy.

The chokepoint's never-clobber *is* pinned — I removed it and `test_a_caller_supplied_completed_at_is_never_clobbered` went red. It's only the raw-SQL half that's uncovered. One test: insert a row with a known `completed_at`, run a sweep, assert it's unchanged. That's the whole ask.

## One thing to add to the body, not the code

`_median_attempt_seconds` (`core/health.py:86-88`) selects on `completed_at IS NOT NULL`, so after this lands it starts including failed and interrupted attempts, and that median feeds `est_drain_seconds` — a user-visible ETA. Today's population is 838 successes to 7 failures; afterwards the failures are the majority, and the seven I can measure run about 1.5x the success median:

```
median today (mostly successes) : 1322.0s
median of failed/interrupted    : 1983.0s
```

I think this is *correct* — they are completed attempts, and #245 is precisely that their duration was unmeasurable — so I'm not asking you to change behaviour. But the body doesn't name `_median_attempt_seconds` or the ETA, and the body becomes the commit message. Whoever notices the drain estimate move should be able to find out why from the log. A couple of sentences alongside your existing note about the metrics window.

Everything else here I'd land as-is. Thanks for the care on the three raw-SQL sweeps in particular — that they bypass `update_attempt` entirely is the part a chokepoint fix would most easily have missed, and you found it and said so.

Sreekant13:
Both addressed.

## The COALESCE is pinned now

You were right that the property was carried by nothing but the reading of the SQL. `test_the_sweeps_never_clobber_a_timestamp_that_is_already_there` sets a known timestamp on a row, triggers the sweep by creating the next attempt, and asserts it is unchanged.

The setup detail worth naming: a row can be `in_progress` and already carry a `completed_at`, because a caller stamps it and the worker dies before the status flips. The sweep matches on status alone, so it reaches exactly that row. That is the shape the delivery-receipt path is exposed to.

I confirmed it fails under your mutation, and on its own assertion rather than incidentally:

```
drop COALESCE in all 3 sweeps     -> FAILED ...never_clobber...
   AssertionError: the sweep overwrote a completed_at that was already set
restored                          -> 55 passed
```

I also pinned `failure_reason` in the same test, since its `COALESCE(NULLIF(TRIM(...), ''), ...)` is the identical one-token mutation away from replacing a caller's real reason with the generic superseded text:

```
drop NULLIF/TRIM in all 3 sweeps  -> FAILED ...never_clobber...
   AssertionError: the sweep replaced a real failure_reason with its generic text
```

Test only. `core/db.py` is untouched and still 5112. Manifest repinned, `--strict OK: 1569 file(s)`, `reanchor_citations.py` VERDICT=OK, `tests/test_db.py` 55 passed.

## The drain ETA is in the body

Added, naming `_median_attempt_seconds` and `est_drain_seconds` by symbol and line, with your 1322.0s against 1983.0s measurement attributed to you since my local DB holds zero attempt rows and I cannot re-derive it. I agree it is correct and have not changed behaviour: an ETA that silently excluded every attempt that failed was measuring the wrong population, which is #245's complaint restated.

**One nuance I added that you did not raise**, because I would rather it be in the log than found later. The three sweeps stamp `datetime('now')` at *sweep* time, which is when the next attempt started, not when the dead worker stopped. For those rows the derived duration is slot occupancy, start until superseded, rather than process wall time. For a drain ETA I think occupancy is arguably the better input of the two, so I am not proposing a change. It is still not the same quantity as a clean attempt's wall time, and `ORDER BY completed_at DESC` sorts those rows as recent. Flagging it as a property of the data.

Thanks for finding the one mutation that stayed green. That is the second time on this PR that the thing I proved was adjacent to the thing that actually needed proving.

eyalgolan (changes requested):
The engineering is done and I verified it — this is one mechanical thing, and it also affects #263, so I've written the detail here and pointed that one back at this comment.

**Both of my asks are properly closed.**

The test pins exactly the property that was uncovered. I ran the mutation that was green last time:

```
drop COALESCE in all 3 sweeps (keep the stamp)
  -> FAILED tests/test_db.py::test_the_sweeps_never_clobber_a_timestamp_that_is_already_there
     1 failed, 54 passed
restored -> 55 passed
```

That is the one I could not make fail before, so the gap is genuinely shut rather than differently phrased. And the new "What this moves that is visible to a user: the drain ETA" section is better than what I asked for — naming `core/health.py:81-89` and `:171` and carrying the actual numbers means whoever sees the estimate move can find out why from the commit message. Manifest re-derives clean, `OK: 1569 file(s)`.

---

## The one thing: `tests/test_db.py` came back as CRLF

Your latest push rewrote the whole file with Windows line endings. It is not a style opinion — it is measurable and it is new:

```
tests/test_db.py at 8b0aab3a (what I reviewed) : CR = 0
tests/test_db.py at 30cb4e62 (now)             : CR = 1029   (every line)
every other file under tests/                  : CR = 0
400 tracked .py/.md/.toml files on main        : CR = 0
```

Two consequences, and the second is the one that actually costs us:

1. **The diff is unreadable.** GitHub and `git diff` both report 983 lines removed and 1029 added, because every line changed. The real change is 47 lines and one new test — I only know that because `git diff --ignore-cr-at-eol` says so. Anyone reading this file's history later gets a wall instead of your test.
2. **Nothing catches it.** There is no `.gitattributes` and no line-ending gate, so CI stays green and it lands silently. That is worse than a red check, because the next file you touch inherits the same setting and the repo drifts a file at a time.

The fix is one command before your next push — `dos2unix tests/test_db.py`, or `sed -i '' 's/\r$//'`, or flipping the editor's EOL setting for this repo. Worth finding whichever tool changed, because it hit both PRs on the same day and will keep doing it.

Nothing else here. Once the endings are back to LF I will re-approve on sight; the substance I have already checked and it is right.

Sreekant13:
Fixed, and thank you for measuring it rather than calling it a style nit. You are right that the silent-landing half is the expensive one.

```
tests/test_db.py pushed now:  bytes=45340  CRLF=0  bare_CR=0
git diff --stat 8b0aab3a..HEAD -- tests/test_db.py
  1 file changed, 46 insertions(+)
```

So the history now shows the 46 lines you were looking for instead of a wall.

**I found the tool, and it was mine, not the editor.** My patch scripts wrote files back with `pathlib.Path.write_text()`, which goes through Python's text layer, and on Windows that translates every `\n` to `\r\n` on the way out. `core.autocrlf` is `false` here, so nothing normalised it on the way into the commit. That is why it hit two PRs on the same day and nothing else: both were touched by the same script, and every file I did not run it over stayed LF.

I checked every tracked text file on both branches rather than only the ones you named. CRLF was in exactly three files: `tests/test_db.py` here, and `verdict_reason.py` plus `test_review_empty_reason.py` on #263. Nothing else on either branch.

**How I know the mechanism rather than guessing at it:** the same script bit me again a few hours later on a different change, and that time it rewrote 64 files at once and produced a 58812 line diff. That is what made the cause obvious. Edits are applied in **bytes** now, and I assert that no file's line endings changed before committing. That check is part of the working method, not a one-off.

One consequence you may want independently of this PR: there is no `.gitattributes` and no line-ending gate, which is exactly why this landed green. I have that written down as a follow-up if you want it, though I would rather not fold it into this PR.

Manifest repinned after the conversion since the hashes moved, `--strict OK: 1569 file(s)`.

eyalgolan (approved):
Approving — endings are back to LF and nothing else moved.

```
tests/test_db.py  CR = 0
content change since 30cb4e62, ignoring CR: RELEASE_MANIFEST.txt only
```

So this is byte-for-byte the version I already verified, just written properly. I re-ran it on the merge result anyway rather than trusting that reasoning:

```
manifest --strict  OK: 1569 file(s)
tests/test_db.py   55 passed
drop COALESCE in all 3 sweeps -> 1 failed  (test_the_sweeps_never_clobber_a_timestamp_that_is_already_there)
```

That mutation is the one that was green before your fix, so the property is genuinely pinned now.

Thanks for turning both of these around quickly, and sorry the line endings cost you a round — I've opened #270 so the next person gets a named failure from CI instead of a reviewer squinting at a diff. Yours to merge.

PR #263, opened 2026-09-11
Give a failed review a reason when the reviewer named nothing

### LINKED ISSUE #249: A review can fail with an empty reason, and once with every checklist item passing
## What happens

An attempt is failed by review, and the reason recorded is literally:

```
failure_reason = "review failed: "
```

Nothing after the prefix. The attempt is spent and the next round is given no
finding to act on.

## Frequency and shapes

Five across the whole history — rare, but the failure mode is bad enough to be
worth closing:

| attempt | checklist items | turns | shape |
| --- | --- | --- | --- |
| `7f579176` a6 | 2, **all passing** | 45 | verdict contradicts its own checklist |
| `73dc6df1` a2 | 0 | 39 | reviewer produced no checklist |
| `e42dfb2f` a1 | 0 | 75 | reviewer produced no checklist |
| `46fe6b92` a1 | 9 | 0 | checklist present, zero turns |
| `f71107e9` a1 | 12 | 0 | checklist present, zero turns |

159 turns total, and five attempt slots. With `max_attempts = 3`, one such
round is a third of a task's allowance spent on a verdict that explains nothing.

## The first row is the sharper defect

`7f579176` a6 recorded `review_passed = 0` while **every** checklist item
passed:

```
[PASS] fix is a no-op on every production call path
[PASS] root-widening into primary checkout relies on the defeatable v1 layer
```

It is the only such case in 529 failed reviews, so this is not the normal path
— but a verdict that contradicts the evidence it publishes is not something a
human or a later agent can reason about, and the next attempt starts with no
signal at all. (Observed live: attempt 7 began immediately, with nothing from
attempt 6 to work from.)

## Suggested direction

Two invariants, both cheap:

1. **A failing review must carry a reason.** If the reviewer returns no
   parseable finding, that is an infrastructure outcome — the existing
   `[session-error]` / "ran but produced no parseable verdict" distinction
   already models this correctly elsewhere and should be reached here rather
   than emitting an empty string.
2. **A verdict must agree with its checklist.** All items passing and
   `review_passed = 0` should be impossible; if the overall verdict comes from
   the model rather than the items, an all-pass checklist with a fail verdict
   should be treated as unparseable (case 1) instead of as a silent failure.

Neither changes what gets failed — only that a failure says why.

Sreekant13:
Closes #249.

One root cause reaches both shapes. `failed` is `blocking_items or failed_items`, and when both are empty the join produces nothing:

```python
detail = "review failed: " + "; ".join(f"{i.label}: {i.evidence}" for i in failed[:3])
```

so the attempt records the literal `"review failed: "` and nothing after it.

## The two shapes are kept apart

They are different facts about what happened, not one bug with two spellings, and the reason says which one occurred:

| shape | attempts | what the reason now says |
| --- | --- | --- |
| reviewer produced NO checklist | `73dc6df1` a2, `e42dfb2f` a1 | no checklist and no finding, treat the verdict as unparseable |
| every item PASSED, verdict failed | `7f579176` a6 | the verdict contradicts the evidence it published |

The second also tells the next round **not to re-implement against it**, because there is no stated defect to implement against. That is what the 45 turns in `7f579176` a6 went into.

## Neither shape takes the session-error marker

Deliberately, and this is the part I most want checked. `reviewer.py` draws the line itself:

> a dead session is infra and parks; a reviewer that ran and produced no parseable verdict, or ran out of turns, still escalates to a person

The reviewer **ran** in both shapes here, so a person is the right destination. What was missing was a sentence for that person to read, not a different route. Taking the marker would route a real verdict into an infrastructure wait with nothing for a human to decide, which is the incident the marker was introduced for.

## Verification

The findings path is asserted **byte-identical** to what it produced before. That path already worked, and this must not quietly rewrite every future failure reason in the fleet.

```
findings present   ->  review failed: a: saw x; b: saw y      (unchanged)
no checklist       ->  review failed: the reviewer produced no checklist and no finding...
all items passed   ->  review failed: every one of the 2 checklist item(s) PASSED and...
```

A bare prefix is unreachable across both shapes and all three checklist sizes exercised (0, 2, 9); those sizes are coverage rather than a census of the historical rows. One AST case pins the **wiring**, because every other test here would still pass if the orchestrator went back to building the string inline, which is where the bug lived. Restoring the inline composition fails exactly that test and nothing else.

## Budget

`core/orchestrator.py` is frozen at 23846 with zero headroom and stays at 23846: the three-line inline composition becomes a one-line call plus a one-line comment, which pays for the import.

## Gate

review-empty-reason, review-fail-closed, structural-budget, post-review-fallback, review-feedback-currency, autonomy and handoff together: **102 passed, 2 failed**, and those 2 are identical **by name** to unmodified `origin/main` in this same checkout (both in `tests/test_handoff.py`, pre-existing on this Windows host). `reanchor_citations.py` VERDICT=OK, `--strict` clean.

This changes what a failure says, never whether it fails.

---

### Correction, 2026-09-11: the count is three, not five

@eyalgolan measured **three** attempts carrying the bare `"review failed: "`,
not the five this description originally claimed, with near-miss variants
explicitly ruled out and the two shapes accounting for all three: two with an
empty checklist and one whose every item passed.

**His number stands and mine does not.** The local DB here now holds zero
attempt rows, so I cannot re-derive where five came from, and his query is
direct and reproducible. Recording that I could not reproduce my own figure
rather than quietly restating his.

The turns figure is unchanged at **159**, which both measurements agree on to
the digit, so the cost argument is untouched: with `max_attempts = 3`, a round
that says nothing is still a third of a task's allowance.

The number also appeared in the shipping module and test docstrings, not only
here, so it was corrected in a commit rather than only in this description.
That commit is docstrings only, verified by comparing the AST of both files
with docstrings stripped before and after: identical, so no behaviour moved.
Manifest repinned, `--strict OK: 1570 file(s)`, `reanchor_citations.py`
VERDICT=OK, 10 passed.


eyalgolan (dismissed):
Approving. Small, surgical, and it fixes the thing it says it fixes.

**The bug is real and I reproduced the count from the fleet DB.** Three attempts carry the literal `"review failed: "` with nothing after it, and they burned exactly the 159 turns you cite:

```
attempts with the BARE 'review failed: ' : 3
turns spent on those                     : 159
```

**Your two shapes are exactly those three rows**, which is a nicer result than the PR claims:

```
failed reviews examined   : 476
...with an EMPTY checklist: 2    <- the "no checklist" shape
...every item passed=True : 1    <- the "contradicts itself" shape
```

So the "one case" you found is confirmed, and 2 + 1 accounts for all three. One number I could not reproduce: you say five attempts landed that way and I measure three, on `failure_reason = 'review failed: '` exactly, with no near-miss variants (I checked for whitespace and length-20 forms and there are none). The turns matching to the digit suggests you counted the same rows a different way. Worth a sentence in the body either way, since we squash and the body becomes the commit message — not a blocker.

**Nothing changes for an ordinary failure.** This was my main worry, since the reason is read by humans and by the next attempt. Driving your function and the old inline expression side by side:

```
SAME  n=1  review failed: a: x
SAME  n=2  review failed: a: x; b: y
SAME  n=5  review failed: l0: e0; l1: e1; l2: e2
VERDICT: identical for every non-empty case
```

Same prefix, same three-item cap, same separator. Taking `failed` as an argument instead of re-deriving it is the right call for exactly the reason you give.

**Both shapes are pinned.** Reverting to the bare join takes five tests red; collapsing the two messages into one takes the contradiction test red on its own:

```
revert to the bare join  -> 5 failed, 5 passed
collapse the two shapes  -> FAILED test_an_all_passing_checklist_names_the_contradiction
restored                 -> 10 passed
```

**The judgement I care most about is the one about `REVIEW_SESSION_ERROR_MARKER`,** and you got it right: the reviewer *ran* in both shapes, so this escalates to a person rather than parking as infra. Staying on the documented side of that line, and saying so in the module, is what makes this safe to land.

Gates on the merge result: `--strict OK: 1570 file(s)`, `123 passed`.

**North star:** with `max_attempts = 3`, a round that says nothing is a third of a task's allowance spent on no signal. This turns that into a sentence the next agent can act on. Good catch and a clean fix.

Approving as of main `2ee83719`; ping me if a landing makes it dirty.

Sreekant13:
Thanks for the approval, and for catching the count.

**You are right and I was wrong: it is three, not five.** I cannot reproduce my own figure. The DB on this machine now holds zero attempt rows, so there is nothing here to re-derive it from, and your query is direct and rules out the near-miss variants explicitly. I have not tried to reconstruct a reading of the data that makes five come out, because that would be working backwards from a number I already know to be unsupported.

What I find reassuring rather than worrying is that **the turns agree to the digit at 159**. Whatever I counted wrong, it was the row count and not the cost, so the argument the PR rests on is untouched: with `max_attempts = 3`, a round that says nothing is still a third of a task's allowance. Your 2 + 1 breakdown accounting for all three is also a tighter result than I claimed, since it means the two shapes are exhaustive rather than merely the ones I happened to find.

**I corrected it in a commit rather than only in the description**, because the number was in the module and test docstrings too, and those ship. The commit is docstrings only. I checked that by parsing both files before and after with docstrings stripped and comparing the ASTs: identical, so nothing executable moved. I also fixed the related claim that the parametrised checklist sizes `(0, 2, 9)` were the sizes the historical rows had; they are coverage, and I should not have described them as a census.

Manifest repinned, `--strict OK: 1570 file(s)`, `reanchor_citations.py` VERDICT=OK, `tests/test_review_empty_reason.py` 10 passed.

Flagging that this pushes a commit onto a branch you approved. It is prose only, but the approval is yours to re-confirm if a docstring change counts as stale for you.

One thing I noticed while running the gates here, unrelated to this PR and not something I have changed: `tests/test_structural_budget.py`'s `scan_tree` calls `path.read_text()` with no `encoding=`, so on Windows it decodes as cp1252 and dies on the first source file containing a UTF-8 multi-byte character. Seventeen tracked files under `src/no_human/` hit it, all of them valid UTF-8, including `core/orchestrator.py`, `api/app.py` and `cli/commands.py`. The whole structural budget gate errors out on this platform. CI is Linux so it never shows there. Happy to open an issue and take it if it is worth having, but I did not want to fold an unrelated fix into this PR.

eyalgolan (changes requested):
Re-reviewing because the push dismissed my approval — and the content change is exactly right, so this is the same single mechanical item as #262.

**The correction is the one I asked for**, and it is the only content change since I approved:

```
git diff --ignore-cr-at-eol 22b6d3c6..HEAD -- src/no_human/review/verdict_reason.py
  added=3 removed=3

-`"review failed: "` and nothing after it. Five attempts across the whole
-history landed that way, spending 159 turns and five attempt slots
+`"review failed: "` and nothing after it. Three attempts across the whole
+history landed that way, spending 159 turns and three attempt slots
```

Thank you for correcting it in the module rather than only in the PR body — the docstring is where the next reader will look, and "three" is what I measure too. The substance still holds on the new head: `10 passed`, reverting to the bare join still takes 5 tests red, manifest `OK: 1570 file(s)`.

**The one thing:** both of your files came back as CRLF on this push, same as `tests/test_db.py` on #262.

```
verdict_reason.py at 22b6d3c6 (what I approved) : CR = 0
verdict_reason.py now                           : CR = 71   (every line)
test_review_empty_reason.py now                 : CR = 114  (every line)
every other file under tests/ and src/          : CR = 0
```

I've written up why it matters on #262 rather than repeating it here — short version: the diff becomes unreadable (a 3-line change renders as a whole-file rewrite), and nothing in CI catches it, so it lands silently and spreads. Same one-command fix, and worth tracking down whatever changed on your side, since it hit both PRs on the same day.

Flip the endings back and I'll re-approve immediately — I've already checked everything else on this head.

Sreekant13:
Fixed here too.

```
src/no_human/review/verdict_reason.py   CRLF 71  -> 0   bare_CR 0
tests/test_review_empty_reason.py       CRLF 114 -> 0   bare_CR 0

git diff --stat 22b6d3c6..HEAD
  src/no_human/review/verdict_reason.py | 4 ++--
  tests/test_review_empty_reason.py     | 9 +++++----
  2 files changed, 7 insertions(+), 6 deletions(-)
```

Seven lines against the commit you approved, which is what the change actually is.

I have put the root cause on #262 to keep it in one place. Short version: it was **my patch script**, not the editor. `pathlib.Path.write_text()` goes through Python's text layer and translates every `\n` to `\r\n` on Windows, and `core.autocrlf` is `false` here so nothing normalised it on the way in. I scanned every tracked text file on both branches, not just the ones you flagged: CRLF was in exactly these two and `tests/test_db.py`, nowhere else.

Edits are applied in bytes now, with a check that no file's line endings moved before anything is committed.

Manifest repinned after the conversion, `--strict OK: 1570 file(s)`.

Thanks for re-reviewing quickly after my push dismissed the approval. That dismissal was the cost of putting the corrected count in the docstring rather than only in the description, and I still think that was the right side of the trade.

eyalgolan (approved):
Approving — same story as #262, endings fixed and nothing else changed.

```
verdict_reason.py            CR = 0
test_review_empty_reason.py  CR = 0
content change since 0397c04d, ignoring CR: RELEASE_MANIFEST.txt only
```

Re-checked on the merge result: `manifest --strict OK: 1570 file(s)`, `10 passed`. The "three attempts" correction in the module docstring is what I asked for and it matches what I measure.

Yours to merge. #270 should stop the CRLF round-trip happening to anyone else.

PR #268, opened 2026-09-11
Every read_text() in the harness states its encoding

### LINKED ISSUE #267: read_text() without encoding= makes the test suite uncollectable on Windows
## Summary

`Path.read_text()` is called without `encoding=` throughout the test suite. On Windows that resolves to the ANSI code page (cp1252 here), so **every test that reads a repo source file dies on the first UTF-8 multi-byte character in it.**

This is not a handful of edge cases. On `main` at `2ee8371`:

- **pytest cannot collect the suite.** `tests/test_knowledge_triggers.py` reads `core/orchestrator.py` at module scope, so it raises during collection, not during a test.
- **44 failures and errors** across 12 test files, all from this one cause.
- **The structural budget gate cannot run at all.** `scan_tree` walks every `*.py` under `src/` with `path.read_text()`, and dies on the first one.

## Reproduction

On Windows, from a clean checkout:

```
python -m pytest tests/ -q --co
```

```
ERROR tests/test_knowledge_triggers.py - UnicodeDecodeError: 'charmap' codec
  can't decode byte 0x9d in position 456288: character maps to <undefined>
12389 tests collected, 1 error
```

Position 456288 is in `src/no_human/core/orchestrator.py`.

## Why CI is green

There **is** a Windows job, `Windows bundle (unsigned)`, and it does run pytest. It runs three files:

```
tests/test_windows_portability.py
tests/test_windows_credential_file.py
tests/test_check_release_manifest.py
```

None of them reads a repo source file, so none of them touches this. **Three files out of 12389 tests.** The Linux jobs default to UTF-8 and pass for free, which is the same trade the manifest CRLF comment in that job already calls out.

## The scale of the exposure

80 tracked text files are valid UTF-8 and undecodable as cp1252:

| area | files |
| --- | --- |
| eval | 25 |
| src | 17 |
| tests | 17 |
| web | 11 |
| docs | 5 |
| README (en, ja, ko, zh-CN) | 4 |
| desktop | 1 |

The 17 under `src/no_human/` include the files the gates read most:

```
core/orchestrator.py   api/app.py        cli/commands.py
cli/tui.py             core/merge_policy.py
core/multi_repo.py     core/pr_evidence.py
```

Every one of them is valid UTF-8. Nothing is wrong with the files.

## What breaks, by gate

These are the gates this repository relies on, so the failure mode is that a Windows contributor cannot run the checks that decide whether their change is allowed to land:

```
tests/test_knowledge_triggers.py        collection error, whole module
tests/test_structural_budget.py         6 errors + 2 failures, gate unusable
tests/test_egress_allowlist.py          8 failures
tests/test_guard.py                     failures
tests/test_db_concurrency.py            2 failures
tests/test_adoption_harness.py          2 failures
tests/test_cli_shell_model.py           1 failure
tests/test_full_report_surface.py       1 failure
tests/test_frozen_snapshot_guard.py     1 failure
tests/test_already_satisfied_subject_tree.py  1 failure
tests/test_web_dist_staleness.py        1 failure
tests/test_doctor_editable_install.py   1 failure
```

## Proposed fix

Pass `encoding="utf-8"` explicitly at the sites that read repository files. That works on every supported interpreter and does not depend on an environment variable being set.

I want to flag a scoping question rather than decide it alone, because the repository has a clear preference for chokepoints over call-site sweeps and this is genuinely a call-site problem:

1. **Narrow.** Fix only the sites that demonstrably break today, roughly 25 calls in 12 files, and prove the 44 go green. Smallest diff, but nothing stops the next one.
2. **Narrow plus a ratcheting guard.** The same fix, plus a budget test in the spirit of `FROZEN_FILE_LINES`: a count of unencoded reads that is only allowed to go down. Catches the next one without a sweep of all 2608 unencoded text I/O calls in the repo, most of which read ASCII fixtures a test just wrote and are harmless.

I lean towards 2 and am happy to build 1 if you would rather keep it minimal.

`requires-python = ">=3.12"`, so PEP 686's UTF-8 default is not available to lean on. Setting `PYTHONUTF8=1` would fix CI but not a contributor running pytest locally, which is the case that matters here.

## Notes

Found while running the gates for #262 and #263 on Windows. Unrelated to either.

Sreekant13:
Closes #267.

`Path.read_text()` with no `encoding=` uses the platform's preferred encoding. On Windows that is the ANSI code page, not UTF-8, so every read of a repository source file dies on its first multi-byte character.

**80 tracked text files here are valid UTF-8 and undecodable as cp1252**, 17 of them under `src/no_human/`, including the three the gates read most: `core/orchestrator.py`, `api/app.py` and `cli/commands.py`. Nothing is wrong with the files.

On Windows, before this:

```
pytest could not COLLECT the suite   test_knowledge_triggers.py reads
                                     orchestrator.py at MODULE scope
44 failures and errors               across 12 test files, one cause
the structural budget gate           could not run at all: scan_tree walks
                                     every *.py under src/ with read_text()
```

CI never saw it. The Linux jobs default to UTF-8 and pass for free, and the `Windows bundle (unsigned)` job runs three test files, none of which reads a repository source file.

## Why a flat zero instead of a narrower rule

You offered me the choice in the issue between fixing only what breaks and fixing it with a guard. I went for the guard, and the reason is that **I could not write a correct narrow rule.**

The first attempt flagged only reads that touch repository files, tracking names rooted at an anchor, at `__file__`, or at a `glob()` walk. It missed, in three different shapes:

- `scan_tree`, because `files = root.rglob(...)` then `for path in files:` is two hops rather than one;
- `_ORCHESTRATOR_SRC`, a module constant whose name was in no anchor list;
- `test_db_concurrency.py`, where the path comes out of a function call, `for path in _package_sources()`, which no static rule is going to follow.

Each repair to the heuristic left a hole somewhere else, and I only found the third one because the tests were still red after I thought I was done. **"Every read says what it is decoding" has no holes, needs no heuristic, and is a rule a reader can apply by eye.** 503 calls in `tests/`, 2 in `scripts/`.

## Scope held

**`write_text` is not swept.** There are roughly 1700 unencoded `write_text` calls in `tests/`, essentially all ASCII literals a test writes and reads back, where the platform default round-trips fine. The one that mattered is fixed: `test_egress_allowlist.py` writes source it had just read, and once the read is correct the write is what raises next.

**`src/` is not covered, and that is a real remaining exposure rather than an oversight:** 57 unencoded `read_text` and 35 unencoded `write_text`. That is the product reading a *user's* files rather than the harness reading its own, so it deserves its own change and its own thought about what should happen when a user's file genuinely is not UTF-8. Happy to take it as a follow-up.

## Verification

**The 12 files that were broken, before and after, same checkout:**

```
UnicodeDecodeError sites   3 -> 0
failures and errors        44 -> 17
```

The 17 that remain are **not** encoding. They are pre-existing Windows failures in the venv install guard, POSIX path assumptions, and they fail identically without this change.

**The guard is pinned against a planted regression, not assumed:**

```
revert scan_tree to an unencoded read
  -> FAILED test_no_read_text_in_the_harness_omits_its_encoding
     naming tests/test_structural_budget.py:1836
restored -> 4 passed
```

It also carries a test that the **scanner itself** can see a known positive, because a guard scoring zero on a broken scanner looks exactly like a guard scoring zero on a clean tree. And one asserting the premise: those three source files are valid UTF-8 and genuinely not cp1252-decodable, so if they ever became pure ASCII a reader can tell that the guard has stopped being load-bearing.

## One thing I got wrong first, worth recording

The first pass produced a **58812 line diff**. Writing through Python's text layer on Windows turned every LF into CRLF across 64 files, which is precisely the regression the manifest comment in the Windows job already warns about. I caught it by reading `git diff --stat` rather than assuming the edit was small.

Redone in **bytes**, because `ast` column offsets are UTF-8 byte offsets rather than character offsets, so a character-based edit would also have corrupted any line with non-ASCII before the call. The final diff is **503 insertions for 503 deletions, exactly one line per call, and no file changed its line endings** (checked explicitly).

Manifest repinned with the new file staged first, `--strict OK: 1569 file(s)`. `reanchor_citations.py` VERDICT=OK.


eyalgolan (changes requested):
The reasoning in this one is the best part of it, and I want to say that before the finding. "I could not write a correct narrow rule, and here are the three shapes that defeated it" is a far more useful thing to read than a clever heuristic would have been, and the `for path in _package_sources()` case is genuinely unfollowable statically. A flat rule that a human can apply by eye is the right call.

**Both of your headline numbers reproduce exactly:**

```
tracked text files          : 1533
valid UTF-8, NOT cp1252     : 80
...of those, under src/     : 17
```

And the change touches `tests/` and `scripts/` only — no production file — which is what makes a sweep this size reviewable at all.

---

## The gap: the suite still cannot collect

I reproduced the failure class on macOS with `LC_ALL=C` (preferred encoding `US-ASCII`), which fails on the same bytes cp1252 does. On main:

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 1904
```

On **your branch**, the identical error, from `src/no_human/testing/pytest_isolated_home.py:88`:

```
ISOLATED_HOME = _activate()                     # line 166, module scope
  _repo_root_or_none()                          # line 126
    text = pyproject.read_text()                # line 88   <- no encoding
```

That runs at import during pytest bootstrap, before any test, so it is still a whole-suite collection blocker on Windows. Adding `encoding="utf-8"` to that one line changes the outcome from "cannot collect" to a suite that collects and runs (`6 failed, 13 passed` under the ASCII locale, the remainder being ordinary locale-sensitive assertions).

So the PR fixes 503 calls and the thing it set out to fix is still broken — by one line.

**Why I think the boundary slipped rather than the judgement.** Your docstring draws the line in the right place: the harness reading *its own* files versus "the PRODUCT reading a user's files ... its own change and its own thought about what should happen when a user's file genuinely is not UTF-8". I agree with that completely. But `GUARDED_AREAS = ("tests", "scripts")` implements it as a *directory* rule, and `src/no_human/testing/` is harness by your definition while living on the product side of that directory line. `pytest_isolated_home.py:88` reads *this repository's* `pyproject.toml` — it literally checks for `name = "no-human"` — so there is no user file and no ambiguity about what encoding it should be.

**What I'd change:** add `src/no_human/testing/pytest_isolated_home.py` to the sweep and to `GUARDED_AREAS`'s reach, and say in the docstring why that one subtree is harness. The rest of `src/` staying out is right and I don't want it widened here.

Two neighbours I checked so you don't have to:

- `src/no_human/testing/structural_budget.py:100,168` — also unencoded, but already safe: `UnicodeDecodeError` is a `ValueError`, and both sites catch `(OSError, SyntaxError, ValueError)` and return `set()`. Worth an `encoding=` for honesty, not for correctness.
- `src/no_human/testing/runner.py:1050` reads the *target* repo's `pyproject.toml`. That one is genuinely product-side and belongs in the separate change you describe. Leave it.

---

## Smaller notes

The guard itself is sound and I confirmed it runs (`4 passed`) and scans by AST rather than by grep. The `write_text` exclusion is argued well, and "once the read is correct the write is what raises next" is the right reason to have fixed the one that mattered.

North star: a contributor on Windows currently cannot run the suite at all, which makes the whole platform undevelopable. That is worth this much churn. I'd just like it to actually reach that outcome before it lands.

Add the one line and I'll re-check the same reproduction and approve — everything else here I'm happy with.

Sreekant13:
You are right, and this is the more useful kind of finding: the PR fixed 503 calls and left the thing it existed to fix broken by one line.

**The diagnosis is exactly right and I want to name it precisely.** My docstring drew the line in the right place, the harness reading its own files versus the product reading a user's. `GUARDED_AREAS = ("tests", "scripts")` drew it somewhere else, as a **directory** rule. `pytest_isolated_home.py` is harness by the definition I wrote and lives on the product side of the folder line, so the code and the prose disagreed and the prose was the correct one.

Fixed, and the guard now reaches `src/no_human/testing/` with the reason written down.

## I checked the whole subtree rather than only the three you named

Your three are fixed:

```
pytest_isolated_home.py:88      blocks collection at bootstrap
structural_budget.py:100,168    already safe (UnicodeDecodeError IS a
                                ValueError and both sites catch it);
                                encoded for honesty, as you said
```

There were **five** more in that subtree, and I agree with your call on all of them, so they are exempted by name in `PRODUCT_SIDE_READERS` rather than swept:

```
runner.py:174,186,1050   take (repo_path / ...) under the TARGET repo
repro_gate.py:122        target repo's MANIFEST; errors="replace" is
                         deliberate and documented there
ui_evidence.py:160       same
```

You had flagged `runner.py:1050`; `:174` and `:186` are the same shape and I would have missed them if I had only patched what you listed. Keyed by file rather than by line so the list does not rot.

## Result

```
collection: 12427 tests, 0 errors    (was 12389 collected + 1 error)
guard:      5 passed
manifest --strict OK: 1569 file(s)   reanchor VERDICT=OK
```

## One honest limit on my verification

**I could not reproduce your failure on Windows**, and I would rather say so than imply I confirmed your repro:

```
pyproject.toml:  utf-8 OK    cp1252 OK    ascii FAILS at byte 1904
```

cp1252 decodes that file where ASCII does not, so on this host that one site was **latent** rather than firing, which is why my own collection run never pointed at it and why I did not find it myself. Your `LC_ALL=C` reproduction is what exposed it. The defect and the fix are identical either way, and the collection-error count going from 1 to 0 is the Windows-visible half of the same thing.

Also: no changed file carries CRLF, checked before committing. Thanks for opening #270 so that stops being something a reviewer has to catch.

eyalgolan (dismissed):
Approving. The blocker is genuinely closed, and you fixed it better than I asked.

**The headline outcome is now real.** Same reproduction as before — `LC_ALL=C`, which gives Python a `US-ASCII` preferred encoding and dies on the same bytes cp1252 does:

```
main   -> UnicodeDecodeError at import: the suite cannot COLLECT
branch -> 6 failed, 13 passed: it collects and runs
```

Under a normal locale the branch is `19 passed`, so those six are my locale stand-in rather than anything you did. I traced one to `src/no_human/core/db.py:903` (`sql_file.read_text()` over a migration) — product-side, inside the 57 reads you deliberately deferred. So the honest summary is **Windows is unblocked, not green**, which is exactly what this PR set out to do and worth saying plainly in that order.

**You moved the boundary rather than patching the line, which is the better fix.** `GUARDED_AREAS` now reaches `src/no_human/testing`, with the genuinely product-side readers exempted by name and each already carrying `errors=`. It bites:

```
re-break pytest_isolated_home.py:88 -> 1 failed, 4 passed
  "...Pass encoding='utf-8': src/no_human/testing/pytest_isolated_home.py:88"
restored -> 5 passed
```

A failure that names the file, the line, the issue and the fix is worth more than the sweep itself.

And the sentence you added to the docstring — *"The line that matters is WHOSE files are being read, not which folder the reader sits in"* — is the part I hope survives. Recording that the first version shipped with the suite still uncollectable, in the file that now prevents it, is how a repo stops relearning things.

Gates on the merge result: `--strict OK: 1572 file(s)` with 142 rows changed, citations `VERDICT=OK`, and the full CI suite is **`12166 passed, 31 skipped, 0 failed`**. For a 141-file mechanical sweep that is the number that matters.

---

**One gap, proven rather than suspected, and I'm not blocking on it.**

The guard matches `read_text()`, so `open(f).read()` walks straight past it. I narrowed the candidates carefully — 17 raw matches, 9 after dropping binary and write modes, and exactly **one** that genuinely reads this repository's own files:

```
tests/test_store_fixture_convergence_guard.py:34 and :47
    for i, line in enumerate(open(f).read().splitlines(), start=1):
        # f iterates glob.glob("tests/**/*.py") — our own sources
```

It is a real failure, not a theoretical one:

```
LC_ALL=C  -> 2 failed :: UnicodeDecodeError at that exact line
normal    -> 2 passed
```

(The other near-misses are clean: `test_ci.py:165` reads a temp file the test itself just wrote, and the rest are `"rb"`, `tarfile.open` or `urllib`'s `opener.open`.)

Your docstring names `write_text` and `src/` as deliberate exclusions and is silent on `open()`, so I think this one is an oversight rather than a decision — and by the rule as you state it, "every read says what it is decoding", it qualifies.

Two lines plus an `open` branch in the AST matcher, if you want to do it while you're here — I'll re-approve on sight. If you'd rather ship this and move on, say so and I'll file it; it's a loud failure rather than a silent one, so it can wait.

Thank you for the care on this one. Sweeps this size are usually unreviewable, and this one was reviewable because you measured the blast radius first and said what you were leaving out.

Sreekant13:
Done while I was here, and you were right that it is an oversight rather than a decision. My docstring named `write_text` and `src/` as exclusions and said nothing about `open()`, and by the rule as I stated it, *every read says what it is decoding*, it plainly qualifies.

## The gap is closed

Three read-mode sites:

```
test_store_fixture_convergence_guard.py:34,47   reads our own sources
test_ci.py:165                                  reads a temp file it wrote
```

I took `test_ci.py` too even though you correctly showed it is safe, because the value of a flat rule is that nobody has to work out which reads are safe.

Only the **builtin** open is matched, by `ast.Name`, so `tarfile.open` and `opener.open` are not reported. Both are pinned as negative cases rather than left to reasoning. Eight parametrised cases cover the branch: default and explicit read modes flagged; `encoding=`, `rb`, `w`, `a` and the two non-builtin opens clean.

Plant on the real sites:

```
revert both open() calls -> FAILED, naming :47 and :34
restored -> 13 passed
```

**Write-mode `open()` stays out, and the docstring now says so explicitly** rather than being silent a second time. Eight calls here plus the ~1700 `write_text` are ASCII literals a test writes and reads back.

## A second Windows bug was hiding under the first one

Worth flagging because it changes what "Windows is unblocked" means. With the decode error gone, `test_store_fixture_convergence_guard.py` **still fails on Windows**, now for an unrelated reason:

```
E  AssertionError: found=['tests\test_db.py', ...]
                expected=['tests/test_db.py', ...]
```

`glob.glob("tests/**/*.py")` returns backslash paths on Windows, while `_PATH_COUPLED_VARIANTS` and the `f == _SELF` self-exclusion both use forward slashes. The sharp detail: **line 25 already normalises `_SELF` with `.replace(os.sep, "/")`**, so the separator problem was known and the fix went on one side of the comparison only.

Both tests fail identically on unmodified `origin/main`, so it is pre-existing and a different bug class. **Not folded in.** One line if you want it as a follow-up, and I am happy to take it.

That also sharpens your framing: **Windows is unblocked, not green**, and the remainder is at least two classes, encoding in `src/` and path separators, rather than one.

On your `db.py:903` trace, agreed, that is inside the 57 `src/` reads I deferred and it belongs in that change rather than this one.

Gates on this head: `--strict OK: 1572 file(s)`, reanchor `VERDICT=OK`, guard `13 passed`, no changed file carries CRLF.

eyalgolan (approved):
Approving. You closed it with the same instrument that found it, which is the right way to end this.

**The three sites are fixed and the guard now sees the shape.** Checked on the merge result:

```
test_store_fixture_convergence_guard.py:34,47  open(f, encoding="utf-8")
test_ci.py:165                                 open(body_path, encoding="utf-8")
guard                                          13 passed
```

And it bites — re-breaking one of them names the exact line:

```
re-break :34 -> 1 failed, 12 passed
                test_store_fixture_convergence_guard.py:34
```

Most importantly, the reproduction I used to report it now comes back clean:

```
LC_ALL=C on test_store_fixture_convergence_guard.py -> 2 passed
```

That file was `2 failed :: UnicodeDecodeError` at that exact line when I raised it, so the finding is closed rather than argued away.

**On the `Path.open()` question, which I went looking for before approving:** your comment explains that you match the builtin `open` by name only, because `tarfile.open` and `opener.open` are Attribute calls that take no `encoding` and would report as false positives. That reasoning is right, and it does leave `p.open()` outside — which shares the same default. So I scanned the guarded areas for text-mode `.open()` calls on anything that isn't a known-safe receiver:

```
TEXT-mode .open() calls that escape the guard: 0
```

I positive-controlled the scanner against a planted file before trusting that zero — it flags `p.open()`, ignores `p.open("rb")` and `p.open(encoding=...)`. So the gap is real in principle and empty in practice, and I'd rather have your false-positive-free rule than a noisier one. Worth one line in the docstring saying `Path.open` is out of scope and why, so the next person doesn't re-derive it.

Manifest `--strict OK: 1572 file(s)`.

**Where this leaves Windows**, for the record rather than for you to act on: unblocked, not green. `src/no_human/core/db.py:903` still reads migrations unencoded, inside the `src/` set you deliberately deferred, and that is the right call — it deserves its own change and its own thought about non-UTF-8 user files.

Thanks for three rounds of this without once arguing with a finding. Yours to merge.

Sreekant13:
Merged at `6891e0e2`, and #267 closed with it.

**On the `Path.open()` docstring line: agreed, and I am deliberately NOT pushing it here.** A one-line comment would dismiss your approval for a third time on this PR and cost you a whole re-review to re-confirm something you have already checked. I will fold it into the `LC_ALL=C` lane PR, which touches the same subject and needs a review round anyway.

Your scan is the better evidence for it than my reasoning was, so I will put your number in the docstring rather than just the argument:

```
TEXT-mode .open() calls that escape the guard: 0
positive-controlled first: flags p.open(), ignores p.open("rb") and p.open(encoding=...)
```

The positive control is the part worth recording. A zero from an unverified scanner is the same shape as the mistake this whole PR was about, and you checked it before trusting it.

**Agreed on where Windows stands: unblocked, not green.** For the record, there are at least two classes left, not one:

- **encoding in `src/`**, including the `db.py:903` migration read you traced, which stays deferred for the reason you give; and
- **path separators**, which the `glob.glob` finding in `test_store_fixture_convergence_guard.py` exposed once the decode error stopped masking it.

Next from me, in the order you asked for: the `LC_ALL=C` lane as its own PR, carrying the `Path.open` note. Then `src/` as a separate change, where I will measure whether `-W error::EncodingWarning` can be scoped as you suggested instead of sweeping the 1700.

Thanks for three rounds on this. Both findings were things I could not have got to on this hardware, and the second one only existed because the first fix uncovered it.

PR #270, opened 2026-09-11
A CRLF file hides its own diff, and nothing here catches it

eyalgolan:
## Why

Two contributor PRs came back this morning with files rewritten in Windows line endings. On #262 a 47-line change rendered as **983 lines removed and 1029 added**, and the one new test inside it was unreadable in review until I passed `--ignore-cr-at-eol` by hand. The same push did it to both files on #263.

Nothing caught it. There is no `.gitattributes` and no gate, so CI was green both times — the only reason it did not land is that I noticed the diff was the wrong shape. That is not a control, and it fails in the direction where the next file inherits the setting and the repo drifts a file at a time.

## What it does

`tests/test_line_endings.py` asserts that no tracked **text** file contains a CR byte, over `git ls-files`, skipping anything containing a NUL — the same binary heuristic `git grep -I` uses, so a font or a PNG full of incidental CRs is not mistaken for text.

Measured before writing the rule, which is why it carries **no exception list**:

```
$ git grep -I -l -P '\r' origin/main
(nothing)
```

The repo is already uniformly LF, so this breaks nothing that exists and needs no grandfathering.

## Verification

Planted the real offending blob rather than a synthetic one:

```
tests/test_db.py from PR #262 head 30cb4e62 (1029 CR bytes)
  -> FAILED test_no_tracked_text_file_uses_crlf
     tests/test_db.py (1029 CR byte(s))
restored -> 2 passed
```

The failure message names the file, the count, and the fix (`dos2unix`), so a contributor can act on it without asking.

The second test drives the predicate on planted bytes directly — including that a NUL-bearing blob is treated as binary — so the rule is still exercised when the tree is clean. A check whose only evidence is a green tree is one nobody has read.

Gates: `--strict OK: 1569 file(s)`, and `98 passed` across line-endings, test-lanes, deselect-lists, structural-budget and classify.

## What this deliberately does not do

It is **not** a `.gitattributes` with `* text=auto eol=lf`. That changes what every existing clone writes to disk on its next checkout — a bigger and much quieter change than a test that names the offending file. If we want normalisation too, it should be its own decision rather than a side effect of this one.

## Checklist

* [x] Tests cover the change, and they fail on the unfixed case (planted the real blob).
* [x] Test count did not go down.
* [x] No dependency change.


Sreekant13:
I caused the regression this gate exists for, so I went looking for ways it could fail rather than just agreeing with it. **I think it will fail closed on a clean checkout for most Windows contributors, which is the population it is meant to protect.**

## The finding

`test_no_tracked_text_file_uses_crlf` reads `path.read_bytes()`, the **working tree**. But `core.autocrlf=true` is the default that Git for Windows' installer recommends ("Checkout Windows-style, commit Unix-style"). Under it, git writes CRLF to disk while the blob stays LF. So the bytes on disk are not the bytes in the repository, and the test is reading the wrong one.

Cloned this repo with that setting and measured:

```
$ git -c core.autocrlf=true clone --branch main . probe
$ cd probe && git ls-files --eol | head -2
i/lf    w/crlf  attr/    .claude-plugin/marketplace.json
i/lf    w/crlf  attr/    .github/CODEOWNERS

index CRLF   : 0
worktree CRLF: 1527   (of 1533 tracked text files)
```

Then ran your test there unmodified:

```
E   AssertionError: these tracked text files contain CR bytes, so their next
    edit will render as a whole-file diff:
E       .claude-plugin/marketplace.json (34 CR byte(s))
E       .github/CODEOWNERS (5 CR byte(s))
E       .github/ISSUE_TEMPLATE/bug_report.yml (53 CR byte(s))
      ... 1527 offenders
```

Nothing is wrong with that clone. Every blob is LF. The contributor has changed nothing, and the advice in the failure message, `dos2unix <file>`, would tell them to "fix" 1527 files that are already correct, which is the one action that would actually introduce the problem.

My own clone has `autocrlf=false`, which is why I did not hit this and why it took a deliberate probe to see.

## The fix, verified in both directions

Ask git what is in the **index** rather than what is on disk. `git ls-files --eol` reports both, and `i/crlf` is the thing you actually care about.

**No false positives** under the Windows default:

```
autocrlf=true clone:  i/crlf count = 0    (w/crlf = 1527)
```

**Still catches the real regression.** A worktree at `30cb4e62`, the #262 head that carried my bad blob:

```
$ git ls-files --eol | grep '^i/crlf'
i/crlf  w/crlf  attr/    tests/test_db.py

index-CRLF files: 1
```

Exactly the one file, which is the same answer your planted-blob test gives, reached without depending on the reader's checkout settings.

Two incidental benefits: it is one subprocess call instead of 1533 file reads, and git does the binary detection itself, so `carries_cr`'s NUL heuristic is no longer load-bearing. I would keep `carries_cr` and its non-vacuity test anyway, since the predicate is still worth exercising.

## What I am not saying

The rule, the empty `CRLF_ALLOWED` measured rather than assumed, the planted real blob, and the decision to keep this separate from a `.gitattributes` change are all right, and the `.gitattributes` reasoning in particular is the part I would have got wrong. This is only about which bytes get inspected.

Happy to push the change to your branch if you want it, or you take it, whichever you prefer. The probe clone is reproducible in two commands if you want to see it fail yourself.

eyalgolan:
Landed on main as f44eb993c9c3e020fa8d866c0b6d406ad6e0fe21 (identity-normalized squash after independent review and merge-result gate).

PR #274, opened 2026-09-11
Catch the encoding class in CI with one job instead of a platform

Sreekant13:
Follow-up to #267 and #268. Your idea, from the issue: one `LC_ALL=C` job catches this class earlier than a Windows runner does, at one job rather than a platform. Kept as its own PR rather than riding along, as you asked.

## Why CI could not see #267

It was invisible in both directions. The linux jobs default to UTF-8 and passed for free. The `Windows bundle (unsigned)` job runs three test files, none of which reads a repository source, so it passed too. The only thing that found the bug was running the suite on Windows by hand.

`LC_ALL=C` gives an ASCII preferred encoding, which fails on a strict **superset** of the bytes cp1252 fails on. It would have caught this on the machine of whoever wrote the line.

## Collection only, which is the load-bearing decision

This runs `pytest --collect-only`, not the suite, for two reasons:

**It is exactly what broke.** `test_knowledge_triggers.py` read `core/orchestrator.py` at module scope, so pytest raised during collection before any test ran. `--collect-only` imports every test module, which is where module-scope reads happen, and stops there.

**It is the part with no locale-sensitive assertions in it.** You measured this yourself on the #268 branch: `6 failed, 13 passed` under `LC_ALL=C` against `19 passed` normally, and you traced the six to locale-dependent behaviour rather than to anything the PR did. A lane that is red for reasons that are not defects is a lane that gets switched off, which is the failure mode the `-n auto` and windows-job comments in this workflow already warn about.

`PYTHONUTF8=0` is pinned deliberately: without it UTF-8 mode can be enabled by the environment and silently defeat the lane, passing for the wrong reason. The step asserts the preferred encoding really is ASCII **before** running anything, so a runner whose locale does not take is a failure rather than a false pass.

## Also carries the `Path.open()` note you asked for

With your measurement rather than just my argument, because the measurement is the better evidence:

```
TEXT-mode .open() calls that escape the guard: 0
positive-controlled first: flags p.open(), ignores p.open("rb") and p.open(encoding=...)
```

The positive control is the part I recorded most carefully. A zero from an unverified scanner is the same shape as the mistake this whole thread was about. The docstring now also says what to do if a `p.open()` ever appears: an `ast.Attribute` case with a receiver check, not a match by name, or `tarfile.open` and `opener.open` come back as false reports.

## Tests

`test_the_ascii_locale_lane_still_guards_collection` pins the job, both env vars and `--collect-only`. A workflow edit that dropped the lane would otherwise be completely silent, and this lane is the only thing standing between this bug class and a green CI.

```
remove the locale job -> FAILED, "the ASCII-locale lane is gone; issue #267
                         could land again with CI green, which is exactly how
                         it landed the first time"
restored -> 14 passed
```

## What I could not verify, and what I would do about it

**I cannot run `LC_ALL=C` on this host.** Windows takes its preferred encoding from the ANSI code page, and patching `locale.getpreferredencoding` at the Python level does not reach the C-implemented `io.open` — I checked that rather than assuming it, and the patch had no effect on either `open()` or `read_text()`.

So **this lane's first run on this PR is its first real execution.** Local collection is clean (`12453 collected, 0 errors`) and the guard is locale-independent AST, but I would rather say plainly that the lane is unproven on my machine than imply I confirmed it. If it goes red on something unrelated to encoding, narrowing it to a slice is the right response and I will take that rather than loosening the check.

Manifest repinned, `--strict OK: 1572 file(s)`, reanchor `VERDICT=OK`, no changed file carries CRLF.


eyalgolan (changes requested):
This is the lane I wanted, built better than I described it — and it has already earned its keep by failing.

**The design decisions are the right ones, and each is defended in the file.** `--collect-only` is the one I'd have argued for: collection is exactly what #267 broke, and it's the part with no locale-sensitive assertions in it. I measured that myself earlier — running test *bodies* under `LC_ALL=C` fails six tests in one file alone for reasons that aren't defects. A lane that is red for the wrong reason gets switched off.

**The self-check is the part I like most.** Asserting the preferred encoding really is ASCII before running anything means the lane cannot pass for the wrong reason — which is precisely how a locale trick fails silently. I ran it both ways:

```
with LC_ALL=C PYTHONUTF8=0 : preferred encoding: US-ASCII   rc=0
without                     : rc=1
```

So it fails closed if the environment ever stops taking. Most people would have written the `pytest` line and stopped.

Manifest is pinned — `--strict` alone, no regeneration, `OK: 1572 file(s)`.

---

## It's red, and it's right to be

```
ASCII locale (collection)   FAILURE
everything else             SUCCESS
```

I reproduced it on the merge result before looking at your CI, and the cause is real:

```
e2e/lane_model.py:53
    text = js_text if js_text is not None else BOARD_LANES_JS.read_text()

  imported at module scope via tests/test_board_e2e_lane_model.py
  -> UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 81
```

**`e2e/` is outside `GUARDED_AREAS`** (`"tests", "scripts", "src/no_human/testing"`), so the AST guard structurally could not see it. Your lane found on its first run a live instance of exactly the class it was built for, in the one directory the static check doesn't cover. That's the best possible argument for the lane existing, and it's also why it can't merge as-is.

**The minimal fix is one line.** I determined that rather than guessing — fixing only `:53`:

```
LC_ALL=C ... pytest -q --collect-only  ->  12453 tests collected, rc=0
```

`lane_model.py:89` is *not* in the collection path (it's called at test time, not import time), so it isn't needed to make the lane green. Add it anyway if you like the symmetry; I'd rather you didn't feel obliged.

**The bigger question, and I think it's yours to answer, not mine to impose:** should `e2e` join `GUARDED_AREAS`? It would flag roughly fourteen more reads across `adoption_run.py`, `personas.py`, `compose.py` and `serve_demo.py`. That's the same judgement you made about `src/` — fix the one that breaks, defer the sweep with the reasoning written down — and I'd support either answer as long as the docstring says which you chose and why.

---

**One limitation worth a line in the comment**, since the comment is otherwise exhaustive: `--collect-only` sees module-scope reads only. A repo-file read inside a test *body* is invisible to this lane and to nothing else, because the guard doesn't reach `e2e/` either. That's a deliberate trade-off rather than an oversight — I just want the next reader to know it's a trade-off.

**And one for me, not you:** this job isn't in the required set on `main` yet (six contexts, none of them this). It'll be advisory until I add it. I'll do that once it's green so we don't wire a red check into the gate.

Fix the one line and I'll approve on sight. Genuinely good work — you went and found a better mechanism than the one I asked for, argued the case with measurements, then built the adoptable one anyway and it caught something the same day.

Sreekant13:
**The lane went red on its first run, on exactly the thing it exists for.** I said above that its first CI execution would be its first real execution, so here is what it found.

```
ERROR tests/test_board_e2e_lane_model.py
  e2e/lane_model.py:53   text = ... BOARD_LANES_JS.read_text()
  UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 81
12439 tests collected, 1 error
```

That read is `web/src/boardLanes.js`, **this repository's own file**, at module scope in `e2e/lane_model.py`, which `tests/` imports. So it was still a whole-suite collection blocker after #268 landed.

**Neither of the other two instruments would have found it.** The guard did not cover `e2e/`, and a Windows run cannot see it: cp1252 decodes `0xe2` where ASCII does not, which is the same reason `pytest_isolated_home.py:88` was latent on the platform the whole sweep was written for.

The preferred-encoding assertion earned its place too. The runner reported `ANSI_X3.4-1968`, so the red was a genuine decode error rather than a locale that failed to take.

## It is the same mistake as last time, one directory further out

You caught me drawing the boundary by **directory** rather than by whose files are read, and we moved `src/no_human/testing/` inside it. `e2e/` is harness by that same definition and was outside for that same reason. Twice now.

So the docstring no longer just records the fix, it records the pattern: enumerate the harness roots here rather than reason about them again each time. `GUARDED_AREAS` gains `e2e`, and the 12 unencoded reads there are fixed across 4 files.

## Now green

```
ASCII locale (collection)   SUCCESS
File inventory              SUCCESS
Web board                   SUCCESS
Wheel carries the board     SUCCESS
Container images            SUCCESS
```

I would rather this landed having caught something than having looked clean. A lane whose entire evidence is a green first run is one nobody has tested, which is the same objection your own non-vacuity tests exist for.

Manifest repinned, `--strict OK: 1572 file(s)`, reanchor `VERDICT=OK`, guard 15 passed, no changed file carries CRLF.

eyalgolan (approved):
Approving. You took the larger of the two options I offered and it was the right one.

**The lane is green and the fix went further than the minimum.** I asked for one line; you fixed the reads in `lane_model.py`, `personas.py`, `compose.py` and `adoption_run.py`, and put `e2e` into `GUARDED_AREAS` so the cheap static check covers the directory the lane just caught. Verified on the merge result:

```
AST guard (with e2e in scope)   15 passed
lane self-check                 preferred encoding: US-ASCII   rc=0
lane collect-only               12454 tests collected          rc=0
manifest --strict (alone)       OK: 1572 file(s)
```

**Both instruments bite on the same planted regression**, which is what I wanted to see before agreeing the belt and braces are worth their cost. Re-breaking `lane_model.py:53`:

```
AST guard   1 failed, 14 passed   -> e2e/lane_model.py:53
lane        12440 collected, 1 error
restored    15 passed
```

The guard names the line in under two seconds; the lane catches it even if the guard's rule is ever narrowed. Different failure modes, same defect — that is the shape I trust.

**Two notes, neither blocking.**

The lane sees module-scope reads only, by construction: a repo-file read inside a test *body* is invisible to `--collect-only`. With `e2e` now guarded, the AST check covers that gap for every directory we care about, which is a better answer than the one I had in mind when I raised it.

And the job is still advisory — `main`'s required contexts are the six that existed before this. I'll add it now that it's green rather than wiring a red check into the gate; that is mine to do, not yours.

**North star:** a contributor on Windows could not run the suite at all, and nothing we had would have told them why. This is one job that catches the class on the machine of whoever wrote the line. Good work, and thank you for arguing the PEP 597 case first — I'd rather have the better mechanism on the record even where we didn't adopt it yet.

PR #273, opened 2026-09-11
docs: make the install snippets release-agnostic and describe KNOWN_ISSUES.md accurately

L4XB:
Closes #241
Closes #242
Closes #243

## Changes

- `docs/INSTALLER.md`: the two `hdiutil attach` snippets no longer name `no_human-0.1.0.dmg`. The verification snippet globs `packaging/dist/no_human-*.dmg` with a one-line note that exactly one DMG is expected there after `npm run dist`; the unsigned-build snippet globs `no_human-*-UNSIGNED.dmg` so it stays distinct. `grep -n "0\.1\.0" docs/INSTALLER.md` prints nothing.
- `docs/LINUX.md`: the user-facing install snippets (`.deb` and AppImage, section 5) use a visible `<version>` placeholder, with a sentence above saying to take the filename from the release page. The dated build-log findings earlier in the file still say 0.1.0, as the issue asks.
- `docs/README.md`: the `KNOWN_ISSUES.md` index row now says the file also keeps closed entries as the record of what the fix proved, which is how the file describes itself (KI-1 and KI-4 are closed). No entry text changed.

## Evidence

```
$ uv run pytest tests -q -k "citation or readme"
167 passed, 13 skipped, 12260 deselected, 2 warnings in 24.18s
```

```
$ grep -n "0\.1\.0" docs/INSTALLER.md
(no output)
```

PR #269, opened 2026-09-11
fix: web board UX polish — pick a rough edge and smooth it

Mr-Neutr0n:
Fixes #22

Hide the Settings "!" badge while the one-time AI configuration popup is showing so first-run presents a single next action instead of two overlapping nudges.

Could not run the suite locally: ule '/tmp/oss-contrib-qce2g20x/no_human/tests/test_abandoned_draft_closed.py'.. Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.

PR #275, opened 2026-09-11
The manifest re-pin is required, but CONTRIBUTING called the section "Optional"

eyalgolan:
## Why

Both first-time contributors today — #269 and #273 — had CI fail on `File inventory` for the same reason: they changed tracked files without re-pinning `RELEASE_MANIFEST.txt`. In both cases the `contributors/<handle>.md` they added **to sign the CLA** was itself one of the unlisted files. Signing the CLA is what broke their build.

Two for two is not a coincidence about those two people.

## What was actually wrong

The rule was already documented, and documented correctly — *"you must re-pin it in the same commit"*. Two things hid it:

- The heading read **`### Optional: the pre-commit manifest gate`**. The *hook* is optional; the *re-pin* is mandatory. Anyone skimming headings sees one word and skips a required step.
- It lives under **Setup**, which someone fixing a typo in a doc has no reason to read. **Before you open a PR** — which they do read — said nothing about it.

## What changed

- The heading now names the requirement: `### Re-pinning RELEASE_MANIFEST.txt — required for every content change`.
- `**Optional:**` moves to the sentence where the optional hook actually starts.
- **Before you open a PR** gains the three commands, the fact that the CLA file counts, and the name of the job that fails without them.

The rule itself is unchanged. No new guard: the gate that catches this already exists and works — it caught both PRs. This closes the reason a newcomer didn't know to run it.

## Verification

```
manifest --strict (alone, no regeneration)  OK: 1572 file(s)
citations                                   VERDICT=OK
pytest test_doc_anchors, test_readme_claims,
       test_test_lanes                      140 passed, 12 skipped
```

I checked for links to the renamed anchor before touching it — there are no `CONTRIBUTING.md#` references anywhere in the tree, and the two `manifest-gate` hits are unrelated code comments.

## Checklist

* [x] Tests cover the change — docs-only; the anchor and claim tests pass.
* [x] Test count did not go down.
* [x] `RELEASE_MANIFEST.txt` re-pinned in the same commit (verified with `--strict` alone, which is the check CI runs).


eyalgolan:
Landed on main as 8fe9ca4c01ae49070ac831e33b69898538bcabdf (identity-normalized squash after independent review and merge-result gate).

PR #59, opened 2026-09-04
One shared exception-safe store fixture in tests/conftest.py; every local aiosqlite store fixture and in-body close converges on it

eyalgolan:
> **Review passed** (1 round) · **Tests passed** (11247 passed, 0 failed) · **Merge policy** ready

## Evidence
| Check | Result |
|---|---|
| Independent review | ✅ **PASSED** — 1 round · [proof](https://github.com/no-human-ai/no_human/blob/8a0e6d143e7565a430a48d321c4195a2fcd896f4/.nh-evidence/7b62b6def13d4046a9a01c650ecbb2d6/review.md) |
| Verifiers | ✅ 1 of 1 satisfied · [proof](https://github.com/no-human-ai/no_human/blob/8a0e6d143e7565a430a48d321c4195a2fcd896f4/.nh-evidence/7b62b6def13d4046a9a01c650ecbb2d6/verifiers.md) |
| Tests | ✅ PASS — 11247 passed, 0 failed, 0 errors · [proof](https://github.com/no-human-ai/no_human/blob/8a0e6d143e7565a430a48d321c4195a2fcd896f4/.nh-evidence/7b62b6def13d4046a9a01c650ecbb2d6/tests.md) |
| Merge policy | ✅ ready — 6 of 6 rules satisfied · [proof](https://github.com/no-human-ai/no_human/blob/8a0e6d143e7565a430a48d321c4195a2fcd896f4/.nh-evidence/7b62b6def13d4046a9a01c650ecbb2d6/merge-policy.md) |

<details><summary>Verifiers (1)</summary>

- ✅ tests-assert-something — 198 files

</details>
<details><summary>Merge-ready policy (6 rules, source: default)</summary>

- ✅ review_passed — review PASSED on head
- ✅ tests_ran_and_passed — tests: 0 failed of 11247 run
- ✅ tamper_guard_clear — tamper guard did not fire
- ✅ repro_gate — repro gate pass
- ✅ verifiers_all_satisfied — 1 verifiers, none failed
- ✅ ci — ci: none reported (tolerated)

</details>

## Acceptance criteria
- (none stated)

#### Final Report

[...]

PR #279, opened 2026-09-11
js-yaml 4.3.1 -> 4.3.2: the desktop app ships the vulnerable copy (GHSA-2883-xcg3-v3hh)

eyalgolan:
Closes Dependabot alert **#59** (high): `maxTotalMergeKeys` does not limit CPU use for empty merge sources, so a hostile YAML document can pin a core.

## Why this one is not just hygiene

`js-yaml` appears four times in `desktop/package-lock.json`. Three of those pulls are build-time — `app-builder-lib`, `builder-util`, `dmg-builder` — and are marked accordingly. The fourth is **`electron-updater`**, which is the desktop app's *only* production dependency:

```json
"dependencies": { "electron-updater": "6.8.9" }
"devDependencies": ["electron", "electron-builder"]
```

Neither `electron-updater` nor `js-yaml` carries `"dev": true` in the lockfile — unlike their neighbours. So **4.3.1 ships inside the packaged app**, parsing update metadata, rather than living only on a maintainer's build machine. That is the difference between "fix at leisure" and "fix now".

## The change

```
npm update js-yaml --package-lock-only --ignore-scripts

-      "version": "4.3.1",
+      "version": "4.3.2",
1 file changed, 4 insertions(+), 3 deletions(-)
```

Exactly one package moves. `^4.1.0` already admitted 4.3.2, so no declared range needed widening and no `overrides` entry was necessary.

## Verification

- `npm ci --ignore-scripts` installs cleanly, and `require('js-yaml/package.json').version` reports **4.3.2 on disk** — not just in the lockfile.
- `npm audit`: **0 vulnerabilities**.
- Desktop suite is **identical on this branch and on unmodified main**: `437 tests, 436 pass, 0 fail`, run twice on each.
- `RELEASE_MANIFEST.txt` re-pinned in the same commit; `--strict` **alone** gives `OK: 1575 file(s)`.

One honest note: my first test run showed 3 failures and a *lower* test count (429 vs 437). That was two `npm test` suites of mine racing each other in adjacent worktrees, not this change — re-running each serially gave identical results on both trees. I mention it because a changed test count is exactly the shape a real regression takes, and it deserved ruling out rather than hand-waving.

## Checklist

* [x] Lockfile committed with the dependency change.
* [x] Test count did not go down (437 on both trees).
* [x] `RELEASE_MANIFEST.txt` re-pinned, verified with `--strict` alone.


eyalgolan:
Landed on main as b50743d6c0ad36b3d92104032e88a86949ddc230 (identity-normalized squash after independent review and merge-result gate).

eyalgolan:
## Landed as `b50743d6` — and recording the advisory here, because the commit message omits it

This landed on main as `b50743d6c0ad36b3d92104032e88a86949ddc230` (identity `eyalgolan`, no trailers, 2 files). Gate on the landing tree: `12329 passed, 222 skipped, 8 warnings in 539.92s`, exit 0; manifest `OK: 1575 file(s) match`.

**My error, recorded where the record lives.** I wrote the squash message about lockfile integrity and supply-chain verification and said nothing about *why the bump exists*. Anyone reading `git log` sees a routine dependency bump. It is a **security fix**, and the advisory is the most useful fact in that commit:

```
GHSA-2883-xcg3-v3hh   severity: HIGH
  js-yaml: maxTotalMergeKeys does not limit CPU use for empty merge sources
  npm/js-yaml   vulnerable >= 4.0.0, < 4.3.2   patched 4.3.2
```

Verified against the GitHub advisory API just now, not taken from the PR title. The desktop app was shipping 4.3.1 — inside the vulnerable range — and `b50743d6` moves it to the first patched version.

Cause of the omission: I reviewed the diff, verified the tarball's own sha512 against the pinned integrity, and structurally diffed the lockfile JSON — then wrote the message from that evidence without reading this PR's own title. Thorough about whether the artifact was authentic, silent about what it was for.

I am not rewriting a pushed commit on a public trunk to fix a message; that is the operator's call. This comment is the durable record in the meantime, and the commit is correct in substance — 4.3.2 is the patched version, its integrity matches the published tarball byte for byte, and the structural diff showed exactly one added key plus three changed values, all `node_modules/js-yaml`.

PR #281, opened 2026-09-11
The CRLF gate reads the working tree, so it fails on a clean Windows checkout

Sreekant13:
Fixes the gate that landed at `f44eb993` (#270). I raised this on that PR before it landed, so this is the fix rather than the report, and it is verified against the commit that is on `main` now.

## The problem

`test_no_tracked_text_file_uses_crlf` reads `path.read_bytes()`, the **working tree**. `core.autocrlf=true` is what the Git for Windows installer recommends ("Checkout Windows-style, commit Unix-style"), and under it git writes CRLF to disk while the blob stays LF. The bytes on disk are not the bytes in the repository.

Cloned `main` at `f44eb993` fresh from GitHub with that setting:

```
index CRLF    :    0
worktree CRLF : 1535   (of 1535 tracked text files)

E  AssertionError: these tracked text files contain CR bytes...
       .claude-plugin/marketplace.json (34 CR byte(s))
       .github/CODEOWNERS (5 CR byte(s))
       ... 1535 offenders
```

Nothing is wrong with that clone. Every blob is LF, the contributor has changed nothing, and the failure message tells them to `dos2unix` 1535 already-correct files, which is the one action that would actually introduce the problem.

**And CI cannot see it.** Linux runners default to `autocrlf=false`, so this is invisible in exactly the place #267 was invisible, and it breaks the platform this repo spent today unblocking.

## The fix

Ask git what is in the **index**: `git ls-files --eol`, and test `i/crlf`.

Verified in both directions against the real historical blob rather than a synthetic one:

```
autocrlf=true clean checkout   -> 10 passed   (1535 w/crlf on disk)
worktree at 30cb4e62           -> FAILED, naming tests/test_db.py
```

`30cb4e62` is the #262 head that carried the original CRLF regression, so the positive case is the incident this gate was written for.

## My first parser was vacuous, and only the positive test caught it

Worth recording, because it is the same class as the bug this file exists to prevent.

`git ls-files --eol` **space-pads** its first three fields and tab-separates only the path:

```
i/crlf  w/crlf  attr/                 <TAB>tests/test_db.py
```

My first version split the whole block on `"/"` and got `lf    w/crlf  attr/` as the value. It matched nothing, parsed all 1570 rows without error, and reported a clean tree on a repository that genuinely had a CRLF blob in the index. **It looked exactly like success**, and the `autocrlf=true` check still passed, so the negative direction alone would have shipped it.

So the parsing is extracted into `parse_eol_line` and driven on planted records including `crlf` and `mixed` — the same non-vacuity argument your `carries_cr` test already makes. A guard that scores zero because its parser is broken is the same shape as the bug.

## What is kept

`carries_cr` and its non-vacuity test stay. Git does the binary classification itself now, but the predicate is the readable statement of what the rule means, and I would rather not delete the thing that says so.

Everything else about #270 is untouched and was right: the empty `CRLF_ALLOWED` measured rather than assumed, the planted real blob, and above all keeping it separate from a `.gitattributes` change, which would silently alter what every existing clone writes on its next checkout.

Incidentally it also replaces 1533 file reads with one subprocess call.

Manifest repinned, `--strict OK: 1576 file(s)`, reanchor `VERDICT=OK`, no changed file carries CRLF.


eyalgolan (approved):
You were right and I was wrong, and the bug you found is worse than "a false positive" — my gate would have told every Windows contributor to corrupt their own checkout. Thank you for fixing it rather than just reporting it.

**I reproduced your finding against the commit that is on `main`, not against your description.** Fresh clone of `f44eb993` with `core.autocrlf`:

```
index    :  1535 i/lf    36 i/-text    6 i/none
worktree :  1535 w/crlf  36 w/-text    6 w/none
```

Then my gate, unmodified, in that clone:

```
FAILED test_no_tracked_text_file_uses_crlf
  .claude-plugin/marketplace.json (34 CR byte(s))
  .github/ISSUE_TEMPLATE/bug_report.yml (53 CR byte(s))
  ... 1537 lines of offenders
```

Every one of those blobs is LF. The remedy my own message prints — `dos2unix` — is the single action that would put real CRLF into the index. I wrote a gate that fails closed on correct work and instructs the contributor to break it, and CI could never see it because Linux runners default to `autocrlf=false`. That is the same blind spot as #267, in the file written to close #267's class.

**Your fix is right, and I checked both directions myself rather than taking the PR body's word:**

```
your version, autocrlf clone            -> 10 passed
your version, real CRLF blob in index   -> FAILED, naming planted_crlf.txt
```

For the positive case I staged a genuine `i/crlf` blob rather than reusing your example, so the two of us have not tested the same fixture. Asking git for `i/` instead of reading bytes off disk is the correct call, and folding 1533 file reads into one subprocess is a pleasant side effect rather than the point.

---

**One gap, and it is in the test you added specifically to prevent this class.**

`test_the_eol_line_parser_reads_the_index_field` cannot catch the bug its own docstring describes. I mutated the real parser line to exactly the mistake you name:

```python
-    token = fields[0].split()[0]          # "i/lf"
+    token = fields[0]                      # the bug you describe
```

Result: **10 passed**. The guard still reports clean, which is precisely the vacuous failure you wrote that test to make impossible.

The reason is the fixtures. Real `git ls-files --eol` **space**-pads:

```
real    : 'i/lf    w/lf    attr/                 \t.github/CODEOWNERS'
fixture : 'i/lf\tw/lf\tattr/\t\tsrc/a.py'
```

With tab-separated fixtures, `fields[0]` is already `"i/lf"`, so `.split()[0]` is a no-op and removing it changes nothing. The parser handles the real format correctly — that is why the gate works on this repo — but the test that guards the parser is driven on a shape git never emits.

Using the real spacing in those rows closes it. That is a one-character-class change per fixture, and then the mutation above goes red.

I am approving on the strength of the fix, because leaving my broken gate on `main` is worse than landing this with a test-fixture gap. Push the fixture change and I will re-approve on sight; if you would rather ship as-is, say so and I will file it and fix it myself — it came out of my file and my mistake.

Gates on the merge result: merges clean, `--strict` alone gives `OK: 1576 file(s)`, `10 passed`.

PR #282, opened 2026-09-12
src/ declares its encodings too, and the guard now covers the repo

Sreekant13:
The deferred half of #267. #268 swept the harness and named `src/` as a real remaining exposure rather than an oversight, because it is the **product reading a user's files** and needed a decision rather than a sweep. This is that decision.

## The decision, and why it is the conservative one

What happens today to a user file that is not valid UTF-8:

```
linux / CI  (utf-8 default)    RAISES UnicodeDecodeError
windows     (cp1252 default)   silently mis-decodes, wrong data, no error
```

So `encoding="utf-8"` **changes nothing on linux or in CI**, where the preferred encoding is already utf-8 and the read already raises. On Windows it replaces a silent mis-decode with the same loud error CI would have given. Every platform ends up doing what the platform we actually test on already did.

**Adding `errors="replace"` would have been a behaviour change on every platform**, suppressing failures that surface today. What should happen to a user's non-UTF-8 file is a product question, and it is not this change's to answer. The eleven sites that already pass `errors=` keep it untouched: this is only about *which codec*, never about whether a failure is swallowed.

## Scope

89 sites across 30 files: 54 `read_text`, 35 `write_text`.

The writes matter for the same reason in the opposite direction. An unencoded `write_text` on Windows encodes with cp1252 and raises `UnicodeEncodeError` on anything outside latin-1, so **today a report containing an em dash cannot be written at all**. I hit exactly that while probing this, which is how it got into scope.

## Verification

**No regressions, measured in the same checkout rather than argued:**

```
clean main   39 failures
with sweep   37 failures
nothing in `after` that is not in `before`
```

**Two tests it fixes**, both Windows decode failures on main:

```
tests/test_docs_gen.py::test_wiki_generator_success
tests/test_docs_gen.py::test_wiki_generator_regeneration_replaces_block
```

**The frozen budgets are safe.** Applied in bytes, 89 insertions for 89 deletions, one line per site, and no file's line count moved. That is load-bearing here: `cli/commands.py` (8666) and `config.py` (3657) are both at zero headroom.

Collection clean at `12552 tests, 0 errors`. Manifest `--strict OK: 1576 file(s)`, reanchor `VERDICT=OK`, no changed file carries CRLF.

## The guard now covers the repository, and the exemption list is deleted

`GUARDED_AREAS` becomes `("tests", "scripts", "src", "e2e")`, and `PRODUCT_SIDE_READERS` is **removed rather than left empty**. It existed to exempt reads of the *target* repository; those sites now declare utf-8 alongside their existing `errors=`, so there is nothing left to reason about before trusting a call.

That boundary was drawn by directory and was wrong **twice, in two different shapes, both found by you after a sweep that claimed to be complete**: `src/no_human/testing/` is harness living on the product side of the folder line, and `e2e/` was outside the list entirely until the ASCII locale lane caught it. Enumerating every shipping root is the answer, because now there is no rule left to misapply.

## What this does not do

`write_text` is still not *pinned* by the guard, though `src/` now declares it. There are ~1700 unencoded `write_text` calls under `tests/` and eight write-mode `open()` calls, essentially all ASCII literals a test writes and reads back where the platform default round-trips fine. Sweeping those would be noise, and you suggested a better route on #267 anyway: `-W error::EncodingWarning` can be scoped with a filter, so the stronger check could start here rather than being all-or-nothing. With `src/` done that is now measurable, and I am happy to take it next if you want it.


eyalgolan (approved):
Approving. The reasoning is the strongest part, and it is the reasoning I'd have wanted rather than the sweep.

**"This changes nothing where we test, and replaces a silent wrong answer with a loud one where we don't"** is exactly the right frame for a 89-site change to shipping code. And declining to add `errors="replace"` was the call — that would have been a real behaviour change on every platform, quietly swallowing failures that surface today, and it is a product question about a user's non-UTF-8 file rather than an encoding cleanup.

**I verified the "no behaviour change" claim structurally rather than by reading the diff.** Pairing every removed line with its replacement and normalising away the kwarg:

```
removed lines: 89   added lines: 89
lines whose only change is NOT the encoding kwarg: 0
```

So there is no site where something else moved under cover of the sweep. And the budget concern you flagged holds — I checked every changed file in `src/` and not one line count moved, which matters because `cli/commands.py` and `config.py` are both at zero headroom.

`errors=` sites are untouched: 79 on main, 79 on the branch. Nothing swallowed, nothing un-swallowed.

**The guard now bites over `src/`.** I re-broke one read and it named the line:

```
re-broke src/no_human/telemetry.py -> 1 failed, 14 passed
  src/no_human/telemetry.py:443
restored -> 15 passed
```

Gates on the merge result: `--strict` alone `OK: 1576 file(s)`, and `35 passed` across the encoding guard, the structural budget and the line-endings gate.

**Deleting `PRODUCT_SIDE_READERS` rather than emptying it is the right instinct.** An exemption list that exists but is empty is an invitation; a rule with nothing left to misapply is a rule. And your summary of why the directory boundary was wrong twice is fair — it was wrong twice, in two shapes, and both times the sweep claimed completeness before the gap was found. Enumerating the shipping roots removes the judgement call that kept failing.

On the `write_text` remainder: I agree with leaving the ~1700 test-side calls alone, and yes — take the `-W error::EncodingWarning` route next if you want it. Scoped by filter it can start at `src/` now that `src/` is clean, which is the measurable version of the argument you made on #267 and a better end state than a sweep.

North star: a contributor on Windows could not run the suite this morning, and a report containing an em dash could not be written at all. This is the half that reaches the product rather than the harness. Yours to merge.

PR #283, opened 2026-09-12
Drive the eol parser test on the shape git actually emits

Sreekant13:
The gap you found on #281. You offered to file it yourself since it came out of your file; I would rather close it, because the vacuous test was mine.

## You were right, and the irony is the point

I wrote `test_the_eol_line_parser_reads_the_index_field` specifically to make a vacuous parser impossible, and it could not catch the bug its own docstring describes. Your mutation, verified here:

```
-    token = fields[0].split()[0]          # "i/lf"
+    token = fields[0]
-> 10 passed
```

The fixtures were the cause. Real `git ls-files --eol` **space**-pads its three eol/attr fields and tabs only before the path:

```
real     i/lf    w/lf    attr/                 <TAB>path
fixture  i/lf<TAB>w/lf<TAB>attr/<TAB><TAB>path
```

With tabs throughout, `fields[0]` was already `"i/lf"`, so `.split()[0]` was a no-op that no mutation of it could kill. The parser handled the real format correctly, which is why the gate worked on this repo, but the test guarding it was driven on a shape git never emits.

## The fix, and the same mutation dying

Fixtures are now built by `_real_eol_record`, which reproduces the padding:

```
AssertionError: At index 1 diff:
  'lf    w/lf    attr/                 ' != 'lf'
```

I also added the `autocrlf=true` row as a case in its own right — LF in the index, CRLF on disk — because reading the wrong half is exactly what reported 1535 offenders on a clean checkout, and it deserves to be a named case rather than an implication.

## The part I think matters more than the fixture

`test_the_fixtures_match_what_git_actually_emits` takes a **real** record from this repository and asserts git still space-pads and tab-separates the way the builder assumes.

Writing fixtures from memory is what caused this, and fixing the literals alone would leave the same failure mode one `git` release away. Now the fixtures are pinned to reality rather than to my description of it: if that format ever changes, the test says so instead of quietly going unkillable again.

Test-only change. Manifest repinned, `--strict OK: 1576 file(s)`, reanchor `VERDICT=OK`, 12 passed, no CRLF.

Thanks for mutating it rather than reading it. Three times now on this thread a test of mine has passed for a reason I did not intend, and every one was found by someone running it rather than reviewing it.


eyalgolan (approved):
Approving. The gap is closed, and I checked it with the same mutation that exposed it rather than by reading the diff.

**Before, on #281's head:**

```
token = fields[0]        (the exact bug the test's docstring describes)
-> 10 passed
```

**Now:**

```
same mutation
-> 6 failed, 6 passed
   test_the_eol_line_parser_reads_the_index_field[i/lf    w/lf    attr/ ... \tsrc/a.py]
   test_the_eol_line_parser_reads_the_index_field[i/crlf  w/crlf  attr/ ... \ttests/b.py]
restored -> 12 passed
```

**And `_real_eol_record` is faithful, which is the part that actually decides whether this works.** A generator that emits a *slightly* wrong shape would have moved the vacuity rather than removed it, so I compared it against git itself:

```
generated: 'i/lf    w/lf    attr/                 \t.github/CODEOWNERS'
real git : 'i/lf    w/lf    attr/                 \t.github/CODEOWNERS'
IDENTICAL: True
```

Byte-for-byte, including the column padding. That is the right way to build a fixture for a parser whose bug *was* the padding.

**The `i/lf` + `w/crlf` row is the best addition here** and it wasn't in my report. It pins the autocrlf shape directly at the parser level — index LF, worktree CRLF, verdict must be `lf` — so the distinction the whole file turns on is now asserted in a unit test rather than only in the integration check. If someone later "simplifies" the parser to read the second field, that row fails on its own.

Gates on the merge result: `--strict` alone `OK: 1576 file(s)`, `12 passed`. And I re-ran it inside the real `autocrlf` clone of `main` — the one where my original gate reported 1535 false offenders — and it is `12 passed` there too.

Thank you for turning this round so fast, and for taking the fixture point seriously rather than treating it as a nit. The gate that was actively harmful this morning is now correct and non-vacuous in both directions, and every step of that came from you.

PR #297, opened 2026-09-12
Desktop shell runs for a PR that touches desktop/, label or not

eyalgolan:
PR #279 changed exactly one file — `desktop/package-lock.json`, a js-yaml security bump that ships inside the packaged app's `app.asar` — and carried no labels. All three packaging jobs gate on the same condition:

```yaml
    if: >-
      github.event_name != 'pull_request'
      || contains(github.event.pull_request.labels.*.name, 'desktop')
```

so all three skipped:

```
Desktop shell                                      skipping
Linux bundle (deb + AppImage) + installed-app run  skipping
Windows bundle (unsigned)                          skipping
```

The desktop signal for that change arrived only *after* it merged, on the push to main — run `34643186981`, where `Desktop shell`, `Windows bundle` and `Linux bundle` are all green on `b50743d6`. None of the three is a required check, so nothing was blocked; the exposure is that a desktop-only PR currently gets no desktop CI while it is reviewable.

## What this does

GitHub has no per-job `paths:` filter, so the new `changed` job asks the files API what the PR touched and publishes one output. `Desktop shell` gains that as a third way in. The label stays as the manual override, and the push-to-main clause is untouched.

## Three things the wiring has to get right

Each has a test that goes red without it — I planted all three and watched them fail:

| Mutation | Test that caught it |
| --- | --- |
| drop the paths clause from `desktop`'s `if` | `test_desktop_job_consumes_the_changed_job` |
| unanchor the match to `desktop/` anywhere | `test_the_script_matches_desktop_paths[files3-false]` |
| gate `changed` on `pull_request` | `test_the_changed_job_is_never_conditional` |

The third is the one worth spelling out: **a skipped dependency skips its dependents exactly like a failed one.** An `if:` on `changed` that excluded pushes would have taken `Desktop shell` off the push to main — deleting the only coverage that exists today while appearing to add coverage. `desktop` also opens with `!cancelled()` so an API hiccup in `changed` can't silence the push run.

Past the files endpoint's 3000-entry cap, "no desktop files" would describe the cap rather than the diff, so that case fails towards running the job.

## How I verified this

The tests drive the job's own shell script over a stubbed `gh` rather than reading it, so the matching logic is executed:

```
$ python -m pytest tests/test_desktop_job_follows_paths.py -q
..........                                                               [100%]
10 passed
```

The negative control — a Python-only diff yielding `desktop=false` — is what makes the positive cases mean anything; without it a script hardcoded to `true` would pass every other case.

Every test in the repo that parses `ci.yml` (9 files), against this branch:

```
$ python -m pytest -n 4 -q tests/test_ci_network_step_bounds.py tests/test_test_lanes.py \
    tests/test_readme_claims.py tests/test_deselect_lists_agree.py tests/test_review_depth_routing.py \
    tests/test_telemetry_ci_environment.py tests/test_text_reads_declare_encoding.py \
    tests/test_release_updater_feed_shipped.py tests/test_ci_upload_assertions_not_line_ending_dependent.py \
    tests/test_desktop_job_follows_paths.py
235 passed, 12 skipped in 27.33s
```

Manifest re-pinned, `--strict` run alone on the committed tree:

```
OK: 1581 file(s) match RELEASE_MANIFEST.txt
```

## What this deliberately does not do

`windows` and `linux_release` keep the label-only gate. They bill at 2x and 45 minutes against `Desktop shell`'s 20, and the exhausted-Actions-minutes reasoning in that comment block is a cost decision — widening it inside a fix for the cheap job would be me making that call quietly. Say the word and it's a two-line follow-up.

## What this PR's own CI proves, and what it can't

This branch touches no `desktop/` file, so `Desktop shell` **should skip here** — that's the negative control running live. The positive control can't come from this PR's diff; I'll push a throwaway commit touching `desktop/`, confirm the job fires, and drop it again before this is reviewable, then post the run link here.


eyalgolan:
Both controls are in, and the probe commit has been dropped — this branch is back to the clean `c397adac`, with `desktop/main.mjs` byte-identical to main.

**Positive control** — throwaway commit `46668943`, which touched `desktop/main.mjs` and carried no label ([run 34700661569](https://github.com/no-human-ai/no_human/actions/runs/34700661569)):

```
success   Which areas a PR touches
success   Desktop shell
skipped   Windows bundle (unsigned)
skipped   Linux bundle (deb + AppImage) + installed-app run
```

Under today's `ci.yml` that same commit would have skipped all three, which is exactly what happened to #279.

**Negative control** — this branch's clean tip, which touches no `desktop/` file ([run 34701096945](https://github.com/no-human-ai/no_human/actions/runs/34701096945)):

```
success   Which areas a PR touches
skipped   Desktop shell
skipped   Windows bundle (unsigned)
skipped   Linux bundle (deb + AppImage) + installed-app run
```

So the job fires on desktop changes and stays quiet otherwise, and the two expensive jobs keep the label-only gate either way. The first attempt at the negative control got cancelled by `cancel-in-progress` when the probe push superseded it; the run linked above is the re-collected one on the clean tip.

eyalgolan:
## Review of my own branch: one defect found and fixed in `394ec78d`

**The `changed` job failed CLOSED on an API error, which skips `Desktop shell`
on a pull request — the exact outcome this PR exists to prevent.**

The script already reasons correctly about the 3000-entry cap:

> so **fail towards RUNNING the job, never towards skipping it**

but did not apply that to the call itself:

```bash
set -euo pipefail
...
files=$(gh api --paginate "repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/files" --jq '.[].filename')
```

A transient failure exits non-zero → `set -e` fails the step → the `changed`
job fails → `needs.changed.outputs.desktop` is empty → on a pull request the
desktop `if` evaluates false → **the job skips**. `!cancelled()` covers the push
to main (there the `event_name != 'pull_request'` clause carries the job by
itself), and `test_a_failing_changed_job_cannot_silence_the_push_run` pins that
half — but the pull-request half had neither a guard nor a test.

Fixed by failing open with the same wording as the cap branch. The new test
drives the path over a `gh` stub that exits non-zero rather than asserting over
the script's text, and I confirmed it is load-bearing with a single-change
mutation:

```
fixed tree                          11 passed in 2.60s
revert ONLY the fail-open hunk      1 failed, 10 passed in 1.59s
   FAILED tests/test_desktop_job_follows_paths.py::test_a_failing_files_api_runs_the_job_rather_than_skipping_it
restored                            11 passed in 1.53s
```

`python scripts/check_release_manifest.py --strict` → `OK: 1581 file(s) match`,
exit 0.

## Left deliberately undone — an operator call, not a defect

This PR fixes **one of the three jobs its own body says skipped.** `windows`
(`ci.yml:701`) and `linux` (`ci.yml:939`) still gate on labels alone, so the
motivating incident is only a third addressed: PR #279's
`desktop/package-lock.json` ships inside `app.asar`, which the Windows and Linux
bundles package too, and a `desktop/` change still gets no signal from either
until after it merges.

Extending `needs: changed` to them is one line each. The reason I have not done
it is billing, not difficulty: it would make every `desktop/`-touching PR pay
for the Windows and Linux bundle jobs as well, and the surrounding comments are
explicit that these jobs are label-gated precisely to avoid ~20 Linux minutes a
run. My recommendation is to extend it — a change that ships in the packaged app
is exactly the change that should pay for packaging signal — but that is a cost
decision and I would rather ask than quietly triple the bill.

eyalgolan:
Pushed `6abdf491`. An adversarial review of the first two commits found **a real bug in the detector** and, worse, a test file that structurally could not see it. Correcting the record here rather than editing the body.

## The detector was wrong above ~2000 filenames

```bash
if printf '%s\n' "$files" | grep -q '^desktop/'
```

`grep -q` exits at the first match, `printf` then dies on SIGPIPE, and `pipefail` makes the pipeline non-zero — so the `if` took the **else** branch. Reproduced locally:

```
n=1000  -> desktop=true
n=2000  -> desktop=FALSE   <- wrong
n=2999  -> desktop=FALSE   <- wrong
```

Nondeterministic in the band below that (same input, different answer between runs), deterministic above it, and **silent** either way: script exits 0, job green, and the log prints `(none)` — an affirmative false statement in the evidence trail.

The comment three lines above it promises *"fail towards RUNNING the job, never towards skipping it"*. It did the exact opposite, in the band immediately below the 3000-file cap that comment is about. And `desktop/` sorts before `docs/`, `src/`, `tests/`, `web/` while the API returns paths in order — so the match is early in a large diff, precisely the case that triggers it. A herestring has no pipe.

## Three more from the same review

- **A file renamed *out* of `desktop/` was invisible.** The API reports only the new path in `filename`; the jq now takes `previous_filename` too.
- **`!cancelled()` → `always()`.** A `needs` job skips its dependents when it fails *and* when it is cancelled. `tests/test_ci_network_step_bounds.py` — in this repo, on this branch — records us measuring GitHub report a timed-out job as `cancelled`, not `failure`. My guard excluded the one terminal state we've actually seen here.
- **A per-step `timeout-minutes` on the `gh api --paginate` call.** The repo has a test that exists to bound every network-touching step; it's scoped to `jobs.linux` and structurally cannot see this job, so the bound is written by hand with a comment saying why.

## The tests could not see any of it

The review mutated the workflow eleven ways. **Nine stayed green** — including two that make the mechanism completely inert:

| mutation | before | now |
| --- | --- | --- |
| delete the `outputs:` block | green | **red** |
| typo `id: areas` | green | **red** |
| drop `previous_filename` | green | **red** |
| restore the SIGPIPE pipe | green | **red** |
| `always()` → `!cancelled()` | n/a | **red** |

Either of the first two leaves `needs.changed.outputs.desktop` permanently empty and silently reverts the gate to label-only — the exact PR #279 outcome — in a file whose docstring claims to cover *"the workflow wiring (which is what silently regresses)"*. The three mutations I planted did go red, but they were a self-selected set that missed the load-bearing wire.

Five tests added, one per previously-green mutation. The endpoint and jq are now pinned by recording the stub's argv, which a stub alone cannot exercise.

```
$ python -m pytest -n 4 -q <all 16 test files referencing ci.yml>
898 passed, 12 skipped
```

(The first body said 9 files. 16 reference `ci.yml`; that was also a review finding.)

## Deliberately not fixed

`on: pull_request` carries no `types:`, so `labeled` does not start a run — meaning the `desktop` label cannot recover a false negative without a push or reopen. Adding `labeled` would fire full CI on **every** label on every PR. That's a billing decision, not a defect to settle inside this change.

eyalgolan:
Landed on main as `e52414a9` by local squash (identity `eyalgolan`, author and committer, no trailers) after an independent fresh-context review PASSed the final commit `0b01f9e7`. The squashed tree changed **zero non-comment lines** versus the reviewed tree — the only edits after the review marked two claims as inference rather than mechanism, since GitHub's docs do not actually say what a job-level timeout does to `cancelled()`, nor that `always()` starts a not-yet-started job on a cancelled run.

Verified on `origin/main`: all three packaging jobs (`desktop`, `windows`, `linux`) carry `needs: changed` and the same `!cancelled() && (...)` gate.

Closing in favour of the squash.

PR #298, opened 2026-09-12
Onboarding registers the user's email; the welcome send sits behind one seam

eyalgolan:
Opening a PR for work that was already pushed and had none — `no-human/19dd94b3-3` is the third round of task `19dd94b3` and was invisible: rounds 1 and 2 had PRs (#293, #295) that were closed with no comment, and this branch never got one.

Merges clean against trunk. 21 files, +1376/-28.

## What is here

- `src/no_human/email/{__init__,base,send}.py` — the four frozen templates and the transport seam
- `src/no_human/api/app.py` +95 — the registration endpoint, including redaction of the address from `GET /api/config`
- `web/src/onboardingEmail.js`, `Onboarding.jsx` +73 — the required step
- **`web/src/replayScrub.js` + tests** — keeps the address out of session-replay network bodies, which `ph-no-capture` cannot do since it only protects the DOM
- `tests/test_onboarding_email.py` +384, four `web/src/*.test.mjs` files, and `test_structural_budget.py` re-frozen

## What is NOT here, stated plainly

`send.py`'s own docstring says delivery is not wired, and that is correct: SES is sandboxed (`ProductionAccessEnabled: false`, production request **DENIED**, case `178688493800757`), verified live today. `SentLast24Hours: 0` against a 200/day quota. Only `getnohuman.com`, `eyal.golan@` and `eyal@` can receive.

So this lands capture + render + a one-module transport seam. Actual delivery to an arbitrary registrant is blocked on transport, tracked separately.

## Review posture

The task reports `31 passed` Python and `90/90` JS with every criterion MET, and its own reviewer caught a leak of the address through `GET /api/config` which this round fixes. **None of that is taken on trust** — this gets an independent review like everything else, with the specific attack being: intercept the transport and search the wire payload for the address, rather than asserting the scrub class is present.


eyalgolan:
Closing — superseded by #303.

Task `19dd94b3` was sent back for the finding on this thread, re-cut, and pushed
attempt 4 to a new branch. This PR's head (`956a96f6`) is the rejected attempt-3
sha, so leaving it open would double-count the task in the review queue.

The finding that closed this round is addressed on #303; review continues there.

PR #299, opened 2026-09-12
Instrument the onboarding funnel so abandonment is visible

eyalgolan:
Opening a PR for pushed work that had none — `no-human/7ee2d939-3`, third round of task `7ee2d939`. It was stranded: the branch existed, no PR, not on main, invisible to review.

14 files, +1009/-28. **Conflicts on `RELEASE_MANIFEST.txt` only** — the mechanical hash-row class, regenerated at landing:

```
$ git merge-tree --write-tree origin/main origin/no-human/7ee2d939-3
```

No source conflict.

## Why it exists

291 external installs produced 6 `task_created` and 1 `task_completed`, and 276 emitted exactly ONE event. Nothing between "the app started" and "a task was created" is recorded, so nobody can say where users leave. Verified still true on trunk: `telemetry.py` contains zero references to `onboarding_step_viewed`, `repo_selected` or `auth_check_succeeded` (positive control: `_ALLOWED_EVENTS` appears 8 times).

## What is here

- `web/src/onboardingFunnel.js` + `onboardingFunnel.test.mjs` (+143)
- `web/src/Onboarding.jsx` +18, `api.js` +7
- `tests/test_telemetry.py` +19, `test_telemetry_environment.py` +4
- `tests/test_structural_budget.py` re-frozen rather than ignored

## The three things its predecessor got wrong

Recorded here because this round must not repeat them, and the review should check each:

1. A live provider call the user did not ask for, which ran **even with telemetry off** — the spend must be gated before the network hop, not inside `record()`.
2. The same request mutated the running worker's exported credential and active auth profile — "board up = worker up", so that silently changed which profile every later attempt billed.
3. `repo_selected` leaked an install's exact repo count by **cardinality** — bucketing the prop does nothing about the event count — while three shipped files and a green test all asserted the leak was prevented.

Also to verify: every new event name must stay out of `_LAMBDA_EVENTS`, since the ingest Lambda 400s a whole batch on one unknown name and the batch then stays queued forever.


eyalgolan:
Closing — superseded by #302.

Task `7ee2d939` was sent back for the finding on this thread, re-cut, and pushed
attempt 4 to a new branch. This PR's head (`8b82c01c`) is the rejected attempt-3
sha, so leaving it open would double-count the task in the review queue.

The finding that closed this round is addressed on #302; review continues there.

PR #301, opened 2026-09-12
The venv guard reads a Windows path as well as a POSIX one (#105)

eyalgolan:
Refs #105, reported by @Sreekant13 — deliberately NOT `Closes`. See the scope note at the bottom.

Both command guards tokenise with POSIX `shlex`, where `\` is an escape character, so every separator in a native Windows path is deleted before resolution is attempted:

```
_lex(r"C:\Users\me\proj\.venv\Scripts\pip.exe install requests")
  -> ['C:Usersmeproj.venvScriptspip.exe', 'install', 'requests']
denial_reason(...)  -> None          # allowed
```

`_resolve_installer` treats an unresolvable installer path as "structure genuinely cannot decide" and allows — right when a path is merely absent, wrong when the path was destroyed before resolution was attempted. The quieter half is worse: after mangling no `/` remains, so the `shutil.which` branch runs, `_is_installer_name` is False, and not even the WARNING fires.

## Which shell — and why this doesn't have to answer it

Whether the coder's Bash tool runs Git Bash (where `\` genuinely is an escape and POSIX lexing is *correct*) or cmd/PowerShell is not established. Guessing "backslash is literal on Windows" would open a fresh bypass — `nh\ approve` is one token under Git Bash and two under a literal reading.

So this doesn't pick. On Windows the guard resolves **every** spelling `win_readings.readings` offers and the first denial wins.

## What this covers

- The direct native path, both separator spellings.
- Nested shells: `bash.exe -c`, `cmd /c`, `cmd.exe /c`, `powershell -Command`, `pwsh -c`, and a fully-spelled runner path. Runner names now match through `_basename` (so `.exe` is stripped) and lowercased (cmd and PowerShell treat `/C` and `-Command` case-insensitively).
- Bare `pip.exe` / `uv.exe` — a real Windows venv's `Scripts/` contains exactly those spellings, and `_resolve_installer`'s no-separator branch was testing the RAW token, returning before `shutil.which` *and* before the WARNING.
- `PIP.EXE` — installer matching is now case-folded on Windows only, where the filesystem is. On POSIX `PIP` is a genuinely different file and folding would be a text match wearing a structural disguise.

## What this does NOT cover — #312

A **space** in the installer path still launders every check, in both readings, because the split is at the space rather than the separator. `_spaced_path_candidates` closes the nested, installer-first shapes and nothing wider:

- a top-level `C:\Program Files\p\.venv\Scripts\pip install x` is untouched;
- `cd X && <spaced>\pip install y`, and the `echo &&` / `timeout` / `env -i` / `VAR=` forms, move the installer off the front and are untouched;
- `_MAX_PREFIX_JOIN` is a bound and therefore a limit — a path with 8+ spaces is not reconstructed, and removing the bound is measurably O(n²).

Three attempts at that class each closed part and claimed all of it, which is the argument for **#312** handling it with a different shape rather than a fourth patch here. The helper's docstring states the residual, and two `strict=True` xfails pin the uncovered shapes so nothing silently closes or widens.

## How I verified this

Every claim above has a mutation behind it, planted and watched:

| mutation | result |
| --- | --- |
| no alternate reading (pre-fix behaviour) | 10 red |
| alternate reading only, original dropped | spaced-path case red |
| alternate reading applied on POSIX too | POSIX-unchanged case red |
| revert `_flatten` to `PurePosixPath(tok).name` | the `.exe` runner rows red |
| drop `cmd`/`powershell`/`pwsh` from the runner set | those rows red |
| revert the second raw-name site | the bare-`pip.exe` rows red |
| drop the Windows case fold | the `PIP.EXE` rows red |

That third one is worth reading: it first came back **green across 325 tests** — nothing proved the platform gate was load-bearing, so the entire `is_windows` parameter could have been deleted unnoticed. `test_a_posix_host_does_not_gain_the_windows_reading` exists because the mutation found that hole, not because I planned it.

```
$ python -m pytest -n 4 -q tests/test_windows_command_readings.py tests/test_guard.py tests/test_venv_install_guard.py
345 passed, 2 xfailed

OK: 1582 file(s) match RELEASE_MANIFEST.txt
```

`agent/guard.py` is untouched at 2892 lines, its frozen budget.

## What this does not establish

The tests flip `_IS_WINDOWS` on a POSIX filesystem. That covers the lexing and resolution, which is where the defect lives. It does **not** cover Windows filesystem semantics — drive letters, case folding, 8.3 names — and the 15 failures in the issue were measured on a real Windows machine that I have not re-run.

## Scope

`Refs`, not `Closes`: #105 also covers `guard.py`'s `_segment_tokens` and the `del /f /s /q` / `rd /s /q` / `Remove-Item -Recurse -Force` gap, and separately `nh.exe approve` / `gh.exe pr merge` / `git.exe push origin main` fail open at DESTRUCTIVE severity (#305, fixed in #307). The spaced-path class is #312. Auto-closing #105 on merge would bury all of that.


eyalgolan:
Full Python suite on this branch, for the record:

```
$ python -m pytest -n 4 -q
12370 passed, 221 skipped, 8 warnings in 505.34s (0:08:25)
```

That's the same tree CI will see, minus the platform matrix — the caveat in the PR body stands: these tests flip `_IS_WINDOWS` on a POSIX filesystem, so none of this is evidence about Windows filesystem semantics.

eyalgolan:
Pushed `e8b3cff7`, which closes a hole an adversarial review found in the first commit. **Two claims in the PR body above were wrong**, and I'd rather correct them here than quietly edit the body.

## Refuted: "the guard resolves every spelling and the first denial wins"

It didn't, for anything wrapped in a Windows nested shell:

```
bash -c     "<primary>\.venv\Scripts\pip install foo"   ->  DENY
bash.exe -c "<primary>\.venv\Scripts\pip install foo"   ->  ALLOW
cmd /c      "<primary>\.venv\Scripts\pip install foo"   ->  ALLOW
powershell -Command "..."                               ->  ALLOW
```

`_flatten` decided "is this a shell runner whose payload I must recurse into" via `PurePosixPath(tok).name` against five bare POSIX names. `bash.exe` is not `bash`, and `cmd`/`powershell`/`pwsh` weren't in the set at all — so the payload stayed one token and **both readings produced the identical un-recursed stream**. The union couldn't help, because neither reading ever saw the install.

Silent again, for the same reason as the original bug: a quoted payload has spaces, so `_basename` yields `pip install foo`, `_is_installer_name` is False, no WARNING. And quoting isn't exotic — it's mandatory the moment the path holds a space, i.e. `C:\Program Files\...`, the exact shape I'd singled out as the one Windows users have.

`_basename` already strips `.exe` (`_is_pkg_manager_name` routes through it for precisely that reason, issue #107), so the fix is that call at the one site left on the raw name, plus the three Windows shell names and their switches. Mutations planted and watched: reverting to `PurePosixPath(tok).name` reddens the `.exe` rows; dropping the Windows names reddens the cmd/powershell/pwsh rows.

## Refuted: "this is confined to `venv_install_guard`"

My scope table probed POSIX binary names. The Windows spelling of a command includes its executable name, and that alone defeats three guards at **DESTRUCTIVE** severity — higher than the HYGIENE-class venv guard this PR fixes:

```
```

`nh.exe approve` and `gh.exe pr merge` are constraint #2 — *the agent never merges* — failing open on a four-character suffix. I've filed that separately rather than widening this PR; it's a different file (`guard.py`, which is line-frozen) and a different severity, and bundling them would make both harder to review. **The scope table above should be read as refuted, not as evidence.**

## Two smaller corrections

- The body says the eight refused commands are each "measured against a verdict this repo has already committed to". True for five. Three of them — the `VIRTUAL_ENV=`, `source ... activate`, and `uv pip install --python` rows — deny on unfixed main too, because the fixture's PATH makes bare `pip`/`uv` resolve into the primary venv. They're real cases but they don't isolate the backslash spelling. The "10 red" mutation figure is unaffected and accurate.
- New residual worth knowing: on Windows the normalised reading turns backslash paths into path-like tokens for the first time, which makes a *pre-existing* over-block newly reachable — `pip install -e . --find-links <home>\Downloads` and `--cache-dir <home>\...` now deny, with a reason that calls them install targets. They aren't; `--find-links` is a wheel source. Same behaviour already exists on POSIX for the forward-slash spelling, so this PR doesn't create it, but it does expose it on the standard offline-install workflow.

Suite after the round-2 fix: `401 passed` across the guard suites, `37 passed` in the new file, `OK: 1582 file(s)` on `--strict` alone.

eyalgolan:
Landed on main as `0e3eefd7` by local squash (identity `eyalgolan`, author and committer, no trailers) after an independent fresh-context review returned **PASS** on `639bfa0c` — 13 of 13 mutations killed, zero survivors, and the vacuity claim held under attack: two mutations that switch the feature off *without* touching any constant kill 34 and 35 tests.

The squashed tree is byte-identical to the reviewed tree; only the commit message changed, to correct my own undercount ("four POSIX shapes" → **eight**, measured by the reviewer through the real `guard.evaluate`) and to record a class this does **not** close.

That class, filed rather than silently widened: `_mutating_subcommand` skips an extra token only for `tok in _VALUE_FLAGS`, so any *other* value-taking flag leaves its value in the subcommand slot. Seven shapes are ALLOW on trunk, on this branch and on the combined tree — `pip --log <f> install`, `--cache-dir`, `--proxy`, `--timeout`, `--retries`, `--cert`, `--exists-action` — while their `--flag=value` twins correctly deny. Same mechanism as the `-C` bug, one flag over.

Verified on `origin/main`: `_POSIX_SCRIPT_FLAGS` present, `win_readings.py` present, manifest `OK: 1586 file(s) match`.

Closing in favour of the squash.

PR #307, opened 2026-09-12
A trailing .exe no longer bypasses the guard's name matching (#305)

### LINKED ISSUE #305: A trailing .exe bypasses the command guard's name matching, including the never-merge and never-push-to-main gates
A trailing `.exe` defeats the command guard's name matching, and three of the affected checks are `GUARD_DESTRUCTIVE` — including the two that enforce "the agent never merges".

Measured on `cfad1a28`, `env={"PATH": ""}`, `never_push_to=["main"]`, with `_IS_WINDOWS` False and True (identical both ways — this is **not** Windows-gated, the suffix is just never stripped):

```
```

Also allowed: `C:\Program Files\GitHub CLI\gh.exe pr merge 7`, `C:\Program Files\Git\cmd\git.exe reset --hard origin/main`.

## Why

`guard.py` resolves argv[0] with `PurePosixPath(argv[0]).name` and compares against bare names (`:287`, `:446`, `:532`, `:1017`, `:1432`, `:1640`, `:1725`). `nh.exe` is not `nh`.

The repo already knows this matters and fixed it in exactly one place: `_is_pkg_manager_name` (`:612`) routes through `venv_install_guard._basename`, which strips `.exe`, with the comment *"so v1 and v2 agree on `pip.exe` (issue #107)"*. That fix was never generalised to the other name-matching sites.

## Severity

`nh approve` and `gh pr merge` are constraint #2 — the agent never merges; merge is a human action. `never_push_to` is the gate on pushing to main. These are `GUARD_DESTRUCTIVE`, a class above the `GUARD_HYGIENE` venv guard that #105 covers.

Mitigating, and the reason I'm filing rather than hot-fixing: on POSIX a coder would have to *write* `nh.exe`, and `nh.exe` doesn't exist there, so the command fails anyway — the guard fails open but the action doesn't happen. On Windows it is both spellable and runnable. I have not measured whether a coder session actually emits `.exe` spellings in practice.

## Shape of the fix, and the constraint on it

One-line-per-site: route every argv[0] name comparison through `_basename` instead of `PurePosixPath(...).name`, the way `:612` already does.

The catch: **`agent/guard.py` is line-frozen at 2892** (`tests/test_structural_budget.py:1737`, a budget that "only ratchets down"), so the edits have to be in-place substitutions rather than added lines, or the budget needs re-anchoring with a justification.

Found while an adversarial review was refuting the scope claim in #301 — that PR asserted these guards already denied both spellings, which is true for the separator spelling and false for the `.exe` one.

eyalgolan:
Closes #305.

```
nh approve 7          ->  DENY      nh.exe approve 7          ->  ALLOW
gh pr merge 7         ->  DENY      gh.exe pr merge 7         ->  ALLOW
git push origin main  ->  DENY      git.exe push origin main  ->  ALLOW
```

`nh approve` and `gh pr merge` are constraint #2 — **the agent never merges** — and `never_push_to` is the gate on pushing to a protected branch. All three are `GUARD_DESTRUCTIVE`, a class above the `GUARD_HYGIENE` venv guard of #105.

Not Windows-gated: the suffix is simply never stripped, so the bypass is spellable anywhere. On POSIX `nh.exe` doesn't exist, so the guard fails open but the action doesn't happen; on Windows it's both spellable and runnable.

## Two defects, not one

**1. Argv paths.** 20 sites extracted a name with `PurePosixPath(x).name`, which keeps `.exe`. They now route through `venv_install_guard._basename` — the helper `_is_pkg_manager_name` already used *"so v1 and v2 agree on `pip.exe` (issue #107)"*. That fix was never carried to its neighbours.

**2. Raw-text fallbacks**, which my first pass missed entirely. After rewriting all 20 sites, `timeout 30 git.exe push origin main` **still allowed**: `timeout` isn't in `_WRAPPERS`, so the verdict comes from `re.search(r"\bgit\s+push\b", cmd)` — which needs whitespace directly after `git` and so never sees `git.exe push`. Four regexes had that shape. I found it by probing a wider corpus instead of assuming the rewrite had closed the class.

`agent/guard.py` is unchanged at **2892 lines**, its frozen budget, so nothing is re-anchored — `_basename(x)` replaces `PurePosixPath(x).name` in place and the regex edits are inline. The now-unused `PurePosixPath` import is dropped, and the one test reaching through it for `guard.PurePosixPath` was realigned to the helper production actually uses: asserting over a second spelling of the name extraction would test something the code doesn't do.

## What the tests prove — and what they don't

I reverted each site individually to see which cases go red:

| | pinned behaviourally | total |
| --- | --- | --- |
| `_basename` sites | 5 | 20 |
| raw-text regexes | 1 | 4 |

The rest are **redundant**, not untested — the guard checks the same command down several paths, so reverting one still denies via another. That's a virtue in a guard and a problem for a test, and separating "redundant" from "reachable by an input I didn't construct" isn't work this change did.

So the **invariant** is pinned instead: no `PurePosixPath(...).name` anywhere in the file, no bare `\bgit\s` matcher. Both go red when violated (planted and watched). They prove *uniformity* — one helper is the single definition of "the executable name of a token" — **not** that every site is individually exercised, and the test file says so rather than implying otherwise.

The 15 command pairs are chosen to reach different clusters: argv[0] approve/merge binaries, argv[0] with a path, wrapper-prefixed forms where `timeout`/`nice` bypass `_WRAPPERS` entirely, wrappers the guard does peel, a nested `sh -c` payload, and the filesystem-scan family. Each bare spelling is a control — a verdict this repo already commits to — so if one stops denying, these tests are measuring nothing.

```
$ python -m pytest -n 4 -q
12374 passed, 227 skipped in 823.08s

OK: 1581 file(s) match RELEASE_MANIFEST.txt
```

## Residual

Backslash-spelled Windows paths (`C:\Program Files\Git\cmd\git.exe push origin main`) are still mangled by POSIX `shlex` before any of this runs — that's #105, handled in #301, and the two are deliberately separate.


eyalgolan:
**Moved to draft. Do not merge this as written — the body above overstates what it does, and the change introduces a regression I did not catch.**

An adversarial review refuted the central claim. I reproduced every finding below myself before accepting it.

## The class is not closed

```
DENY   gh api -X PUT /repos/o/r/pulls/7/merge
ALLOW  gh.exe api -X PUT /repos/o/r/pulls/7/merge      <- merges the PR
DENY   sh -c "gh pr merge 7"
ALLOW  sh -c "gh.exe pr merge 7"
DENY   rm -rf /
ALLOW  rm.exe -rf /
DENY   sudo rm -rf /
ALLOW  sudo rm.exe -rf /
```

`_FORGE_MERGE` (`:1064`), `_FORGE_MENTION` (`:1547`), `_FORGE_WRITE` (`:2057`) and `_RM_RF` (`:146`) are all the same "binary name followed by `\s`" shape and none were widened. **My body says "Four regexes had that shape."** I probed only `git`, then wrote a sentence claiming I'd found the class "by probing a wider corpus instead of assuming the rewrite had closed the class." That sentence describes what I should have done, not what I did. The review counted 128 residual `bare DENY → .exe ALLOW` pairs.

`rm.exe` is the real binary under Git Bash/MSYS2 on Windows — the platform this issue is about.

## It introduces 198 false denials

```
DENY  timeout 30 install -m755 dist/git.exe /usr/local/bin/
      -> "working-tree-unsafe git blocked: … `git /usr/local/bin/`"
DENY  sudo -H sha256sum dist/find.exe
      -> "filesystem-wide scan blocked: …"
DENY  env -i cat build/rg.exe
      -> "filesystem-wide scan blocked: …"
```

All allowed on `main`. `.exe` is now stripped in the argv-recovery scans at `:2211`/`:2531`/`:2541`, which walk **operands** — so a file *argument* named `git.exe` is treated as the executable. Copying a built binary is an ordinary thing to do, and the reason text is false on its face. My `HARMLESS` negative-control list is argv[0]-only and structurally could not see this.

## Both "invariant" tests are decorative

Test 1 greps for the literal `PurePosixPath(`. Replacing all 17 sites with `posixpath.basename(` — already imported at `:46` — leaves both invariant tests **green** while fully reintroducing #305:

```
ALLOW  nh.exe approve 7
ALLOW  gh.exe pr merge 7
ALLOW  find.exe / -delete
```

Test 2 is evaded by `r'\bgit[ \t]+push\b'`, a semantically identical rewrite. I wrote "Both go red when violated (planted and watched)" — they go red only for the one spelling I happened to plant.

## The coverage table is wrong

I claimed 5 of 20 sites pinned and the rest "redundant, not untested". Five more are **reachable and unpinned**, each with an input that flips DENY→ALLOW when reverted: `:1017` (`uv.exe run pip install requests`), `:1810` (`xargs nh.exe approve 7` — constraint #2), `:2235` (`sh -c "git.exe push origin main"`), `:2531` (`git.exe stash`), `:2541` (`timeout 30 git.exe stash`). Also the diff touches **17** call sites, not 20 — three already used `_basename` on `main`.

## Still open, and now stated rather than implied

`.bat`, `.cmd`, `.com`, `.ps1` and a trailing dot all still walk through, as does an uppercase stem (`GH.EXE pr merge 7`). `PATHEXT` covers the first set; npm/scoop shims are `.cmd`/`.ps1`.

## What I'm doing

Reworking it: strip the suffix only where a token is being treated as an executable (not in the operand walks), one shared `_EXE` fragment across every binary-naming matcher, behavioural invariant tests instead of source greps, and either pinning the five reachable sites or dropping the claim. Back for review when it's right rather than when it's green.

eyalgolan:
Closing in favour of #320, which fixes the same issue better.

@Sreekant13's version is stronger on three counts, and they are worth recording rather than just deferring:

**It found the sites by asserting the count.** The issue listed seven; there are ten, and the three extra are the gates themselves. I read the sites by eye, missed four raw-text regexes, and wrote a body claiming I had "probed a wider corpus" when I had probed only `git`. An adversarial review had to tell me — twice.

**Its seam is `is_windows` as a parameter**, following `fs_roots.is_windows_filesystem_root`, so both readings are reachable from a POSIX runner. Mine stripped unconditionally, which is why it also touched operand-walking scans and widened a pre-existing over-block onto the common Windows spelling — the cost I had to file separately as #315. #320 does not have that problem because it only touches argv[0] resolution.

**It says `Refs`, not `Closes`, and pins the residual.** `test_the_raw_text_matchers_are_still_open_and_that_is_recorded` asserts the three raw-text shapes still pass, so whoever widens `_RM_RF` and the lexical matchers sees it go red. My version claimed the class was closed when it was not, three separate times.

What I am carrying forward rather than dropping:

* The raw-text matchers are still open — `rm.exe -rf /`, `sudo rm.exe -rf /`, `echo "gh.exe pr merge 7" | sh`, `timeout 30 git.exe push origin main`. #320 records them; #305 stays open for them.
* **#315** — the operand-walk over-block, which is pre-existing on `main` and unrelated to either PR.
* **#312** — a space in an installer path, the other half of #105.

Nothing here is lost; it moves to a better implementation and two issues that name the remainder honestly.

PR #309, opened 2026-09-12
fix: prevent unsupported high-confidence blocker root causes

### LINKED ISSUE #223: Blocker reports state an invented mechanism as a high-confidence root cause
## What happens

When an attempt escalates with a structured blocker, the report carries a
`root_cause_hypothesis` and a numeric `confidence`. The evidence the agent
gathers for these is often excellent. The **mechanism** it names can be pure
invention, stated at high confidence, and nothing checks it against the code the
agent can read.

## The measured case

An attempt escalated `MISSING_ACCESS` after every `Write` and every mutating
`Bash` call was denied. Its `tried` list is genuinely good work — it isolated
the failure to mutating and multi-operation calls, confirmed inert single
commands still pass, and ruled out a local settings override by checking that
neither `.claude/settings.json` nor `.claude/settings.local.json` exists.

Then it reported:

```json
"confidence": 0.9,
"root_cause_hypothesis": "This coder session was resumed/restarted without the
  bypassPermissions mode that agent/claude_backend.py sets by default, so every
  mutating tool call ... is rejected by the permission layer ... isolating the
  defect to the harness's resume path, not to this task's files or approach."
```

That is false, and falsifiable from the repository the agent had checked out.
`ClaudeBackend._options` builds the options object the same way on every call:

```python
return ClaudeAgentOptions(
    ...
    permission_mode=self.permission_mode,
    resume=resume,
    ...
)
```

`permission_mode` is passed unconditionally; `resume` is a sibling argument, not
a branch. There is no resume path that drops the mode. The actual cause was an
organisation policy disabling `bypassPermissions` (#220).

## Why it costs real money

The hypothesis is the first thing a human reads and the thing the next attempt
is steered by. A confident, wrong mechanism sends the investigation at the
harness instead of the environment. In this case the same wrong direction was
re-asserted across several attempts.

This is the same class the project already recognises elsewhere — evidence
handed to a reader must carry its provenance, and a claim must not outrun what
was established.

## Suggested direction

The blocker schema already separates `tried` (observations) from
`root_cause_hypothesis` (an inference). Make that separation mean something:
require a hypothesis that names a code mechanism to cite the file and symbol it
rests on, and treat an uncited mechanism as an observation, not a cause. A high
`confidence` on an uncited mechanism is the specific thing to refuse.

AyushCipher:
## Problem
Agent-authored blocker reports could claim a specific internal code mechanism with high confidence even when the mechanism was not supported by the captured evidence. When agents encountered failures (such as `MISSING_ACCESS` / 403), they could hallucinate internal class methods, attributes, or flags in `root_cause_hypothesis` (e.g., `ClaudeBackend._options is missing permission_mode setting during resume`) and assign high confidence (>= 0.60, e.g., 0.95), misleading operators into debugging fictitious internal code defects.

## Root Cause
`parse_blocker()` trusted the model-provided confidence without validating whether concrete code-mechanism claims (such as dotted symbol paths, source file references, explicit function calls, backticked code tokens, PascalCase class names, and snake_case identifiers) were grounded in the captured `evidence`, `tried` attempts, or task `goal`. Furthermore, an incidental match on an unrelated token could previously mask an unsupported mechanism.

## Fix
- Added `extract_code_mechanisms(text)` to extract code-mechanism references (dotted symbols, source file paths, backticked expressions, explicit function calls, camelCase/PascalCase/snake_case identifiers).
- Excluded general prose terms, protocol terms, and `no_human` taxonomy vocabulary (derived from `BlockerCategory`, `Blocker`, `BlockerOption`, and `TaskStatus`) to avoid false-positive detections on terms like `MISSING_ACCESS` or `rate_limit`.
- Added `is_code_mechanism_supported(mechanisms, evidence, tried, goal)` requiring that all asserted code mechanisms are grounded in the captured corpus (including dotted-part decomposition and file basename matching).
- Integrated grounding validation into `parse_blocker()` at the trust boundary: if an agent-authored blocker asserts code mechanisms with confidence >= 0.60 that are not fully supported by evidence, tried, or goal, its confidence is demoted to `0.50` (below the 0.60 escalation threshold).
- Exported helper functions in `src/no_human/blockers/__init__.py`.
- Added CLA agreement for author at `contributors/ayushcipher.md`.
- Re-pinned `RELEASE_MANIFEST.txt` matching all 1,580 files.

## Tests
- Added 8 regression and unit tests in `tests/test_blockers.py`:
  - `test_unsupported_code_mechanism_hypothesis_confidence_demoted_to_low`: reproduces Issue #223, verifying unsupported `ClaudeBackend._options` with 0.95 confidence is demoted to 0.50.
  - `test_incidental_token_does_not_whitelist_unsupported_mechanism`: verifies that having an incidental supported token (`auth_mode`) alongside unsupported mechanisms (`ClaudeBackend._options`, `permission_mode`) still demotes confidence to 0.50.
  - `test_supported_code_mechanism_hypothesis_retains_high_confidence`: verifies supported mechanisms retain high confidence.
  - `test_taxonomy_vocabulary_does_not_trigger_false_mechanism_detection`: verifies taxonomy words (`MISSING_ACCESS`, `rate_limit`, `transient_infra`) do not trigger false mechanism checks.
  - `test_non_code_mechanism_hypothesis_retains_confidence`: verifies plain-English hypotheses without code mechanisms retain high confidence.
  - `test_unsupported_code_mechanism_with_empty_evidence_demoted`: verifies missing evidence triggers demotion.
  - `test_already_low_confidence_unsupported_mechanism_not_increased`: verifies hypotheses already below the threshold remain unchanged.
  - `test_extract_code_mechanisms_and_support_helpers`: comprehensive unit tests for extractor exclusions, multi-part grounding, and all-supported invariant verification.

### Verified Results:
- 111 blocker tests passed (`pytest tests/test_blockers.py`)
- 140 relevant blocker tests passed (`pytest tests/test_blockers.py tests/test_blocker_challenge.py tests/test_blocker_answer_reuse.py tests/test_api_legacy_blocker_list_fields.py`)
- 15 encoding tests passed (`pytest tests/test_text_reads_declare_encoding.py`)
- Release manifest check passed for 1580 files (`scripts/check_release_manifest.py --strict`)
- Context file check passed (`scripts/check_context_files.py --claude CLAUDE.md`)

Fixes #223


eyalgolan (changes requested):
Thanks for this, @AyushCipher — and for shipping it with tests, which is not the default and makes a real difference to how fast I can review. #223 is my own issue and the framing you've picked (validate at the trust boundary in `parse_blocker`, don't trust self-reported confidence) is the right one.

I'm asking for changes on two things, plus one mechanical.

## 1. One grounded token whitelists a fictitious one

`is_code_mechanism_supported` returns `True` as soon as **any** extracted mechanism appears in the corpus. Run through `parse_blocker` itself:

```
#223's hypothesis, bare                              -> conf 0.50   (your fix working)
SAME claim + " ; auth_mode handling" appended        -> conf 0.95   (not demoted)
   evidence: "auth_mode=subscription\nError: 403 Forbidden ..."
```

The hypothesis still asserts `ClaudeBackend._options is missing permission_mode during resume`. It keeps 0.95 because one incidental token happened to be grounded. The same model writes the hypothesis *and* the evidence summary, so this is one word away from being defeated — and an agent that hallucinates confidently is exactly the one likely to mention a real symbol in passing.

I think the check wants to be "**is every** asserted mechanism grounded", not "is any". `all()` over the mechanisms would be the small version; a stricter reading would score only the mechanisms that look like internal code (dotted symbols, PascalCase classes) and ignore incidental snake_case.

## 2. The demotion changes routing for 3 categories — and not for the one #223 is about

This surprised me, so I checked `_ROUTING` directly. Only four categories park, and `triage()` only overrides on `route.parked`:

```
parked=True   TRANSIENT_INFRA, QUOTA, DEPENDENCY_WAIT, USER_PAUSED
parked=False  MISSING_ACCESS, AMBIGUITY, SCOPE_EXPLOSION, IMPOSSIBLE, NOVEL_UNKNOWN, ...
```

`MISSING_ACCESS` — the category in #223 and in your repro test — is `parked=False`, so it escalates at 0.95 and at 0.50 alike. Your `assert triage(b).target_status == TaskStatus.ESCALATED` therefore passes with or without the fix; the `confidence == 0.5` assertion above it is the one doing the work. Worth either dropping that line or moving the repro to a parked category so it pins something.

Where it *does* change routing, it can misfire. Two descriptions of the same honest transient failure:

```
"The runner hit a rate_limit and retry_after was not honoured"  -> 0.50 -> ESCALATED, notify=True
"The runner was throttled and the retry was too early"          -> 0.95 -> BLOCKED,  notify=False
```

Identical situation, opposite routing, decided by whether the agent used underscores. A transient blip that should park and auto-retry now wakes a human — which is the resource #223 exists to protect. The extractor treating the taxonomy's own vocabulary as a "code mechanism" is the same root: `MISSING_ACCESS` and `rate_limit` both match pattern 7.

Excluding the taxonomy's own category names and the field names of `Blocker` itself would remove most of this class cheaply.

## 3. Missing CLA ledger entry

There's no `contributors/ayushcipher.md`, so the **CLA ledger** check will fail and I can't land a foreign author without it. Copy the shape of any existing file in `contributors/` — it's three lines. (Filename lowercased: CI does `tr '[:upper:]' '[:lower:]'` on your handle.)

## What's already right

- Manifest is genuinely correct: `--strict` alone gives `OK: 1580 file(s)`, same count as `main`, so your re-pin is clean.
- `138 passed` across `test_blockers.py`, `test_blocker_challenge.py`, `test_blocker_answer_reuse.py`, `test_api_legacy_blocker_list_fields.py` — matches your number exactly.
- Branch is current with `main` (merge-base `cfad1a28`).
- No security surface: pure text heuristics, no new I/O, nothing touching credentials or the command guards.
- The `tried` corpus does work, even though nothing tests it — the six test names in your description don't match the six in the diff, and the `tried` one listed there isn't in the file. Worth a quick pass so the description matches what landed.

Two notes on process, neither your fault: CI hasn't run here because fork PRs from first-time contributors wait on my approval — I'll enable it so your next push gets real signal. And I'm still running the full suite locally; if it turns up anything beyond the above I'll add it rather than leave you guessing.

The core instinct here is good and I'd like this to land. Mostly it needs the grounding check to be stricter and the extractor to stop firing on our own vocabulary.

eyalgolan:
Closing the loop on the full suite I said I was running — it's clean:

```
$ python -m pytest -n 4 -q
12342 passed, 227 skipped in 951.43s
```

So nothing to add beyond the review above: no regressions anywhere else in the tree. The three items stand (the `any` vs `all` grounding check, the extractor firing on our own taxonomy vocabulary, and the missing `contributors/ayushcipher.md`), and none of them is a correctness problem in the rest of the codebase — they're all inside the new logic.

CI is enabled on your branch now too, so your next push will report properly instead of sitting at "action_required".

eyalgolan (changes requested):
Re-reviewed at `14198a5b`. **Both of my findings are genuinely fixed**, and I checked them the same way I found them rather than taking the commit message's word:

```
bare hallucination                         -> 0.50   (was 0.50)
same claim + one grounded token appended   -> 0.50   (was 0.95 — the defeat, now closed)

extract_code_mechanisms("Category MISSING_ACCESS applies: …")  -> []
extract_code_mechanisms("hit a rate_limit and retry_after …")  -> []
```

The `any` → `all` change closes the padding attack properly, and dropping the taxonomy's own vocabulary removes the class of false positive I flagged. `140 passed` across the four blocker suites. CLA ledger is in. Nice work — that was a quick, accurate turnaround.

Two things before this can land.

## 1. The manifest doesn't list your new file

```
FAIL: 1 tracked file(s) not listed in RELEASE_MANIFEST.txt:
  contributors/ayushcipher.md: tracked but not listed
```

`git add` the file first, then `python scripts/check_release_manifest.py --write`, then commit both. The **File inventory** check will be red until then. (Entirely our fault for how discoverable that is — I only fixed the CONTRIBUTING wording for it yesterday.)

## 2. `all()` moved the false-positive class one layer out, onto hostnames

The taxonomy vocabulary is gone, but the dotted-symbol pattern can't tell `ClaudeBackend._options` from a hostname:

```
DEMOTED  mechs=['api.anthropic.com']     "The connection to api.anthropic.com was reset"
DEMOTED  mechs=['github.com']            "github.com returned 502 for three attempts"
DEMOTED  mechs=['registry.npmjs.org']    "A DNS failure for registry.npmjs.org stopped the install"
DEMOTED  mechs=['requirements.txt']      "requirements.txt lists a package the index no longer serves"
DEMOTED  mechs=['PyPI']                  "The wheel on PyPI is 1.2.3 but the lock pins 1.2.4"
```

With `any()` these survived on some other grounded token. With `all()` a single unechoed hostname demotes the whole hypothesis — and these are exactly the sentences a `TRANSIENT_INFRA` blocker writes. That category **parks**, so the demotion changes routing:

```
"The connection to api.anthropic.com was reset mid-stream"  -> 0.50 -> ESCALATED, notify=True
"The upstream connection was reset mid-stream"              -> 0.95 -> BLOCKED,  notify=False
```

Same failure, same evidence. Naming the host you couldn't reach wakes a human; being vague about it auto-retries. That's the wrong incentive — it rewards less specific reports, which is the opposite of what #223 wants.

Cheapest fix in the same spirit as your taxonomy exclusion: drop tokens whose last dotted part is a TLD-ish word (`com`, `org`, `io`, `net`, `dev`, `ai`) or a known data extension (`txt`, `md`, `lock`, `cfg`), before the grounding check. A tighter version would require a dotted symbol to have a code-ish shape — a `_`-prefixed or PascalCase part — but the exclusion list matches what you already did and is easier to review.

## Still worth knowing, and not yours to fix

I checked `_ROUTING` while looking at the above: only `TRANSIENT_INFRA`, `QUOTA`, `DEPENDENCY_WAIT` and `USER_PAUSED` park, and `triage()` applies the low-confidence override **only** to parked routes. `MISSING_ACCESS` — the category in #223's example and in your repro test — is not among them, so it escalates at 0.95 and 0.50 alike and the demotion cannot change its routing. Your `assert triage(b).target_status == TaskStatus.ESCALATED` passes with or without your fix; the `confidence == 0.5` line above it is what's doing the work.

That's a gap in the issue, not in your PR — I've written it up on #223. Worth deciding there whether #223 wants a routing change or a report-surface change, because confidence isn't the lever for the category it names.

CI is approved on this push, so you'll get real signal this time.

eyalgolan:
CI has now run properly on this PR — first time, since fork PRs from first-time contributors were sitting unapproved (that's #314, our problem not yours).

The result is clean confirmation of the review above: **`File inventory` is the only red.** Everything else passes, including `Python`.

```
fail   File inventory
pass   Python
pass   CLA ledger
pass   ASCII locale (collection)
pass   Web board
pass   Wheel carries the board
pass   Container images build; the MCP one answers a handshake
```

So the manifest re-pin is genuinely the last mechanical thing standing between this and a green board:

```bash
git add -A
```

`python scripts/check_release_manifest.py --strict` should then print `OK`.

AyushCipher:
Hi @eyalgolan, the latest CI run (#779) for PR #309 is showing “Action required” again. The branch is clean and the release manifest check passes locally. Could you please approve the workflow run when you get a chance? Thanks!

eyalgolan:
Re-checked at `26e1d560`. The manifest is clean now — `--strict` alone gives `OK: 1587 file(s)`, so **File inventory** should go green on this push.

The hostname finding from my last review is unchanged, though, and it's the one thing I'd still want fixed:

```
DEMOTED  mechs=['api.anthropic.com']     "The connection to api.anthropic.com was reset"
DEMOTED  mechs=['github.com']            "github.com returned 502 for three attempts"
DEMOTED  mechs=['registry.npmjs.org']    "A DNS failure for registry.npmjs.org stopped the install"
DEMOTED  mechs=['requirements.txt']      "requirements.txt lists a package the index no longer serves"
```

No rush if you're mid-way through it — I'm flagging that it's still open rather than asking again.

The reason I keep coming back to it: `TRANSIENT_INFRA` is one of the four categories that parks, so the demotion changes routing there, and it changes it in the wrong direction:

```
"The connection to api.anthropic.com was reset mid-stream"  -> 0.50 -> ESCALATED, notify=True
"The upstream connection was reset mid-stream"              -> 0.95 -> BLOCKED,  notify=False
```

Same failure, same evidence. Naming the host you couldn't reach wakes a human at 3am; being vague about it parks and auto-retries. For an issue whose whole point is making blocker reports *more* trustworthy, rewarding vagueness is backwards.

The cheapest fix stays what I suggested last time and matches what you already did for the taxonomy vocabulary: drop a dotted token whose last part is a TLD-ish word (`com`, `org`, `io`, `net`, `dev`, `ai`) or a data extension (`txt`, `md`, `lock`, `cfg`) before the grounding check.

Everything else from the last round is genuinely closed and I re-verified it rather than assuming: the padding defeat is gone (`0.50` with a grounded token appended, where it used to survive at `0.95`), and the taxonomy vocabulary no longer extracts.

eyalgolan (changes requested):
Re-reviewed at `368e67b3`. **All three findings are closed**, and I checked the thing an exclusion list usually gets wrong — whether it over-reaches and quietly makes the guard inert:

```
ClaudeBackend._options is missing permission_mode   -> ['ClaudeBackend._options', 'ClaudeBackend', 'permission_mode']
the orchestrator calls make_backend() without config -> ['make_backend']
self._attempt_backend is never reset                 -> ['self._attempt_backend', '_attempt_backend']
core/orchestrator.py:2454 reads the mode inline      -> ['core/orchestrator.py:2454', 'orchestrator.py']

api.anthropic.com / github.com / registry.npmjs.org / requirements.txt / PyPI  -> []
```

Real mechanisms still register, hostnames and data files don't. And the behaviour that matters end-to-end still holds:

```
#223's hallucination, bare      -> 0.50
same claim + grounded padding   -> 0.50
honest hostname hypothesis      -> 0.95
```

That's the incentive pointing the right way now: naming the host you couldn't reach no longer costs you an escalation. 141 blocker tests pass, and you added 9 tests for the exclusion rather than trusting it.

## One thing left, and it's a merge artifact rather than your change

```
  .nh-local: listed but not in the tree
```

`main` removed `.nh-local` in `326dd067` and dropped its manifest row in the same commit. Your branch merged `main`, and git merged the manifest *cleanly* while keeping the stale row — `grep -c '\.nh-local' RELEASE_MANIFEST.txt` is `0` on main and `1` on your branch.

This one bites everyone, because a generated file merges without conflict and is wrong anyway. The recipe we use for it:

```bash
git checkout origin/main -- RELEASE_MANIFEST.txt   # take main's, don't merge it
python scripts/check_release_manifest.py --write   # re-derive on your tree
git diff origin/main -- RELEASE_MANIFEST.txt       # should be ONLY your files
```

The third line is the point: after re-deriving, the diff against main should contain exactly the rows for files *you* changed and nothing else. If `.nh-local` or anything unrelated shows up, the merge went wrong rather than your edit.

Fix that and I'll approve — the substance is done as far as I'm concerned, and this was a genuinely good iteration. You closed a finding, I found the next layer out, you closed that too, and the guard is sharper each round without becoming a blunt instrument.

(This is the fourth manifest round across contributors this week, all mechanical, none anyone's fault — #314 tracks making the tooling carry it instead of you.)

eyalgolan:
Thanks for this @AyushCipher — the problem you're targeting is real and worth fixing: a blocker report that asserts a root cause at high confidence without support is exactly what makes escalations expensive to read. I gave it an independent review and then reproduced the findings myself against the shipped predicates. The direction is right; the gate as written doesn't hold yet.

Measured with `extract_code_mechanisms` / `is_code_mechanism_supported` on this branch, with a control:

```
verdict      mechanisms                    case
KEEPS HIGH   ['ClaudeBackend._options',…]  evidence merely RESTATES the hypothesis
KEEPS HIGH   []                            same claim, no backticks, no evidence
demoted      ['ClaudeBackend._options',…]  same claim WITH backticks, no evidence  <- control
KEEPS HIGH   []                            EMPTY hypothesis, confidence 0.99
demoted      ['permission_mode']           correct symbol + prose evidence
```

Four things follow from those rows.

**1. The gate is satisfied by the model restating itself.** Row 1's evidence is `"I believe ClaudeBackend._options may be setting permission_mode incorrectly."` — the same claim, reworded. Both fields are written by the same model in the same turn, so "supported" currently means only "the identifier appears somewhere in evidence+tried+goal". The control on row 3 shows the gate does fire otherwise, so this is specifically the self-reference case.

**2. Removing the backticks removes the gate.** Rows 2 and 3 assert the same thing; only the formatting differs. `is_code_mechanism_supported([], …)` returns `True`, so a hypothesis phrased in prose is never checked. As written, the gate asks the model to format its claim, not to support it.

**3. It fails open on absent input.** An empty (or whitespace-only) hypothesis with confidence 0.99 keeps 0.99. Absent input should be a failure, not a pass.

**4. The failure direction is backwards on the commonest real shape.** Row 5 is an agent that read the source, named the symbol correctly, and summarised its evidence in prose — and it gets demoted, while rows 1–3 sail through.

**One more, which decides whether any of this takes effect:** the `0.6`/`0.5` constants are hardcoded, but the threshold is user-settable — `escalate_on_low_confidence_below` (defined in `config.py`, read in `orchestrator.py`, documented in `docs/configuration.md`). Set it to 0.4 and the demotion changes nothing. The code comment "Demote confidence below the escalation threshold (0.6)" is only true at the default.

**Also worth changing before this lands:** `_build_excluded_terms` hand-lists ~45 terms, and that list will go stale — `rate_limit` is excluded but `rate_limiting` isn't, `timed_out` is but `timing_out` isn't. This repo has been bitten by hand-written closed sets repeatedly; if you can express the rule rather than enumerate the members, it'll survive. Relatedly, `test_non_code_mechanism_hypothesis_retains_confidence` currently pins the bypass in point 2 as intended behaviour, so it would keep passing after a fix — worth revisiting alongside it.

**Suggested shape:** make "supported" mean something a restatement cannot satisfy — evidence that names a file, a line, a command and its output, or a test id — and treat an unparseable or empty hypothesis as unsupported. Then read the threshold from config instead of hardcoding it. Tests written from the failing case first would be ideal here: the current suite is green against all four rows above.

Small housekeeping: `contributors/ayushcipher.md` has an empty `Name:` field — every other ledger entry fills it, and under our identity rule your commits land authored as you, so it's worth setting.

Happy to look again once you've had a pass at it.

eyalgolan:
Re-checked at `88b87df4`. The substance is unchanged and still correct — I re-ran it rather than assuming:

```
hallucination, bare            -> 0.50
same claim + grounded padding  -> 0.50
honest hostname hypothesis     -> 0.95
ClaudeBackend._options ...     -> ['ClaudeBackend._options', 'ClaudeBackend', 'permission_mode']
138 passed
```

The manifest row is still there, though, and I think the recipe I gave you last time was the problem rather than you:

```
main (0e5b9249):  .nh-local rows = 0,  --strict OK: 1593
your branch:      .nh-local rows = 1
```

You merged `main` twice since, and the row survived both. That is the merge working as designed: git sees no textual conflict in `RELEASE_MANIFEST.txt`, so it keeps your side, and your side still lists a file `main` deleted. **Another merge will not fix it** — which is what I should have said explicitly.

The line that actually matters is the first one, and it is not a merge:

```bash
git fetch origin
git checkout origin/main -- RELEASE_MANIFEST.txt   # take main's file wholesale
python scripts/check_release_manifest.py --write   # re-derive against YOUR tree
python scripts/check_release_manifest.py --strict  # expect: OK
```

`git checkout origin/main -- <file>` overwrites your copy instead of merging it. Then `--write` re-adds the rows for the files you actually changed. Worth sanity-checking with `git diff origin/main -- RELEASE_MANIFEST.txt` before you commit: it should show only your files and nothing about `.nh-local`.

Your branch is also not descended from current `main` yet, so merge or rebase at the same time.

Sorry for the extra round on this — the generated-file merge is genuinely counter-intuitive, and my first version of the recipe buried the important line in a list. Everything else here has been right for two rounds.

eyalgolan (approved):
Approving. Four rounds, and each one closed a real finding without the fix becoming a blunt instrument — that last part is the hard bit and you got it right.

## Verified

`main` (`0e5b9249`) is an ancestor of your head, so what I checked is what lands.

```
--strict (alone)   OK: 1594 file(s)     .nh-local rows: 0
full suite         12605 passed, 227 skipped, 2 xfailed, 1 failed
CI                 all green
```

The one failure is `test_base_tree_gate.py::test_pre_existing_red_test_excused_when_runner_rewrites_the_command`, and it **fails on `main` too** — I ran it there to be sure. Not yours.

Re-ran the probes that found each finding, rather than reading the diff:

```
bare hallucination              -> 0.50
same claim + grounded padding   -> 0.50     <- the padding defeat, closed
"Category MISSING_ACCESS ..."   -> mechs=[] <- taxonomy vocabulary, excluded
ClaudeBackend._options ...      -> ['ClaudeBackend._options', 'ClaudeBackend', 'permission_mode']
```

And the one I cared about most, because it was the reason the whole thing mattered:

```
"The connection to api.anthropic.com was reset"  -> 0.95 -> BLOCKED
"The upstream connection was reset"              -> 0.95 -> BLOCKED
```

Identical. A blocker no longer costs someone a 3am escalation for naming the host they couldn't reach, and vagueness buys nothing. That was the point.

## What earned it

The risk with an exclusion list is a guard that stops guarding, so I went looking for that specifically — and real mechanisms still extract, including dotted symbols, `self._attr` forms, `make_backend()` calls and `path.py:LINE` citations. You narrowed it without hollowing it.

You also added nine tests for the exclusion rather than trusting it, which is why I could check this quickly instead of reconstructing your reasoning.

## Two things on me

The manifest cost you three extra rounds. The recipe I first gave buried the line that mattered (`git checkout origin/main -- RELEASE_MANIFEST.txt` — a merge cannot fix it, because git resolves a generated file cleanly and wrongly). That is now #314, so the tooling carries it instead of contributors.

And a finding beyond this PR, from reading `_ROUTING` while checking your work: **confidence is not the lever for `MISSING_ACCESS`** — it is `parked=False`, and `triage()` only applies the low-confidence override to parked routes, so it escalates at 0.95 and 0.50 alike. Your repro test's `triage` assertion passes with or without the fix; the `confidence == 0.5` line above it is what does the work. That is a gap in **#223**, not in your PR, and I have written it up there.

Per the usual caveat: this approval is clean as of `0e5b9249` and expires on the next landing.

Issue #312, opened 2026-09-12
A space in an installer path launders every venv-guard check, in both readings


eyalgolan:
Split out of #105 / PR #301, where it survived three rounds of fixes and a fourth attempt this evening. It is a distinct defect with a distinct shape, and bolting it onto that PR is not working.

## The defect

A quoted payload's own quoting is consumed by the OUTER lex, so re-lexing splits an installer path at a SPACE. The split is at the space, not at the separator, so the alternate Windows reading cannot help — the token is destroyed in both readings, and silently: the mangled `Files/proj/.venv/bin/pip` names no installer, so not even the allow-and-log WARNING fires.

Measured against a primary checkout whose path contains one space, on `56d6c480`:

```
SPACED  ALLOW   {p}/.venv/bin/pip install foo          NOSPACE  DENY
SPACED  ALLOW   cd {p} && uv sync                      NOSPACE  DENY
SPACED  ALLOW   env -i {p}/.venv/bin/pip install foo   NOSPACE  DENY
SPACED  ALLOW   timeout 300 {p}/.venv/bin/pip install  NOSPACE  DENY
```

Four of the eight `BYPASS_TEMPLATES` that `tests/test_windows_command_readings.py` already refuses flip to ALLOW purely because of the space. `guard.evaluate` allows the top-level case end to end, so this reaches the PreToolUse boundary.

`C:\Program Files\...` and `C:\Users\<First Last>\...` are the ordinary Windows shapes, so this is not an exotic input.

## Why the obvious fix doesn't hold

PR #301 tried rejoining leading tokens to rebuild the lost path (`cmd` itself resolves the longest leading prefix that names an executable). Three attempts, each closing part and leaving part:

1. **Nested payloads only** — the top-level spelling, which is this issue's own reproduction, walked straight through.
2. **Anchored at token 0** — `cd X && <spaced>/pip install y`, `echo hi && ...`, `timeout 5 ...`, `env -i ...`, `VAR=1 ...` all moved the installer off the front and re-opened it. It also joined ACROSS `&&`, producing tokens that can never resolve.
[...]

PR #320, opened 2026-09-13
A trailing .exe no longer walks past the never-merge and never-push gates

### LINKED ISSUE #105: Command guards tokenise with POSIX shlex, so Windows path separators are deleted and the venv guard fails open

Sreekant13:
**Refs #305** (not Closes: the raw-text half is still open, see below).

All five rows in the issue reproduce on real Windows, and fixing them turned up two things the issue does not say.

## Ten sites, not seven

The issue lists seven. There are ten, and **the three it misses are the gates themselves**: `_git_push_invocations`, `_forge_invocations` and `_git_invocations`.

I found that by asserting the count rather than editing the listed line numbers — the wiring script expected 7, found 10, and stopped. Had it not, the fix would have landed with never-push and never-merge still open, which is the exact failure mode this thread keeps producing.

## Three defects at those sites, not one

The issue is about the suffix. Measuring its rows turned up two more, and a fix for any one alone leaves the others open:

```
git.exe push origin main       the suffix is never stripped
GIT push origin main           never case-folded, though on Windows
                               GIT and git are the same file
C:/tools/git.exe push ...      PurePosixPath does not split a backslash
```

`GIT push origin main` is the sharpest of the three: **no suffix, no path, and it was allowed.**

## Shape

New `agent/exec_names.py`, split out for the same reason `fs_roots.py` was: `guard.py` is frozen at 2892 and sits exactly at the line. It gains no lines — the import grew a name in place, all ten call sites were edited in place, and I verified 2892 before and 2892 after.

`is_windows` is a **parameter**, following `fs_roots.is_windows_filesystem_root`, so both readings are reachable from a POSIX runner and the Windows behaviour is tested on the same CI that tests the POSIX behaviour rather than trusted.

The gating matters in both directions, and I took the reasoning from your own comments:

- On POSIX a backslash is a legal filename character, so splitting would invent a name the user never wrote — the same argument you made for not picking a lexing.
- On POSIX `GIT` and `git` are different files, so folding would be a text match masquerading as a structural one — `_is_installer_name` already says this for the venv guard.

The `.exe` strip stays **ungated**, following `_basename`'s precedent from #107 rather than inventing a second rule for the same suffix.

## What this does not close, asserted rather than left unsaid

An **unquoted** backslash path still passes:

```
C:\tools\git.exe push origin main     -> still ALLOW
"C:\tools\git.exe" push origin main   -> DENY (closed here)
```

POSIX `shlex` deletes the separators before any name resolution, so it arrives as `C:toolsgit.exe` and no basename can recover it. That is the half of #105 that #301 fixed for the venv guard only: `guard.py` does not consult `win_readings`. `_basename`'s docstring already assumes a caller has been offered the `/`-normalised spelling by then, and `guard.py` never is. A space splits the token additionally, which is #312.

`test_the_unquoted_backslash_path_is_still_open_and_that_is_recorded` asserts that gap, so whoever closes #105's remaining half sees it go red and deletes it deliberately, instead of the boundary being rediscovered from scratch a third time.

## Verification

```
34 test cases
revert all ten sites -> 9 failed, each naming its own spelling
zero regressions     -> 37 failures across guard, venv-guard and
                        structural-budget, before and after, same checkout
guard.py             -> 2892 lines before, 2892 after
```

Manifest repinned, `--strict OK: 1588 file(s)`, reanchor `VERDICT=OK`, no changed file carries CRLF.

---

## Updated after review: widened, and no longer claiming to close the issue

**`Closes` over-claimed and is now `Refs`.** Three of the rows review measured survive, and no name resolver can reach them, because they decide through **raw-text matchers** rather than argv[0]:

```
rm.exe -rf /                    _RM_RF matches text, not a basename
echo "gh.exe pr merge 7" | sh   lexical merge-stack matcher
timeout 30 git.exe push ...     timeout is not in _WRAPPERS
```

`test_the_raw_text_matchers_are_still_open_and_that_is_recorded` asserts that residual, so whoever widens those patterns sees it go red and removes it deliberately.

**Every spelling Windows executes, not only `.exe`.** The point that decided it is review's: a CLI installed by scoop or npm is `gh.cmd`, so `.exe` alone closes the spelling a reviewer thinks of first and leaves the one users have. Seven more rows now deny:

```
gh.cmd / gh.ps1 / nh.bat / git.com / gh. / nh.exe. / gh.exe::$DATA
```

Trailing dots and the NTFS `::$DATA` suffix are Windows-gated, because Win32 resolves both to the same program while on POSIX a trailing dot and a colon are ordinary filename characters.

## The per-site measurement found a hole in my own tests

Reverting each of the ten on its own, against the three suites:

```
caught individually  1725, 2235, 2494
NOT caught           287, 446, 532, 1017, 1432, 1640, 2531
```

Rather than assume those seven were redundant, I drove a 23 command corpus through `evaluate` in both readonly modes with each site reverted. **Six are individually reachable, and three guarded a bypass nothing in this file covered:**

```
revert 446 / 532 / 1017 -> grep.exe -r secret /   DENY becomes ALLOW
                           rg.exe secret /        DENY becomes ALLOW
                           find.exe / -name x     DENY becomes ALLOW
```

A whole-volume read that the bare spelling refuses. Now pinned. I would not have found it by reading, and your warning about writing six sites off as redundant is the reason I measured.

**What I am not claiming:** 287, 1432 and 1640 move no verdict in this corpus. That is not evidence they are redundant — 1725 moves none either, yet four tests catch it, which means coverage here depends on the configuration passed to `evaluate`. Recorded as measured rather than concluded.

61 cases, passing under both the Windows and the forced-POSIX reading. `guard.py` 2892 before and after. 37 failures across the guard suites, identical to baseline.


eyalgolan (changes requested):
@Sreekant13 — this is the best-argued PR I've reviewed on this repo, and I want to say why before I get to the changes, because the reasoning is the part worth copying.

**You found ten sites where the issue said seven, by asserting the count instead of editing the line numbers I listed.** The three extra are the gates themselves. My own attempt at this (#307, draft) did the opposite — I read the sites by eye, missed four regexes, and shipped a body claiming I'd "probed a wider corpus" when I'd probed only `git`. Your method caught what my reading didn't.

`GIT push origin main` is the sharpest case in the PR and it's yours, not the issue's: no suffix, no path, previously allowed.

And the residual section is exactly right — an honest "what this does not close", with `test_the_unquoted_backslash_path_is_still_open_and_that_is_recorded` so whoever closes #105's other half sees it go red and deletes it deliberately. That's the discipline I've been failing at all week.

Verified independently:

```
guard.py          2892 before, 2892 after      (frozen budget holds)
--strict alone    OK: 1588 file(s)
guard suites      352 passed
revert all ten    11 failed
```

## Changes: `Closes #305` over-claims

The issue's title is *"A trailing `.exe` bypasses the command guard's name matching, **including the never-merge and never-push-to-main gates**"*. These still pass, measured on your tip with `_IS_WINDOWS=True`:

```
ALLOW  rm.exe -rf /                        _RM_RF is raw text, not argv[0]
ALLOW  gh.cmd pr merge 7                   scoop/npm install exactly this spelling
ALLOW  nh.bat approve 7 / git.com push origin main / gh.ps1 pr merge 7
ALLOW  nh.exe. approve 7 / gh. pr merge 7  Win32 strips trailing dots on resolution
ALLOW  gh.exe::$DATA pr merge 7
ALLOW  echo "gh.exe pr merge 7" | sh       raw-text matchers, not argv[0]
ALLOW  timeout 30 git.exe push origin main
```

(bare twins all DENY)

Your ten sites are all argv[0] resolution. The guard also decides through **raw-text matchers** — `_RM_RF`, `_FORGE_MERGE`, `_FORGE_MENTION`, `_FORGE_WRITE`, `_GIT_WRITE`, `_LEXICAL_MERGE_STACK`, `_LEXICAL_LIVE_SERVER` — which name binaries directly and never see a basename. `timeout`/`nice` aren't in `_WRAPPERS`, so those verdicts come from the raw-text half exclusively.

I hit this same wall on #307 and it took an adversarial review to show me. I'm not asking you to close all of it here — but either widen those matchers too, or change to `Refs #305` and say which half this covers. Merging as `Closes` would auto-close an issue whose `rm -rf` and never-merge-via-shell cases are still open.

`.cmd`/`.ps1` is the one I'd most want in scope: on Windows those are the *ordinary* spellings for a CLI installed via scoop or npm, so `.exe`-only closes the spelling a reviewer thinks of first and leaves the one users have.

## One suggestion, not a blocker

Per-site rather than all-at-once, against `test_exec_names.py` + `test_guard.py` + `test_venv_install_guard.py`:

```
revert ALL TEN   -> 11 failed
revert each one  -> 4 of 10 caught;  6 not: lines 287, 446, 532, 1017, 1432, 1640
```

Those six are plausibly redundant with each other — but I'd measure rather than assume. On my own PR I wrote off six sites as "redundant" and a reviewer showed all six were reachable, two only in readonly mode. A corpus diff per reverted site is what settled it.

## Overlap, disclosed

I have a draft PR (#307) on this same issue. Yours is CI-green, better scoped, and uses a cleaner seam — `is_windows` as a parameter following `fs_roots`, rather than my ungated strip, which also avoids an over-block class I had to file separately as #315. If you widen the raw-text matchers here, I'll close mine in favour of this and fold my extra findings (#312-adjacent trailing-dot cases) into follow-ups rather than competing.

Sreekant13:
Both addressed, and the suggestion turned out to be the more valuable of the two.

## `Closes` was over-claiming, and is now `Refs`

Every row you measured reproduces here. Three of them survive any name resolver, because they decide through **raw-text matchers** rather than argv[0]:

```
rm.exe -rf /                    _RM_RF matches text, not a basename
echo "gh.exe pr merge 7" | sh   lexical merge-stack matcher
timeout 30 git.exe push ...     timeout is not in _WRAPPERS
```

`test_the_raw_text_matchers_are_still_open_and_that_is_recorded` asserts that residual, so whoever widens those patterns sees it go red and deletes it deliberately.

## `.cmd`/`.ps1` was the right thing to prioritise

Your framing settled it: **`.exe` alone closes the spelling a reviewer thinks of first and leaves the one users have.** `_EXECUTABLE_SUFFIXES` is now `.cmd .bat .ps1 .exe .com`, longest-first. Trailing dots and the NTFS `::$DATA` suffix are stripped too, both Windows-gated, since Win32 resolves them to the same program while on POSIX a trailing dot and a colon are ordinary filename characters.

Seven more of your rows now deny: `gh.cmd`, `gh.ps1`, `nh.bat`, `git.com`, `gh.`, `nh.exe.`, `gh.exe::$DATA`.

## Your "suggestion, not a blocker" found a real hole in my tests

I reproduced your per-site numbers, then did the corpus diff you described rather than assuming the uncaught seven were redundant:

```
caught individually  1725, 2235, 2494
NOT caught           287, 446, 532, 1017, 1432, 1640, 2531
```

23 commands through `evaluate`, both readonly modes, each site reverted in turn. **Six are individually reachable, and three guarded a bypass nothing in my file covered:**

```
revert 446 / 532 / 1017 -> grep.exe -r secret /   DENY becomes ALLOW
                           rg.exe secret /        DENY becomes ALLOW
                           find.exe / -name x     DENY becomes ALLOW
```

A whole-volume read the bare spelling refuses, and I had no test for it. That is the second time this method has paid for itself on this PR — the first was asserting the site count and finding ten instead of seven. **Your experience of writing six sites off as redundant is the only reason I measured instead of concluding.**

**What I am deliberately not claiming:** 287, 1432 and 1640 move no verdict in this corpus. That is not evidence they are redundant. 1725 moves none either, yet four tests catch it, which tells me coverage here depends on the configuration passed to `evaluate`, not just on the site. Recorded as measured rather than concluded, and I would rather leave the question open than answer it from a corpus that was never designed to be exhaustive.

## State

```
61 cases, passing under both the Windows and the forced-POSIX reading
guard.py      2892 before, 2892 after
guard suites  37 failures, identical to baseline
manifest      --strict OK: 1588 file(s)      reanchor VERDICT=OK
```

On #307: if you would rather land yours and have me close this, say so. Mine has the corpus evidence and the residual test, but there is no sense in two PRs for one issue and I have no attachment to which one carries it.

Sreekant13:
Heads up on this PR's two red checks: **they are inherited from main, not from this branch.**

I brought the branch up to date so it would be mergeable on approval, and that pulled in `83b0b364`, which is red on `main` itself:

```
main @ 83b0b364    FAILED: Web board
                   FAILED: File inventory
```

Both trace to `.nh-local`, committed as a symlink pointing at `/Users/eyalgolan/git/snc/master/no_human/.nh-local`. Filed as #323 with the diagnosis rather than fixing it here, since it is one line in someone else's commit.

Worth one note if you reproduce on Windows: it looks fine there. Without symlink support git materialises mode `120000` as an ordinary file whose contents are the target path, so `os.stat` succeeds and only the manifest half fails. The ENOENT is Linux-only.

Nothing to do on this PR. The seven other checks are green, and once #323 lands I will re-run these two.

eyalgolan (approved):
Approving. This is the standard I want contributions to this repo held to, and I mean that literally — you held it better than I did on my own competing draft.

## Verified on the MERGE RESULT, not just the branch

The branch is `BEHIND` main, and a manifest merges cleanly while being wrong, so I merged `origin/main` (`5774bab6`) into your tip locally and checked there:

```
--strict (alone)     OK: 1593 file(s) match RELEASE_MANIFEST.txt
guard.py             2892 lines   (frozen budget holds)
guard suites         470 passed, 2 xfailed
full suite (branch)  12575 passed, 227 skipped, 0 failed
```

Behaviour on the merge result:

```
DENY  nh.exe approve 7          DENY  gh.cmd pr merge 7
DENY  gh. pr merge 7            DENY  git.exe push origin main
DENY  GIT push origin main
```

I also checked the thing that worries me about any widening — new false denials. One command denies that a developer would reasonably run (`python -m pip install -r requirements.txt`), and it denies identically on `main`, so it is pre-existing and not yours.

## What earned this

**You found ten call sites where the issue said seven**, by asserting the count instead of editing the line numbers I listed. The three extra were the gates themselves. My own attempt at this issue read the sites by eye, missed four regexes, and shipped a body claiming I had "probed a wider corpus" when I had probed only `git`. Your method caught what my reading did not.

`GIT push origin main` — no suffix, no path, previously allowed — is yours, not the issue's.

**`Refs #305`, and the residual is pinned rather than described.** `test_the_raw_text_matchers_are_still_open_and_that_is_recorded` asserts the three raw-text shapes still pass, so whoever widens `_RM_RF` and the lexical matchers sees it go red, and the failure message tells them to delete it and promote the row. That is a boundary that maintains itself.

**The `is_windows` parameter over a host check.** Following `fs_roots.is_windows_filesystem_root` means both readings are reachable from a POSIX runner, so the Windows behaviour is tested on the CI we actually have rather than trusted. My version stripped unconditionally and, as a consequence, widened an operand over-block I had to file separately as #315. Yours does not have that problem because it only touches argv[0] resolution.

## Housekeeping

The branch is behind `main`; merge or rebase before landing. Per the usual caveat: **this approval is clean as of `5774bab6` and expires on the next landing** — if main moves under you, the merge result is a different tree than the one I checked.

I am closing my own #307 in favour of this. It targeted the same issue and yours is the better change.

Sreekant13:
Merged, thank you — and one thing I got wrong on the way in that you should know about.

**The merge auto-closed #305, which is exactly what you asked me to prevent.** I changed this PR's body to `Refs #305`, but my first commit still said `Closes #305.`, and the squash concatenated every commit message into `d34a9c5b`. GitHub read it and closed the issue while the raw-text residual is still open. I checked the body before merging and not the squash message.

**Reopened #305** with the three still-open rows and the cause, so it is back in the state it should be in.

Separately, and to be clear it is not a problem: the squash carries `Co-authored-by: Eyal Golan` because `36e3e286` is a `Merge branch 'main'` you ran on the branch. That credit is correct.

On the merge itself I held to your condition: your approval named `5774bab6` and expired on the next landing, so I confirmed main was still exactly `5774bab6` before merging rather than trusting that it had not moved.

Closing #307 in favour of this was generous. The ten-sites finding only exists because of how you told me to measure.

PR #317, opened 2026-09-12
A cleanup that partly failed is reported as a crashed eval's disk leak

eyalgolan:
> **Review passed** (4 rounds) · **Tests passed** (12620 passed, 0 failed) · **Merge policy** not ready

## Evidence
| Check | Result |
|---|---|
| Independent review | ✅ **PASSED** — 4 rounds · [proof](https://github.com/no-human-ai/no_human/blob/b1a94ee4fc5a2f692ad77488815111371c9924c0/.nh-evidence/811fffb9e68544cd82892d09516f0267/review.md) |
| Verifiers | ✅ 1 of 1 satisfied · [proof](https://github.com/no-human-ai/no_human/blob/b1a94ee4fc5a2f692ad77488815111371c9924c0/.nh-evidence/811fffb9e68544cd82892d09516f0267/verifiers.md) |
| Test-change guard | ⚠️ fired — waived as LEGITIMATE (2) — see below · [proof](https://github.com/no-human-ai/no_human/blob/b1a94ee4fc5a2f692ad77488815111371c9924c0/.nh-evidence/811fffb9e68544cd82892d09516f0267/tamper.md) |
| Tests | ✅ PASS — 12620 passed, 0 failed, 0 errors · [proof](https://github.com/no-human-ai/no_human/blob/b1a94ee4fc5a2f692ad77488815111371c9924c0/.nh-evidence/811fffb9e68544cd82892d09516f0267/tests.md) |
| CI | failure |
| Merge policy | ❌ not ready — 1 of 6 rules failed: ci · [proof](https://github.com/no-human-ai/no_human/blob/b1a94ee4fc5a2f692ad77488815111371c9924c0/.nh-evidence/811fffb9e68544cd82892d09516f0267/merge-policy.md) |

<details><summary>3 findings raised and addressed across review rounds</summary>

- New test read_text lacks encoding — fails existing gate — tests/test_eval_sandbox_cleanup.py:66 calls `marker.read_text()` with no encoding=. The repo's own gua
- diagnose() pushed over the 300-line structural budget — The ~25 lines this diff added inline to diagnose (the residue-measuring advisory block replacing a 10-li
- pre-review test run — the harness's own pre-review test run was RED — failing: tests/test_structural_budget.py::test_no_new_oversized_functions, tests/test_text

</details>
<details><summary>Verifiers (1)</summary>

- ✅ tests-assert-something — 2 files

</details>

### Test-change adjudication
[...]

PR #321, opened 2026-09-13
UI evidence embeds one frame per page state, not one per walk step

eyalgolan:
## The defect, as shipped

PR #317's UI-evidence block rendered three embeds — `landing`, `landing-settled`, `final` — asserting the walk observed three page states. It observed one. The three files are the same file:

```
$ curl -sSLO <evidence-branch raw URL>/{landing,landing-settled,final}.png
landing.png               81304 bytes  f57898ac76324962
landing-settled.png       81304 bytes  f57898ac76324962
final.png                 81304 bytes  f57898ac76324962
```

The block also posted `[walk video](...)`, which reads as an inline player.

## Why the obvious video fix does not work

Not a markdown problem — a hosting one. Measured against the raw host that serves every evidence branch:

```
$ curl -sSLI <...>/walk.webm
content-type: audio/webm
content-disposition: attachment
x-content-type-options: nosniff
```

The browser is told it is audio, told not to sniff, and told to save it. A `<video>` tag pointing there renders an empty box. (A PNG from the same host comes back `content-type: image/png` with no attachment header — so the images were never the problem.)

An inline-renderable GIF would need `ffmpeg` or Pillow. Neither is a dependency of this project and adding one is an operator decision, so this PR does not take it; it makes the line say what the link actually does.

## The fix

`testing/ui_evidence.py` gains `frame_runs`, `repeat_note`, `frame_lines`, `shot_record` and `video_line`. Rendering moved out of `core/orchestrator.py`, which already owns too much and is on a ratchet-down-only budget — it comes out **two lines shorter** than the frozen 23893.

- Only a **consecutive** repeat groups. An A/B/A walk really did change and change back, so each step keeps its own frame.
- A shot carrying no digest **never** groups, so a digest that could not be taken shows *more* frames rather than silently merging distinct ones.
- The digest is taken in the delivery loop where the bytes are already read — no second read.

## Evidence

4 new tests: 2 were red before the fix and green after; 2 are controls that were green **both** before and after. The controls are the load-bearing ones — they are what prove the grouping keys on the bytes rather than on the count, and `test_a_frame_repeating_after_the_page_changed_is_still_its_own_step` is the case a naive dedupe gets wrong.

- `tests/test_structural_budget.py`, `test_ui_evidence*.py`, `test_pr_evidence.py`, `test_evidence_ledger.py`, `test_tamper_guard.py`: **173 passed**
- Replayed the real #317 bytes through the fixed renderer: **1 embed instead of 3**, plus a line naming the two repeats
- Manifest regenerated: 1586 rows before and after, exactly the changed files repinned


eyalgolan:
Landed by hand as `680d6889` after review, re-cut onto current main.

Squashed with `eyalgolan` as both author and committer, per the repo's identity rule — which is why this is closed rather than merged through the GitHub button.

What changed from the branch as it stood:

- **The `(+N more shot(s))` count had no test at all.** It is the one existing number this change redefined — it was `len(delivered) - cap` and is now the sum of shot names inside dropped runs — and a mutation returning `len(dropped)` (runs instead of shots) left the entire file green. Now pinned by a test that builds six runs by alternating two images (runs group consecutively, so A,B,A,B,A,B is six runs from two files) and puts a three-shot repeat past the cap: counting runs says 1, counting shots says 3.
- **`shot_record`'s docstring described a branch that never executes.** It claimed the digest is taken "where the bytes have just been read for delivery". `_write_shot` is the only producer of shot dicts, always stamps `sha256`, and the orchestrator always forwards it — so the local hash is a fallback for hand-built dicts and tests. Corrected, and it now records the consequence: grouping attests to capture-time identity, not to the delivered bytes a reader fetches.
- **"Only a CONSECUTIVE repeat groups" was false as an unqualified claim.** The caller drops shots whose files are missing at delivery, so an A/B/A walk arrives as A/A if B's file vanished and the survivors do group. Qualified rather than removed.
- A test constant's comment described the same blob as both 1x1 and 2x1 in one sentence.

Also corrected in the squash message: the branch's own numbers were stale after the re-cut. Measured on the landed tree — `core/orchestrator.py` is 24073 against its frozen 24073 (untouched by this change), 7 tests added of which 5 were red before the fix, manifest at 1609 files.

Not folded in, and recorded on the relevant issue instead: `agent/guard.py` carries a stale-high frozen pin (2925 actual, 2926 frozen) left by an unrelated change. Correcting it here would have widened this diff's tamper surface and re-staled every open PR touching that file.

Issue #328, opened 2026-09-13
Case-folding in the guard: #320 landed the opposite answer to #307, and four gates are open on a case-insensitive host


eyalgolan:
Both #307 and #320 close issue #305. **#320 landed** (`d34a9c5b`). They disagree on one design question, and the disagreement is load-bearing rather than cosmetic.

## The disagreement

`exec_names.py` (from #320) states its position in its own docstring:

> On POSIX `GIT` and `git` are different files, so folding case there would be a text match masquerading as a structural one.

`venv_install_guard._basename` (from #307) states the opposite, and backs it by asking the OS:

```
$ cp <tmp>/git <tmp>/GIT
cp: <tmp>/GIT and <tmp>/git are identical (not copied).
$ ls <tmp>
git
```

One file, two spellings. Both positions are correct — about different filesystems. #320 is right for case-sensitive ones (Linux CI, most Linux users). #307 is right for case-insensitive ones, which is what **this machine and the dogfood host actually are**.

## What is open on main today

Measured against `origin/main` @ `d34a9c5b` with `guard.evaluate(..., never_push_to=["main"], cwd=None)`:

```
ALLOW  GH pr merge 7                       <- constraint #2: the agent never merges
ALLOW  RM -rf /
ALLOW  FIND / -delete
ALLOW  sh -c "GIT.CMD push origin main"
```

On a case-insensitive filesystem `GH pr merge 7` really invokes `gh pr merge 7`. The lowercase twins all deny, so this is a spelling gap, not a missing rule.

#320 correctly closes the `.exe` family that #305 was filed about (`nh.exe approve 7`, `gh.exe pr merge 7`, `git.exe push origin main` all DENY), and it closes the backslash family too. This is only about the casing half.

## Why I did not just merge #307 over it

Resolving the conflict either way silently picks a winner in a design argument that two reviewed changes disagree on. Reverting main's `exec_names.command_name` call sites back to `_basename` would also fight a structure that is better than what it replaced — #320 split the logic out of `guard.py` specifically because that file is at a ratchet-down budget, which is the right instinct.

## Suggested resolution

Fold when the filesystem is **measurably** case-insensitive, rather than assuming either way. Both docstrings are then true simultaneously, and neither host class is punished:

- on a case-insensitive volume, `GH` and `gh` are the same program, so folding denies nothing that could not already run;
- on a case-sensitive volume, folding is skipped, so no command anyone is entitled to run is refused.

[...]

PR #331, opened 2026-09-13
Strip scheduling wrappers before the merge and git gates

### LINKED ISSUE #329: Scheduling wrappers (taskset/chrt/ionice/setsid) launder past the merge and git gates: 'taskset 0x3 nh approve' is allowed
Scheduling/priority wrappers launder a command past the argv-based gates, and one shape reaches **constraint #2**: `taskset 0x3 nh approve 7` is **allowed**, so an agent can approve its own work by prefixing four characters.

Measured on `main` (`d34a9c5`) with the **Linux reading** (`_IS_WINDOWS=False`), since `taskset`/`chrt`/`ionice`/`setsid` are Linux binaries an agent on the fleet actually has. Every bare form denies; the wrapped forms:

```
gate \ wrapper                     taskset  chrt    ionice  setsid  nice   timeout  env
never-merge  nh approve 7          BYPASS   deny    deny    deny    deny   deny     deny
never-merge  gh pr merge 7         deny     deny    deny    deny    deny   deny     deny
destructive  git checkout -- .     BYPASS   BYPASS  BYPASS  BYPASS  deny   deny     deny
destructive  git stash             BYPASS   BYPASS  BYPASS  BYPASS  deny   deny     deny
destructive  git reset --hard HEAD deny     deny    deny    deny    deny   deny     deny
never-push   git push origin main  deny     deny    deny    deny    deny   deny     deny
```

Two confirmed bypasses:

- **`taskset 0x3 nh approve 7` → ALLOW.** `nh approve` is how a session ends the human merge gate; this is constraint #2, "the agent never merges."
- **`{taskset,chrt,ionice,setsid} git checkout -- .` and `... git stash` → ALLOW.** The working-tree-destructive git forms that the bare command refuses.

## Root cause

`_strip_wrappers` strips `_WRAPPERS = {env, sudo, nohup, command, exec, time, builtin}`, and `_peel_runners` peels `_SHELL_RUNNERS`. The scheduling/priority wrappers are in **neither** — they live only in `_SCAN_WRAPPER_NAMES = {chrt, ionice, nice, setsid, stdbuf, taskset, timeout, xargs}`, which only the *scan* gate (`_peel_scan_wrappers`) consults. So for the argv-based gates, `argv[0]` stays the wrapper name:

```
_strip_wrappers + _peel_runners on "taskset 0x3 nh approve 7"
  -> ['taskset', '0x3', 'nh', 'approve', '7']      argv[0] = 'taskset', never 'nh'
```

`taskset`/`chrt`/`ionice` take a **non-flag operand** (`0x3`, `10`, `-c3`) exactly like `timeout 30`, which is why `_peel_scan_wrappers` already walks past them for the scan gate — but the merge and git gates never got the same treatment.

## Why the coverage is uneven (and why that is the worrying part)

The cells that still deny — `gh pr merge`, `git reset --hard`, `git push` — are not caught by the argv gate. They survive only because a **separate raw-text matcher** (`_FORGE_MERGE`, `_GIT_WRITE`, the push detector) happens to match the string regardless of argv stripping. That is luck, not coverage: `nh approve` and `git checkout -- .` have no raw-text twin (the checkout-clobber decision needs argv + a cwd filesystem check), so for them the argv gate is the only line, and the wrapper walks straight through it.

So the same asymmetry the repo has been closing all week (#105, #305, #315): one spelling of a command is gated and an equivalent one is not.

## Suggested fix

Give the argv-based gates the same wrapper set the scan gate already uses. The cleanest shape is to route `_approve_denial`, `_git_invocations` and `_git_push_invocations` through the peel that already understands operand-taking wrappers (`_peel_scan_wrappers`, or a shared helper), rather than `_strip_wrappers` alone — so `taskset 0x3 <cmd>` reduces to `<cmd>` before the name match, on both readings.

I would rather confirm the exact placement against a corpus than assert it — the raw-text matchers that incidentally catch `reset --hard`/`push` must keep working, and a fix that only widens the wrapper set must not reopen anything. I have the harness from #305/#320 and am happy to take it if useful; flag if you would rather scope it yourself.

## Note on method

`nice`/`setsid`/`chrt` deny `nh approve` while `taskset` allows it, and I have not fully traced why the first three are caught (a secondary path denies them for reasons I have not pinned). I am reporting the matrix as **measured** and the `_WRAPPERS` omission as the **confirmed** root cause of the two bypasses; I am not claiming to explain every cell, because I could not, and would rather say so than publish a mechanism I have not verified.

Sreekant13:
Closes #329. Your diagnosis was exact, and it held up end to end — this is your two-line fix, measured rather than trusted.

## The fix

1. `taskset` into `_TRAILING_ARGV_RUNNERS` (the one scheduling wrapper that was missing; `chrt`/`ionice`/`setsid` were already there). `_FORGE_RUNNER_NAMES` picks it up for free.
2. `_git_invocations` reads `_FORGE_RUNNER_NAMES` instead of the narrower `_SHELL_RUNNERS`, the same set `_forge_invocations` already uses. The comment at `guard.py:2420` records this exact class fixed once for the forge path; the git path never got it.

Both are one existing line each, so `guard.py` stays **2892** (frozen).

## Matrix now fully green

Linux reading, every bare twin denies:

```
gate \ wrapper           taskset chrt ionice setsid nice timeout env
nh approve 7             deny    deny deny   deny   deny deny    deny
gh pr merge 7            deny    deny deny   deny   deny deny    deny
git checkout -- .        deny    deny deny   deny   deny deny    deny
git stash                deny    deny deny   deny   deny deny    deny
git reset --hard HEAD    deny    deny deny   deny   deny deny    deny
git push origin main     deny    deny deny   deny   deny deny    deny
```

Also green on the shapes not in the grid: `taskset -c 0-1` (range operand), `chrt -f 1` (flag operand), `env taskset 0x3 nh approve`, `sh -c "taskset 0x3 nh approve 7"`, `taskset 0x3 sh -c "git stash"`. So `Closes`, not `Refs`.

## Pinned exactly as you asked

**Per-change load-bearing**, each reverted in isolation:

```
remove taskset from the runner set   -> taskset 0x3 nh approve 7 reopens
git gate back on _SHELL_RUNNERS       -> setsid git stash reopens
both source changes reverted          -> 13 of 19 tests red
```

**The control against the wrapper set becoming the only defence:** `git reset --hard` and `git push` deny even with every scheduling wrapper stripped from the runner sets, because a raw-text matcher catches them independently. `test_reset_and_push_deny_independent_of_the_wrapper_set` empties the sets and asserts they still deny, so a future narrowing of the lists cannot silently reopen them.

**No false denials:** the change only widens what the gates *see*, so it cannot deny more than the bare command. `taskset 0x3 git status`, `git commit -am wip`, `git log` behind a scheduler all stay allowed, pinned in `test_a_legitimate_wrapped_command_is_not_newly_denied`.

## Verification

`tests/test_scheduling_wrappers.py`, 19 cases. Guard suites (`test_guard`, `test_exec_names`, `test_venv_install_guard`, `test_windows_command_readings`): **114 failures on this Windows host before and after the branch, identical, zero introduced** (the 114 are pre-existing platform/frozen-budget failures on Windows). Manifest repinned with the new test staged first, `--strict OK: 1594 file(s)`, reanchor `VERDICT=OK`, no changed file carries CRLF.

One scope note in your spirit: `_git_invocations:2541` still uses `PurePosixPath(tok).name == "git"` for the trailing-argv branch rather than `exec_names.command_name`, so a `git.exe` *trailing* token behind a wrapper is a separate gap — that is #305/#320 casing/suffix territory, not this, and I left it alone.


eyalgolan (approved):
Approving. Two lines, six closed bypass cells, and the verification is better than the fix — which is the right ratio for a guard change.

## Verified on the MERGE RESULT

Your branch was behind `main`, and a manifest merges cleanly while being wrong, so I merged `origin/main` (`4b9eb83f`) into your tip and checked there:

```
full suite      12624 passed, 227 skipped, 2 xfailed, 1 failed
--strict alone  OK: 1595 file(s)
guard.py        2892 lines (frozen budget holds)
```

The single failure is `test_base_tree_gate.py::test_pre_existing_red_test_excused_when_runner_rewrites_the_command`, which **fails on `main` too** — I ran it there. Not yours.

Matrix, re-measured independently on your branch:

```
open bypass cells: 0        (main: 6)
```

And the risk side, which is the part a wrapper widening can get wrong: 135 ordinary commands across nine wrapper prefixes, diffed `main` vs your branch — **0 new false denials, 0 newly allowed**. The change only widens what the gates *see*, exactly as your body claims.

Per-change mutations, each reverted in isolation: removing `taskset` from the runner set → 7 red; `_git_invocations` back on `_SHELL_RUNNERS` → 9 red. Both load-bearing.

## What makes this a good change rather than a correct one

`test_reset_and_push_deny_independent_of_the_wrapper_set` empties `_TRAILING_ARGV_RUNNERS` and `_FORGE_RUNNER_NAMES` and asserts `git reset --hard` and `git push` still deny. That pins the *second* line of defence, so a future narrowing of those lists cannot quietly make your widening the only thing standing there. Testing what would make the fix fragile, not just what it fixes — I'd point other contributors at it.

The `_linux` autouse fixture pinning `_IS_WINDOWS = False` "whatever the host" is the other thing worth naming. That is precisely the discipline I failed at on #320: I measured with `_IS_WINDOWS = True`, reported `GIT push origin main -> DENY`, and read it as covering both readings. It does not — on the default reading it is ALLOW, which is #328, still open.

## Residual: confirmed, and correctly scoped

I checked the gap your scope note names rather than taking it on trust:

```
DENY   taskset 0x3 git stash        DENY   git.exe stash
ALLOW  taskset 0x3 git.exe stash    ALLOW  xargs git.exe stash
```

So it is specifically a `.exe` token in *trailing-argv* position behind a wrapper — `guard.py:2541` still using `PurePosixPath(tok).name == "git"`. That is #305/#320 territory as you say, and leaving it alone here is right. Worth folding into whatever resolves #328, since it is the same resolver question.

`Closes #329` is justified: you tested beyond the grid (`taskset -c 0-1`, `chrt -f 1`, `env taskset …`, nested `sh -c`), and the matrix is green on all of them.

Per the usual caveat: clean as of `4b9eb83f`, expires on the next landing. Merge or rebase before landing since you are behind.

Issue #314, opened 2026-09-12
Fork PRs from first-time contributors get no CI until a maintainer approves, so they iterate blind


eyalgolan:
Every fork PR from a first-time contributor lands with **no CI at all** until a maintainer clicks approve. Until that click, the only check that reports is `CLA nudge` — it runs on `pull_request_target`, so it executes in base context and is exempt.

The contributor sees a PR with one green check and no other signal. They cannot tell whether their change passes.

## Measured, twice, tonight

Both open contributor PRs had sat un-run:

```
PR #308  CI  pull_request         completed  action_required
         CLA nudge  pull_request_target  completed  success
PR #309  (same shape)
```

I approved both runs manually. #308's then went red on `Python`; #309's had never reported either.

## What it actually cost

Both contributors missed the `RELEASE_MANIFEST.txt` re-pin — #308 changed six files without re-pinning, #309 added `contributors/ayushcipher.md` without listing it. Both would have been caught by `File inventory` in seconds.

This is **not** a documentation gap. `CONTRIBUTING.md:36-48` on `main` is explicit, and even names the exact case #309 hit:

> pins a SHA-256 of every tracked file in `RELEASE_MANIFEST.txt`, so **any change to a tracked file — including the `contributors/<handle>.md` you add to sign the CLA — has to re-pin it in the same commit**

Followed by the two commands. The instructions are there and prominent. What is missing is the feedback loop: they never saw a red check, so nothing corrected them, and the first correction came from a human review hours later.

Two contributors, two independent misses, same evening, same step. That is a systems signal, not two careless people.

## The setting

Settings → Actions → General → **"Fork pull request workflows from outside collaborators"**. Current behaviour matches *"Require approval for first-time contributors"*. Options, cheapest first:

1. **Loosen to "Require approval for first-time contributors who are new to GitHub."** Established accounts get CI immediately; brand-new accounts still gate. This is GitHub's default and closest to what we want.
2. **Keep the gate, approve promptly.** Works only if someone is watching; tonight both sat until I happened to look.
3. **Split the cheap checks out** onto `pull_request_target` so `File inventory`, `CLA ledger` and lint always report. Careful: `pull_request_target` runs in base context with secrets, so it must not check out or execute PR code. `File inventory` only hashes files, so it could be made safe — but this is the most work and the easiest to get wrong.

[...]

PR #335, opened 2026-09-13
fix(vcs): stage a caller-supplied broken symlink under its own name

### LINKED ISSUE #322: commit_paths resolves caller paths, so a broken symlink is committed away silently
Found during the independent review of #272 (landed as `93de0329`). Pre-existing root cause, orthogonal to that fix, so it was filed rather than folded in.

## The defect

`GitRepo.commit_paths` calls `Path(p).resolve()` on caller-supplied paths before the phantom filter runs. `resolve()` FOLLOWS a symlink, so a broken symlink is rewritten to its missing target, and the filter then evaluates "does this exist" against the target rather than against the link.

Measured, one caller-supplied broken symlink `note.md` plus one real file:

```
after #272 : commit_paths SUCCEEDED   HEAD tree: ['base.txt', 'real.py']   <- note.md silently gone
before     : commit_paths RAISED      GitError: git add -- gone-target real.py failed (128)
```

Main crashes loudly; the landed tree commits without the file. That is the silent-work-loss direction, which is the exact failure mode #272's predecessor was rejected for — just reached through a different door.

## Why it is narrow

`.py`-named broken symlinks survive, because `ls-files --others` re-adds them through the `_CODE_EXTS` producer. Only a non-code extension falls through. It also needs a coder to create a broken symlink **and** pass it explicitly.

## The fix

One line: use `Path(os.path.abspath(p))` rather than `Path(p).resolve()` for the caller-supplied paths, so the link itself is what gets tested. `os.path.abspath` normalises without following links.

Write the failing fixture first — the current suite is green against this, so a fix graded by the existing tests cannot be told apart from no fix.

## Related prose defect

The comment at `src/no_human/vcs/git.py` around the `lexists` check says it keeps "a broken symlink the coder created" from being dropped. It does that only for entries re-entering via the untracked `_CODE_EXTS` producer, never for the entry the caller passed. Accurate wording would name the producer.

L4XB:
## What this changes

`commit_paths` resolved caller-supplied paths with `Path(p).resolve()`, which FOLLOWS a symlink. A broken symlink was therefore rewritten to its missing target before the `os.path.lexists` filter saw it: the filter asked whether the *target* exists, answered no, dropped the entry, and the commit landed without the file, silently. Unresolved, the same input made `git add` exit 128 — loud, and the direction that filter exists to move away from, but the silent version loses the work instead.

The fix resolves the DIRECTORY chain and keeps the entry's own name. That still normalises a symlinked root (macOS `/tmp` → `/private/tmp`), which is what `relative_to` needs to decide repo membership, and a broken symlink reaches `git add` as itself.

Fixes #322

## Evidence

```
$ PYTHONPATH=src pytest tests/test_git_commit_paths.py -q
19 passed, 1 warning in 8.04s
```

Mutants, to show the tests fail on the unfixed code and that each one grades what it claims:

```
--- resolve() the entry again (the pre-fix line): 1 failed, 18 passed
     FAIL test_a_caller_supplied_broken_symlink_is_staged_under_its_own_name
--- purely lexical, parent never resolved:        1 failed, 18 passed
     FAIL test_a_path_reached_through_a_symlinked_repo_root_is_still_inside_the_repo
--- anything outside the repo is accepted:        1 failed, 18 passed
     FAIL test_a_path_outside_the_repo_is_still_skipped
--- FIXED (control):                              19 passed
```

Two of those tests were vacuous before I tightened them, and both traps are worth naming because they are properties of this function rather than of my tests:

- **A code extension hides the bug.** `test_a_broken_symlink_is_not_mistaken_for_a_phantom_path` uses `link.py`, which the untracked `_CODE_EXTS` producer re-adds regardless — so it passes on the unfixed code. The new tests use `.md`, where the caller's own path is the only thing that can stage the file. This is the same reason the issue calls the defect narrow.
- **The `commit_all` fallback hides it too.** With only one path and that path skipped, `rel_paths` is empty, the "none of the paths carry changes" fallback fires, and everything gets committed anyway. The symlinked-root test therefore also passes an ordinarily-addressed file, so the fallback cannot rescue the one under test.

`ruff check` reports the same 29 findings on these two files before and after (all pre-existing; the repo has no ruff config).

## Checklist
- [x] Test count and assertion count did not go down — three tests added, none changed.
- [x] Lockfiles committed if a dependency changed — no dependency changed.
- [x] Suite run locally; command + output tail pasted under Evidence. (Scoped to the owning test file: the wider suite needs project dependencies I did not install.)
- [x] This does not conflict with the project's standing constraints — no merge is performed here, no test is weakened, and no credential is touched.

## Note on the prose defect in the issue

The issue also flags the comment at the `lexists` check, which claims it keeps "a broken symlink the coder created" from being dropped — true only for entries re-entering through the untracked producer, not for the entry the caller passed. With this change that sentence is simply accurate for both, so I left it as it stands rather than rewording a comment that has become correct.


eyalgolan:
Thanks for this — the CLA gate passes and the change itself looks like the right shape. Before you chase the red CI, I owe you a correction: **one of the four failures is mine, not yours.**

```
```

That one is a pre-existing break on `main` — it has nothing to do with your change, it is failing on every PR opened right now, and a fix is in independent review. Please ignore it. If it is still red after you push, that is on me; rebase once `main` is green again and it will clear itself.

The other three are yours, and they are all the same one thing — a mechanical re-anchor, not a design problem:

```
FAILED tests/test_readme_claims.py::test_doc_citations_resolve_to_the_code_they_describe[security.md:vcs/git.py:GitRepo._have_remote_commit:1177]
FAILED tests/test_readme_claims.py::test_doc_citations_resolve_to_the_code_they_describe[security.md::GitRepo.fetch:1513]
FAILED tests/test_reanchor_citations.py::test_check_mode_is_clean_on_this_tree
```

`docs/security.md` cites `vcs/git.py` by line number, and your change shifted those lines by 8. The symbols still resolve, so the rewrite is exact and the tool does it for you:

```
```

Then regenerate the file inventory, which is failing for the same underlying reason (your edit changed file contents and `RELEASE_MANIFEST.txt` pins a hash per file):

```
```

Commit both alongside your change. Order matters here and it bites people: do **all** source edits first, then re-anchor, then regenerate the manifest. Re-anchoring rewrites `docs/*.md`, which changes their hashes, so a manifest generated before the re-anchor is stale the moment it lands.

One gotcha that has cost this repo real time: if `reanchor_citations.py` reports drift you believe you already fixed, delete `__pycache__` and re-run. It loads modules by file path and Python validates its bytecode cache on (mtime-seconds, size), so a rewrite that happens to preserve byte length within the same second can re-report stale line numbers.

I have not reviewed the substance of the symlink fix yet — I will come back to that separately once `main` is green, so you are not debugging against a moving base.

L4XB:
Done in `9b6c594`, in the order you gave — source edit, then re-anchor, then manifest.

`scripts/reanchor_citations.py --apply` moved both `security.md` citations by the 8 lines the change shifted (`_have_remote_commit` 1177 → 1185, `fetch` 1513 → 1521) and reported `VERDICT=OK`; the symbols were unchanged, so the rewrite is exact. `check_release_manifest.py --write` ran afterwards and re-recorded 1595 rows — including the two `docs/*.md` hashes the re-anchor had just changed, which is the trap you flagged.

```
pytest tests/test_readme_claims.py tests/test_reanchor_citations.py
140 passed, 12 skipped
```

`test_an_unreadable_venv_pyvenv_cfg_still_denies_the_install` passes locally on this branch (59 passed in that file), so if it is still red in CI it is the `main` break you described and I will leave it alone — thanks for saying so up front, I would otherwise have spent the time on it.

eyalgolan (changes requested):
@L4XB — the fix is right and the reasoning behind it is the part I'd highlight. One mechanical blocker.

## The fix, verified

I ran your test file against `main`'s source to check it actually pins the bug rather than just passing alongside it:

```
PR's tests vs main's source:  1 failed, 18 passed
  FAILED test_a_caller_supplied_broken_symlink_is_staged_under_its_own_name
PR's tests vs PR's source:    19 passed
```

Load-bearing. And the comment explains the failure mode better than the issue did:

> `Path(p).resolve()` follows a symlink, so a broken one is rewritten to its missing target before the `os.path.lexists` filter below ever sees it: that filter then asks "does the target exist", answers no, drops the entry, and the commit lands without the link — the silent-work-loss direction, where an unresolved name would have made `git add` fail loudly instead.

Resolving the parent chain but not the entry keeps the `/tmp` → `/private/tmp` normalisation that `relative_to` needs, which is the detail a naive "stop resolving" fix would have broken. Silent work loss is the worst failure direction we have, so this is worth having.

## Blocker: the manifest

```
  src/no_human/vcs/git.py: content differs from the manifest
  tests/test_git_commit_paths.py: content differs from the manifest
```

`RELEASE_MANIFEST.txt` pins every tracked file by SHA-256, so any change to one has to re-pin it in the same commit, and **File inventory** will be red until it does:

```bash
git add -A
```

`--strict` should then print `OK`. If you merge `main` at the same time, take its manifest wholesale rather than merging it — `git checkout origin/main -- RELEASE_MANIFEST.txt` then `--write` — because git resolves a generated file cleanly and wrongly, which has bitten two other contributors this week. That's #314, and it's our tooling problem rather than yours.

Your branch is also `BEHIND` main, so rebase or merge while you're there.

Fix the manifest and I'll approve — the code side is done.

eyalgolan (approved):
Approving. Manifest is clean now (`OK: 1607 file(s)`), `main` is an ancestor, and the full suite is green:

```
12742 passed, 225 skipped, 2 xfailed, 0 failed
```

## The test earns its place

I ran your test file against **`main`'s source** rather than only against yours, to check it pins the bug instead of passing alongside it:

```
PR's tests vs main's source:  1 failed, 18 passed
  FAILED test_a_caller_supplied_broken_symlink_is_staged_under_its_own_name
PR's tests vs PR's source:    19 passed
```

Re-ran that after your latest push and it still holds.

## Why the fix is the right shape

The comment is doing real work:

> `Path(p).resolve()` follows a symlink, so a broken one is rewritten to its missing target before the `os.path.lexists` filter below ever sees it: that filter then asks "does the target exist", answers no, drops the entry, and the commit lands without the link — the silent-work-loss direction, where an unresolved name would have made `git add` fail loudly instead.

Two things I want to call out:

**The failure direction is named.** Silent work loss is the worst outcome this codebase has, and a loud `git add` failure is strictly better than a commit that quietly omits a file. Saying so in the comment is what will stop someone "simplifying" this back in six months.

**Resolving the parent but not the entry is the non-obvious half.** A naive "stop resolving" fix would have broken the macOS `/tmp` → `/private/tmp` normalisation that `relative_to` depends on. `abs_p.parent.resolve() / abs_p.name` keeps it. That distinction is easy to miss and expensive to rediscover.

PR #339, opened 2026-09-13
fix(guard): measure case folding instead of assuming it

### LINKED ISSUE #328: Case-folding in the guard: #320 landed the opposite answer to #307, and four gates are open on a case-insensitive host

L4XB:
Fixes #328.

Took the suggested resolution: measure, rather than pick a winner between #307 and #320. Both docstrings are true once the host is asked.

```python
def host_folds_case() -> bool:
    """Whether this host's filesystem resolves two spellings of one name to the
    same file. Measured, not assumed."""
    return _folds_case(Path(__file__).resolve())
```

Swap the case of a filename, ask whether that path is the same file. `samefile`, not `exists`, so a genuinely different file carrying the swapped spelling is not mistaken for a fold. Cached, since the answer cannot change while the process runs and it is read on every guarded command. A path that cannot be stat'ed proves nothing, so it falls back to the host class rather than the permissive answer.

## The name half was not the whole gap

Closing `command_name` alone leaves two of your four rows open — I checked by running them:

| after the name fix alone | |
|---|---|
| `GH pr merge 7` | DENY ✅ |
| `FIND / -delete` | DENY ✅ |
| `RM -rf /` | **ALLOW** |
| `sh -c "GIT push origin main"` | **ALLOW** |

Those two never reach name resolution. `_RM_RF` and `_GIT_DESTRUCTIVE` match the raw command text, and `_looks_like_git_push` — the gate that decides whether `_git_push_invocations` recurses into a quoted payload at all — is `re.search(r"\bgit\b.*\bpush\b", text, re.DOTALL)`. All three compiled without `IGNORECASE`, so `GIT push` never got past the recursion gate and the name fix inside it never ran. They now take `exec_names.case_flags()`, which is the same measurement expressed as a regex flag.

`rm -RF /` is a fifth row in the same family — the flag spelling, which `[a-z]*` misses for the same reason.

One more site while I was there: `_strip_wrappers` resolved a wrapper's target with `PurePosixPath(tok).name`, which folds nothing *and* strips no suffix, so `env RM -rf /` and `env gh.exe pr merge 7` both walked past it. It calls `command_name` now, like every other site.

## guard.py's budget

2926 → **2925**. No line added: `exec_names` was already imported at line 52, the flag argument fits on the existing compile lines, and the `_strip_wrappers` change is a one-for-one substitution. The ratchet only moves down.

## Verification

```
pytest tests/test_guard.py tests/test_exec_names.py tests/test_structural_budget.py
351 passed
```

The existing POSIX table now pins `fold_case=False` explicitly. That is not cosmetic: those rows are the case-sensitive host's contract, and left on the probe they would flip on any developer machine whose own filesystem folds — reading as a regression in the very design decision they exist to record.

New cells: a folding host reads one name; a folding host denies all six rows; a case-sensitive host denies none of them; the probe agrees with an independently derived measurement; an unmeasurable path falls back to the host class.

Mutation-checked:

| mutation | test that fails |
|---|---|
| name fold back to `is_windows` only | the folding-host row test |
| `_RM_RF` compiled without the flag | same |
| the recursion gate compiled without the flag | same |
| the probe hardcoded to "folds" | the unmeasurable-path fallback |

One honest limit, stated in the test's own docstring: on a case-**insensitive** developer machine a probe hardcoded to `True` agrees with every measurement of the real volume and can only be caught by the fallback cell above. On a case-sensitive runner — which CI is — `test_the_probe_matches_this_host` fails it directly. A host that folds cannot observe the difference between measuring and assuming that it folds.

## On #307

Its wrapper-operand half (the `env --unset NH nh approve 7` family) is independent of this argument and not included here — this change is only the case-fold half, so that work stays mergeable on top rather than being silently half-absorbed.


L4XB:
CI follow-up, `6d6a092`: the probe used `pathlib.Path`, which instantiates the class for the *current* platform. `test_the_platform_constants_are_computed_from_os_name` patches `os.name` to `"nt"` and reloads `guard`, so the probe — which now runs at import time via `case_flags()` — raised `NotImplementedError: cannot instantiate 'WindowsPath'` before the test could assert anything, and took the venv-guard test down with it in the same session.

It asks the same two syscalls through `os.path` now, with no platform-bound object in the way. 503 passed across `test_guard.py`, `test_exec_names.py`, `test_structural_budget.py`, `test_venv_install_guard.py` and `test_windows_command_readings.py`; manifest re-recorded.

eyalgolan:
Your CI is red for a reason that is entirely mine, and it is now fixed on `main`.

The only failure on this PR is:

```
```

That was a pre-existing break on `main` — a genuine fail-open in the venv install guard that happened to pass on macOS and fail on the Ubuntu runner, because the test's fixture inherits `/usr/bin` from the host and `/usr/bin/pip` exists on the runner but not on my dev machine. It has nothing to do with your change. It landed fixed as `1d7c28a6` and every CI job passes on that commit.

**Rebase onto current `main` and this PR should go green with no changes from you.** Nothing else in your diff failed.

Apologies for the noise — you opened this during the window where every PR in the repo was inheriting my breakage, and there was no way for you to tell that from the outside.

I have not reviewed the substance yet; I will come back to that separately now that you are not debugging against a red base.

eyalgolan (approved):
Approving. This settles #328 by asking the filesystem instead of asking me to pick between two positions that were each right about their own host — which is the better outcome by some distance.

## Verified

`main` (`1d7c28a6`) is an ancestor, so what I checked is what lands.

```
full suite      12745 passed, 225 skipped, 2 xfailed, 0 failed
--strict alone  OK: 1607 file(s)
guard.py        2925  (budget 2926 — this SHRINKS the file)
guard suites    429 passed
```

Every cell from #328, on the default reading this host actually runs:

```
DENY  GH pr merge 7        DENY  RM -rf /      DENY  FIND / -delete
DENY  GIT push origin main DENY  rm -RF /      DENY  sh -c "GIT push origin main"
```

Lowercase twins still deny; `git status`, `gh pr view 7`, `grep -r foo src/` all still allow.

## What I checked hardest, and what held

**The probe's fallbacks.** A probe that guesses permissively when it cannot tell would be worse than the assumption it replaces. Measured:

```
no letters to swap  ->  os.name == "nt"     (False here)
missing path        ->  os.name == "nt"     (False here)
OSError             ->  os.name == "nt"
```

Always toward the host class, never toward "allow". `samefile` rather than `exists` is the right call too — a genuinely different file carrying the swapped spelling is not mistaken for a fold.

**It covers the half the name resolver cannot reach.** `rm -RF /` and `sh -c "GIT push origin main"` close because `case_flags()` reaches `_RM_RF`, `_GIT_DESTRUCTIVE` and the `_looks_like_git_push` recursion gate. A fix that only folded argv[0] would have left those open and looked complete.

**Both readings stay testable from one host** — `command_name(..., fold_case=...)` is injectable and the tests patch `host_folds_case`. That property is why this gap survived in the first place, so keeping it matters more than usual.

**`os.path` rather than `pathlib`**, with the reason in the comment: `Path(...)` instantiates for the current platform, so a test patching `os.name` to `"nt"` would get `NotImplementedError` from the probe instead of an answer. That is an easy detail to lose in a later cleanup and it is written down.

## One residual, not a blocker

The probe measures the case-sensitivity of **where this module lives**, not of the directory the command will run in. Those can differ, and one direction fails open:

- package on a case-sensitive volume, worktree on a case-insensitive one → answers "don't fold" → the capitalised spellings reopen;
- the reverse → false denials, which is fail-closed and fine.

That is not hypothetical here: worktrees live under `~/.no_human/worktrees/` while the installed package may sit elsewhere. Worth a follow-up that probes the command's `cwd`, or probes both and folds if **either** folds. It is strictly better than `main` either way, so I would rather land this and file that than hold it.

eyalgolan:
Small follow-up on the residual I mentioned: I went and measured it rather than leaving it as a hunch, and it reproduces — filed as #345 with the numbers.

Built a case-sensitive APFS image next to the normal checkout (positive control first: `PROBE.TXT` genuinely doesn't resolve there). Same `guard.evaluate`, same `cwd`, same commands; only the volume the package was imported from differs:

```
package on case-INsensitive vol   GH pr merge 7 -> DENY
package on case-SENSITIVE   vol   GH pr merge 7 -> ALLOW
```

Same flip for `RM -rf /` and `GIT push origin main`.

To be clear this does **not** change my approval — on a normal install your PR turns all three from ALLOW to DENY, and the mixed-volume case was equally open before you touched it. It's a follow-up, not a debt you created. The nice part is `evaluate()` already receives `cwd` and documents it as where the command actually runs, so the fix is mostly asking the probe the question the guard can already answer. Happy for you to take #345 if you want it, no obligation.

eyalgolan:
Reviewed. **The idea is right and the direction is right — but the mechanism does not survive the way this product actually ships, and it fails in the open direction.** Details below, because this is a genuinely subtle one and you could not reasonably have seen it from the source tree.

First, what holds up, so none of it gets reworked: all gates are green, and the regression direction is strictly safe. Over a 42-row corpus run through the real `guard.evaluate` on both trees, **10 rows flipped ALLOW→DENY and 0 flipped DENY→ALLOW**, with 32 control rows agreeing. Unicode folding is deny-direction only (`.lower()` and `re.IGNORECASE` fold *more* aggressively than APFS, e.g. Kelvin sign U+212A → `k`, never less). The probe does no filesystem write and has no meaningful TOCTOU window (`lru_cache`d, first called at import). A planted decoy `EXEC_NAMES.PY` cannot force the "case-sensitive" answer — `samefile` correctly rejects it, verified both ways. And a full run on a real case-sensitive APFS volume is clean. Your `samefile`-not-`exists` design decision is right.

## The blocker

`exec_names.py:147` measures the filesystem by case-swapping **its own source file's name**:

```python
return _folds_case(os.path.realpath(__file__))
```

The shipped desktop server is a PyInstaller **onedir bytecode freeze whose first stated rule is "NO .py FILES IN THE BUNDLE"** (`packaging/nh-server.spec:6`). So at runtime `__file__` names a path that does not exist, `os.path.exists(swapped)` is False, and `_folds_case` returns `os.name == "nt"` → **False on macOS and Linux**.

Built as a real PyInstaller onedir freeze of `guard` + `exec_names` at your head, run on macOS:

```
FROZEN: True
exec_names.__file__: .../_internal/no_human/agent/exec_names.py   exists: False
host_folds_case(): False

  'RM -rf /'                     -> ALLOW
  'GH pr merge 7'                -> ALLOW
  'FIND / -delete'               -> ALLOW
  'sh -c "GIT push origin main"' -> ALLOW
  'rm -rf /'                     -> DENY    <- control
  'gh pr merge 7'                -> DENY    <- control
```

Those first four rows are exactly the four the PR exists to close. The two lowercase controls prove the guard is otherwise working. And the bypass is live, not theoretical — on this machine:

```
$ touch victim.txt && RM -f victim.txt
  RM DID run (file gone)
```

`GH pr merge 7` matters most: `guard.evaluate` is the real PreToolUse veto (`agent/claude_backend.py:470`), and "the agent never merges" is a non-negotiable project constraint.

So: **from source, every test passes and the change reads as closing the class; in the binary users run, it closes nothing.** That is arguably worse than the assumption it replaces, because `is_windows` at least behaved identically in both.

## Second blocker: the fallback IS the permissive answer

```python
except OSError:
    # An unreadable or vanished path proves nothing; fall back to the
    # host class rather than guessing the permissive answer.
    return os.name == "nt"
```

On macOS — the platform this PR was written for — `os.name == "nt"` is False, i.e. *don't fold*, i.e. the permissive answer. The comment asserts the opposite of what the code does.

Two things underneath that:
- That `except OSError` branch is nearly dead: `os.path.exists` swallows `OSError` and returns False, so an unreadable path exits through the **undocumented** `exists()==False` path — same permissive value, no comment, no guard. Mutating the `except` body to `return True` leaves 489 passing.
- `test_an_unmeasurable_path_falls_back_to_the_host_class` asserts `is (os.name == "nt")` and its docstring says "rather than guessing the permissive one". On POSIX that assertion *is* the permissive one, so the suite currently certifies the wrong behaviour.

An unmeasurable probe should **fold** (fail closed), not fall back to `os.name`.

## Third: the probe measures the wrong filesystem

The measurement is taken on the volume holding the *source*, not the volume the *command* runs against. On a real case-sensitive APFS volume holding the package, with cwd on the case-insensitive repo, all four rows reopen. Developers who keep code on a case-sensitive volume are exactly the people who care about this bug. Note too that command resolution is a **PATH** question — `/usr/bin` may fold when `/Volumes/dev` does not.

`evaluate` already receives the `cwd` the command will run in. That is a path that exists at runtime and is on the right volume.

## Smaller things

- **Two of your three source changes are pinned by no test.** Reverting `guard.py:2244` to `PurePosixPath(tok).name` leaves 489 passing; compiling `_GIT_DESTRUCTIVE` without the flag leaves 426 passing, 0 failing — and `_GIT_DESTRUCTIVE` is absent from your own mutation table while `GIT clean -fd` and `GIT reset --hard HEAD~1` are two of the ten rows that flipped. Dropping `samefile` also survives, so your headline design decision is unguarded.
- **The `_strip_wrappers` claim does not hold.** The changed line sits inside `if saw_wrapper and argv and argv[0].startswith("-")`, and instrumenting shows `env RM -rf /` and `env gh.exe pr merge 7` never enter that loop (`argv0='RM'` / `'gh.exe'`, loop entered = False). "like every other site" is also not yet true — six sites still use `PurePosixPath(...).name` (`guard.py:309, 1450, 1843, 1844, 2536, 2574`), including the two direct siblings.
- **Prose to update:** `exec_names.py:84` still says "Pure string work, no filesystem access" (the PR adds `realpath`/`exists`/`samefile`); `:54` still states the POSIX position the PR overturns; `:80-82` calls widening `_RM_RF` "a separate change" while this PR widens it. Measured at runtime, 6 of the 7 patterns that docstring names are still case-blind.
- `tests/test_structural_budget.py:1827` still reads `2926` while `guard.py` is now 2925 — it passes because the ratchet only flags growth, but it leaves unclaimed slack.

## Worth knowing

Your PR body's honesty claim is **true** — on a genuine case-sensitive volume, `test_the_probe_matches_this_host` does fail directly. My reviewer initially got a contrary result, traced it to `pyproject.toml`'s `pythonpath = ["src"]` resolving the module from the wrong tree, and retracted it. Your statement stands.

## What would make this landable

Measure a path that exists at runtime — the `cwd` `evaluate` already gets — or make an unmeasurable probe fold rather than fall back to `os.name`. Then pin `_GIT_DESTRUCTIVE`, the `samefile` choice, and the `_strip_wrappers` change with mutations, and reconcile the four prose statements.

One more thing you may find motivating: the sibling `venv_install_guard._is_installer_name` still folds "on Windows ONLY", and on this host `pip install requests` DENIES while `PIP install requests` ALLOWS — a live bypass using exactly the reasoning your PR rejects. Not yours to fix here, but it shows the thesis is right and under-applied.

PR #342, opened 2026-09-13
fix(desktop): decide the Windows artifact tag from Windows inputs

### LINKED ISSUE #330: The Windows exe's -UNSIGNED tag is decided by the Apple identity variables
`desktop/signing.cjs:106` computes the signing plan from the **Apple** identity
variables only:

```js
const hasIdentity = present(env, "CSC_LINK") || present(env, "CSC_NAME");
```

That plan supplies `plan.artifactTag`, which
`desktop/electron-builder.config.cjs:134` interpolates into
`artifactName: "${productName}-${version}" + plan.artifactTag + ".${ext}"` for
**every** target, Windows included (`:123` `target: ["nsis", "zip"]`).

There is no Windows-specific check anywhere in that decision. So the Windows
artifact's name is honest by accident, not by construction: the windows job
happens not to set `CSC_LINK`/`CSC_NAME`, so the tag resolves to `-UNSIGNED`.
If that job ever inherits those secrets — a shared org secret, a workflow edit,
a matrix that exports them once for all platforms — the exe is named
`no_human-<v>.exe` with no tag while remaining unsigned, because no Windows
certificate exists for them to point at. The variable that flips the name has
nothing to do with the thing the name is claiming.

Measured on the Windows build machine (Windows 11, by the session that owns
that lane), each check with a positive control:

```
Get-ChildItem Cert:\CurrentUser\My -CodeSigningCert   -> 0   (control: 13 certs in the same store)
Get-ChildItem Cert:\LocalMachine\My -CodeSigningCert  -> 0
CSC_LINK / CSC_NAME / CSC_KEY_PASSWORD / WIN_CSC_LINK / WIN_CSC_KEY_PASSWORD  -> all unset (control: PATH set)
no *.pfx / *.p12 under the home directory
Get-AuthenticodeSignature <last local build>.exe      -> NotSigned  (control: notepad.exe -> Valid)
```

So the absence is genuine, unlike the macOS case where the certificate was in
the keychain all along and only an env var was unset.

**A second, related trap for whoever reads a build log.** An UNSIGNED Windows
build still prints `signing with signtool.exe path=...` for every exe. That is
electron-builder's step name, not a signature — `Get-AuthenticodeSignature`
returns `NotSigned` for the artifact it just printed that line about. A reader
checking "did it sign?" by grepping the log gets the wrong answer.

**Acceptance (properties, not mechanisms):**

1. The artifact-name tag for a given platform is derived from whether THAT
   platform's signing inputs are present, so setting an Apple identity
   variable cannot change what a Windows filename claims.
2. A build with no usable certificate for the target platform produces an
   artifact whose name says so, and that property holds under a test that
   sets the other platform's identity variables.
3. The build's own output does not state that it signed something it did not;
   or, if the wording belongs to a vendor tool and cannot be changed, the
   place a reader is directed to check is the artifact's signature, not the
   log line.

Found while preparing the v0.2.3 three-platform release. Not shipped-broken
today: the windows job sets neither variable, so the current names are
correct.

L4XB:
Fixes #330.

## What changed

`windowsSigningPlan(env)` answers the Windows question from `WIN_CSC_LINK`, and `win.artifactName` interpolates its tag instead of the macOS plan's. Setting an Apple identity variable can no longer change what a Windows filename claims — which is acceptance criterion 1, and the test for criterion 2 sets `CSC_LINK`, `CSC_NAME`, `CSC_KEY_PASSWORD` and the full notary set at once and asserts the exe is still tagged.

## Two decisions worth disagreeing with explicitly

**`CSC_LINK` is not read on the Windows side either.** electron-builder does fall back to it for Windows, so a build signed that way is now tagged `-UNSIGNED` when it is in fact signed. That under-claims, and I took it deliberately: a name that under-claims is corrected by setting `WIN_CSC_LINK`, while a name that over-claims is uploaded to a release page. If you would rather the generic fallback counted, it is one `present(env, "CSC_LINK")` — but then an Apple `.p12` in a shared secret re-opens exactly this issue, since nothing at that point distinguishes the two certificates.

**A separate function, not a branch inside `signingPlan`.** The macOS plan also decides notarization, the auto-update permission and electron-builder's `identity` value, none of which exist on Windows; folding a second platform into it would make every field conditional on which platform is asking. The two now share only a filename convention, which is what they actually have in common.

## Criterion 3, the log line

electron-builder prints `signing with signtool.exe path=...` for every exe even when nothing is signed. That is its step name, not a result, and it is vendor output I cannot change — so the Windows banner says where the real answer is:

```
WINDOWS SIGNING: UNSIGNED — NOT SHIPPABLE
no WIN_CSC_LINK — this Windows build is UNSIGNED. SmartScreen will warn on it and auto-update is not offered.
The .exe will be tagged "-UNSIGNED" so it cannot be mistaken for a release.
The build log's "signing with signtool.exe" line is a step name, not a signature — check the artifact with Get-AuthenticodeSignature.
```

## Not touched

`plan.fatal` (`NH_REQUIRE_SIGNED`) still comes from the macOS plan alone. On a Windows job with that variable set the build already refuses today, so changing it is a behaviour change in its own right rather than part of this fix.

## Verification

```
node --test desktop/*.test.mjs
431 pass, 3 fail
```

The three failures are the same three that fail on a clean checkout of `main` — `test_fallback_ui_identical_to_current_state`, `desktop/uiPages.test.mjs`, and `the config exports the config and nothing else` — verified by diffing the failing-test names before and after the change; only the timings differ. `signing.test.mjs` itself is 19/19.

Six new cells, mutation-checked:

| mutation | tests that fail |
|---|---|
| exe name back on `plan.artifactTag` | the wiring cell |
| `CSC_NAME` counted as a Windows input | the Apple-identity cell |
| the banner drops the where-to-check line | the banner cell |

The wiring cell reads the config's `win` block as text and asserts it interpolates `winPlan.artifactTag` and not `plan.artifactTag`. Nothing else can see that line without a real three-platform build, and it is the one line that makes the fix live — the plan can be perfect while the config still names the exe from the macOS tag, which is the state this issue reports.

`RELEASE_MANIFEST.txt` regenerated with `scripts/check_release_manifest.py --write`.


eyalgolan:
Thanks for this — it targets a real issue (#330), and the user-visible half of that bug was confirmed on a real Windows machine earlier today: with `nhCanAutoUpdate=false` the app correctly *detects* an update and then hands off with "This build is not code-signed, so it cannot install updates itself. Download the new version manually." The bug removes exactly that copy, routing a Windows user into a self-install branch that has no Windows certificate behind it. So the thing you are fixing matters more than the tag name suggests.

Before you chase the red CI: **one of the three failures is mine, not yours.**

```
```

That was a pre-existing break on `main` (a real fail-open in the venv guard that passed on macOS and failed on the Ubuntu runner). It is fixed as `1d7c28a6`. Rebase and it clears itself.

The other two are yours, and they are one mechanical thing:

```
FAILED tests/test_readme_claims.py::test_every_line_citation_currently_resolves_exactly
FAILED tests/test_reanchor_citations.py::test_check_mode_is_clean_on_this_tree
```

A doc cites source by line number, and your edit shifted the lines. The symbols still resolve, so the rewrite is exact and the tool does it:

```
```

Order matters and it bites people: **all source edits first, then re-anchor, then regenerate the manifest.** Re-anchoring rewrites `docs/*.md`, which changes their hashes, so a manifest generated before the re-anchor is stale the moment it lands.

Two gotchas worth knowing, both of which cost me real time today:

- If `reanchor_citations.py` reports drift you believe you already fixed, delete `__pycache__` and re-run. It loads modules by file path and Python validates its bytecode cache on (mtime-seconds, size), so a same-length rewrite inside one second re-reports stale numbers.
- `docs/WINDOWS.md` is **not** covered by the re-anchor script. If your change shifts a line cited from there, the tool will report VERDICT=OK and `test_windows_md_code_line_citations_resolve` will still fail. That one has to be fixed by hand, and the citation should point at the executable line, not a comment that happens to contain the same token.

I have not reviewed the substance yet. When I do, the thing I will look hardest at is whether the tag is now decided by inputs that are genuinely Windows-side — the failure mode of the original bug is that a *Mac release operator's shell* silently decided what a Windows artifact claimed about itself.

eyalgolan (dismissed):
Approving. The bug is worse than #330 describes, and your fix is pointed at the right thing.

## What main actually does

I ran the macOS plan on `main` (`1d7c28a6`) against the inputs the Windows config feeds it, since `electron-builder.config.cjs:134` uses `plan.artifactTag` for the Windows `artifactName`:

```
ON MAIN -- what names the Windows .exe
  no vars                     -UNSIGNED
  CSC_LINK set (macOS cert)   -UNNOTARIZED     <- a macOS-only concept, on a Windows exe
  CSC_NAME set (Apple id)     -UNNOTARIZED
  WIN_CSC_LINK set            -UNSIGNED        <- genuinely signed, labelled unsigned
```

So it fails in both directions: setting a **macOS** certificate renames the Windows artifact to a notarization state Windows does not have, and setting the **actual** Windows certificate does not clear the `-UNSIGNED` tag at all. An asset that goes to a release page claiming a signature state it cannot have is the part that makes this more than cosmetic.

## Your branch

```
  no vars                                  -UNSIGNED
  CSC_LINK set (macOS cert)                -UNSIGNED    <- no longer flipped by a mac var
  CSC_NAME set (Apple identity)            -UNSIGNED
  WIN_CSC_LINK set (the real Windows cert) ""           <- signed
  both mac + win                           ""
```

macOS plan unchanged (`-UNSIGNED` / `-UNNOTARIZED` / `-UNNOTARIZED`), so the split did not disturb the lane that was already right.

Verified on the merge result: `--strict OK: 1607 file(s)`, `node --test signing.test.mjs` → **19 pass, 0 fail**.

## What makes it a good change

Two things I would keep in any rewrite of this file:

**The fail-safe direction is stated and correct.** A build signed through electron-builder's generic `CSC_LINK` fallback gets tagged `-UNSIGNED` — it under-claims. Your comment says why that is the right way round: *"A name that under-claims can be corrected by setting `WIN_CSC_LINK`; a name that over-claims is uploaded to a release page."* That is the correct asymmetry for an artifact name, and it is the kind of reasoning that is usually missing from a rename fix.

**`CSC_NAME` is excluded with a reason rather than by omission** — it is an Apple identity name, so there is no Windows certificate for it to point at. Someone will eventually wonder why it is not in the list, and the answer is right there.

Separating `windowsSigningPlan` from `signingPlan` rather than branching inside it is also right: the macOS plan additionally decides notarization, auto-update permission and the `identity` value, none of which exist on Windows. A shared function would have grown a platform flag and then grown bugs behind it.

## One note, not a blocker

This is signing-adjacent, so it deserves saying explicitly: the change alters **naming only**, never whether anything is actually signed. `WIN_CSC_LINK` is read to decide a filename suffix and nothing else, and no credential is read, logged or passed anywhere new. I checked that before approving rather than assuming it from the diff size.

eyalgolan:
Reviewed, and independently a second time on a real Windows machine. Both reviews reached the same conclusion without seeing each other: **this fixes the filename half of #330 and leaves the half with the user consequence in place.** That is a real contribution and the direction is right — but it should not land as-is, because in its current form it makes the artifact's *name* and the artifact's *behaviour* disagree.

## The core finding

`desktop/electron-builder.config.cjs` still has, at top level:

```js
extraMetadata: {
  nhSigning: plan.mode,
  nhCanAutoUpdate: plan.canAutoUpdate,
},
```

`plan` is `signingPlan(process.env)` — the **macOS** plan, unchanged by this PR. Loading the real config under exactly the scenario #330 describes (placeholder values, no real credentials):

```
$ env -u WIN_CSC_LINK CSC_NAME="PLACEHOLDER-NOT-A-REAL-IDENTITY" \
      APPLE_API_KEY=x APPLE_API_KEY_ID=x APPLE_API_ISSUER=x node probe.cjs
PROBE win.artifactName = "${productName}-${version}-UNSIGNED.${ext}"
PROBE extraMetadata    = {"nhSigning":"signed","nhCanAutoUpdate":true}
```

Those two lines are the bug. The exe is now correctly *named* `-UNSIGNED`, but the app inside it is still baked `nhCanAutoUpdate: true`. `main.mjs` `packagedSigning()` reads those fields, `updater.mjs` takes the `AVAILABLE` branch instead of `UNAVAILABLE`, and the manual-download card in `updatePolicy.mjs` — *"this build is not code-signed, so it cannot update itself. Download the new version manually."* — is never rendered. That card was observed working on a real 0.2.1 Windows install; it is the thing a Windows user actually relies on, and it still disappears when a Mac operator has the Apple variables exported.

The three-way comparison makes the new contradiction clear:

```
clean env:        name -UNSIGNED    metadata unsigned/false    agree
Apple vars set:   name -UNSIGNED    metadata signed/true       DISAGREE
WIN_CSC_LINK set: name untagged     metadata unsigned/false    DISAGREE
```

Before this PR both came from one plan and always agreed — on a wrong answer, but they agreed. Now an operator who checks the filename (which is exactly what the filename convention exists for) reads `-UNSIGNED`, concludes Windows is handled, and ships an app that auto-updates anyway.

## Why it is not a one-line change

`extraMetadata` is top-level only in electron-builder — checked against app-builder-lib's `scheme.json`: it exists on the root and on no per-platform definition, so `win: { extraMetadata }` is not available. The workable shape is to bake both verdicts and choose at runtime: add `nhWinSigned` / `nhWinCanAutoUpdate` from the Windows plan alongside the existing fields, and have `packagedSigning()` return the Windows pair when `process.platform === "win32"`, keeping the strict `=== true` so it still fails closed.

## The wiring test cannot see the bug it was written for

`desktop/signing.test.mjs` asserts on the config as **text** (`/winPlan\.artifactTag/`). I aliased `const winPlan = windowsSigningPlan(process.env)` back to `signingPlan(process.env)` — restoring issue #330 verbatim:

```
PROBE win.artifactName = "${productName}-${version}.${ext}"     <-- the bug, restored
# tests 19 # pass 19 # fail 0                                    <-- suite still green
```

A source-text guard passes as long as the string is present, regardless of what the code does. Three other mutations also survived: changing the banner headline from "UNSIGNED — NOT SHIPPABLE" to "fine, ship it" (test 18 matches `/UNSIGNED/` against the *reason* line, not the head it is named for); deleting the `console.log(windowsSigningBanner(...))` call entirely (nothing checks the banner is ever printed); and renaming the `WIN_CSC_LINK` constant (the test writes `[WINDOWS_CERTIFICATE_VAR]`, so the assertion follows the rename and the variable an operator must actually export is unpinned).

The fix is to require the built config **object** and assert on `cfg.win.artifactName` *and* `cfg.extraMetadata` under a set Apple environment — the alias mutation must go red.

## Smaller things

- The new banner says *"auto-update is not offered"*, which the same process then refutes six lines later by printing `nhCanAutoUpdate: true`. Once the metadata is fixed this sentence becomes true; until then it is shipped text asserting what the code does not establish.
- `desktop/signing.cjs` contradicts itself 14 lines apart: one comment says `CSC_LINK` *"cannot sign an exe"*, another says a build *"signed through electron-builder's generic `CSC_LINK` fallback"* gets tagged `-UNSIGNED`. Both ship. (For what it is worth, the second is right — electron-builder's `getCscLink("WIN_CSC_LINK")` falls back to `CSC_LINK`, and `win.signtoolOptions.certificateSubjectName` / `certificateSha1` / `certificateFile` are further routes. Under-claiming is the safe direction, so keying on `WIN_CSC_LINK` only is fine — the comment just needs to stop saying the opposite.)
- The `win` block comment directly above your edited line still asserts *"`nhCanAutoUpdate` below stays false"*, which the probe above disproves.
- `WIN_CSC_LINK` appears nowhere outside the module that reads it — not in `docs/WINDOWS.md`, not in any workflow (positive control: `CSC_NAME` appears in 5 files). The banner instructs an operator to set a variable that nothing documents.
- Presence alone clears the tag, so `WIN_CSC_LINK=/nonexistent.pfx` produces an untagged name. That over-claims, which is the opposite of the docstring's stated fail-safe direction. Worth noting the asymmetry with macOS, where an empty tag needs two independent signals *and* the release pipeline gates on `codesign`/`spctl`/`stapler`.
- The two citation failures are benign drift from your own edit (`+5` net lines, `366 → 371`, same token) — `uv run python scripts/reanchor_citations.py --apply` clears them.

## What I verified and found clean

macOS behaviour is genuinely untouched. Your diff has zero deleted lines in `signing.cjs`, and I compared `main`'s `signingPlan` against yours across six environments — output identical, exit 0. All three macOS modes stay distinguishable and the DMG pipeline is unaffected. Fail-closed on absent/empty/whitespace `WIN_CSC_LINK` also holds. The manifest hashes are correct. Full desktop gate on your head: `# tests 434 # pass 433 # fail 0 # skipped 1`.

One correction on your PR body's verification line: "431 pass, 3 fail" does not reproduce. CI's "Desktop shell" job runs every `./*.test.mjs` **except** `uiPages.test.mjs`; including it is what produces those failures. The repo's actual gate is 433/434 pass, 0 fail.

Happy to look again as soon as `extraMetadata` is platform-aware and the wiring test asserts on the object. This is a genuinely useful fix and I would like it to land.

L4XB:
CI was failing on two things, neither of them the change itself. Both handled, and your `main` merges are merged in rather than overwritten.

**Citation drift.** This branch shifts `desktop/electron-builder.config.cjs`, so `security.md`'s citation to line 366 now resolves at 371, and both citation suites failed on the drift:

```
DRIFT: security.md `desktop/electron-builder.config.cjs:366` -> `...:371` (would re-anchor)
```

`scripts/reanchor_citations.py --apply` moves the anchor and the citation table with it, then `check_release_manifest.py --write` restamps. `VERDICT=OK`; `test_readme_claims.py` + `test_reanchor_citations.py` + `test_git_commit_paths.py` are 159 passed, 12 skipped here, and `desktop/signing.test.mjs` is green.

**`test_unmask_is_one_pass_not_one_per_table_entry` is a timing flake, not a regression.** It asserts `elapsed < 0.4` and the runner measured `0.40029s` — over by 0.3 ms. It passes here three runs in a row, and this branch touches nothing under `guard`. Probably worth its own issue: a bound that fails at 100.07 % of budget will page whoever draws a slow runner.

One `RELEASE_MANIFEST.txt` conflict came up while merging; I regenerated the file rather than picking a side, since it is generated.

eyalgolan (changes requested):
I need to walk back my approval on this one, and the mistake is mine rather than yours.

CI is red on three Python tests. Two of them are real and caused by this branch:

```
FAILED tests/test_readme_claims.py::test_every_line_citation_currently_resolves_exactly
FAILED tests/test_reanchor_citations.py::test_check_mode_is_clean_on_this_tree

DRIFT: security.md `desktop/electron-builder.config.cjs:366`
                -> `desktop/electron-builder.config.cjs:371` (would re-anchor)
```

`6ff72bc1` adds 7 lines and removes 2 in `desktop/electron-builder.config.cjs`, which pushes the `publish:` line carrying `"github"` from 366 down to 371. `docs/security.md:339` cites that line by number, and `tests/test_readme_claims.py:2019` pins the pair, so the anchor has to move with it.

**The fix is mechanical:**

```
```

That updates `docs/security.md` and the table row together. Then please regenerate the manifest (`--write`, then `--strict` on its own to confirm) since `security.md`'s hash changes.

**The third failure is not yours.** `tests/test_guard.py::test_unmask_is_one_pass_not_one_per_table_entry` asserts wall-clock `elapsed < 0.4` on a loaded runner. It passes here on your head and on `main` alike — 266/266 both — so it's a pre-existing flake on a timing bound. I've filed it separately and it shouldn't block you; a re-run will usually clear it.

**On the approval itself:** I verified the manifest and ran `node --test signing.test.mjs` (19 pass, 0 fail), then skipped the Python suite because this looked like a desktop-only change that couldn't touch it. That reasoning was wrong, and specifically wrong for this repo — the security doc carries line-anchored citations *into* `desktop/`, so editing a desktop file is exactly the thing that turns the Python suite red. I've checked and it was already failing at `d557f4f9`, the sha I approved, so the later merges from main didn't cause it and don't excuse it. I should have run the full suite before approving and I didn't. Sorry for the noise — the change itself is still right, and I'll re-review promptly once the anchor moves.

eyalgolan (dismissed):
Re-approving at `4d99bfb0`. The re-anchor is exactly right — `366` -> `371` in both `docs/security.md` and the pinning row in `tests/test_readme_claims.py`, which is the pair that has to move together.

This time with the run I should have done before the first approval:

```
full suite                    12748 passed, 225 skipped, 2 xfailed, 0 failed
--strict alone                OK: 1607 file(s)
reanchor_citations --check    VERDICT=OK
the two previously-red files  140 passed, 12 skipped
```

`main` (`63ab9be4`) is an ancestor, so that is the merge result rather than the branch in isolation.

On the substance, unchanged from my first read and still verified: `windowsSigningPlan(env)` reads `WIN_CSC_LINK` only, so a macOS `CSC_LINK` can no longer name a Windows artifact `-UNNOTARIZED`, and a real `WIN_CSC_LINK` no longer leaves one `-UNSIGNED`. It is naming-only — no credential path changes, nothing new is read or written.

Sorry again for the round trip. The failure was mine: I approved on the manifest and `node --test` alone because this looked like a desktop-only change, and in this repo that reasoning is wrong — `docs/security.md` carries line-anchored citations *into* `desktop/`, so adding lines there is precisely what turns the Python job red. Running the suite regardless of what language the diff is in is the rule I should not have talked myself out of.

Clean as of `63ab9be4`, expires on the next landing.

eyalgolan:
Heads up: this went conflicting when `63ab9be4` landed on main, and it is **only** `RELEASE_MANIFEST.txt` — I checked:

```
conflicted files:        RELEASE_MANIFEST.txt
non-manifest conflicts:  0
```

Nothing to do with your change. The manifest pins a hash of every tracked file, so every landing re-conflicts every open PR — it is the single most annoying thing about contributing here right now and it is on us, not on you.

The one thing worth knowing: **do not hand-resolve it**, and do not merge the two sides. A hash file merges cleanly and wrongly. Take main's copy wholesale and regenerate:

```
git merge origin/main                     # or rebase, your preference
git checkout origin/main -- RELEASE_MANIFEST.txt
git add RELEASE_MANIFEST.txt
uv run python scripts/check_release_manifest.py --strict     # run this ALONE to confirm
```

The `--strict` run on its own matters: running it straight after `--write` only proves my regeneration matches itself. On its own it checks the tree.

Approval stands on the content — it was clean at `4d99bfb0` and this is purely the manifest. Ping me once it is updated and I will re-confirm rather than make you wait on a fresh review.

eyalgolan (dismissed):
Re-confirmed at `a38ea266` — the manifest resolution is right and the approval stands.

Verified on the merge result against current `main` (`859ad6a7`), not on the branch alone:

```
merge into main            0 conflicts
--strict alone             OK: 1608 file(s)
reanchor_citations --check VERDICT=OK
node --test signing.test.mjs   # pass 19   # fail 0
```

And your desktop code is byte-identical to what I reviewed at `4d99bfb0`, so nothing needed re-reading — only the generated files moved.

One note so it doesn't surprise you: `main` moved again while you were fixing this (#348 and #351 landed), so GitHub may show this as behind. It merges clean, so there is nothing to do unless the manifest reconflicts — and if it does, same recipe as before.

Thanks for the quick turnaround on both rounds.

L4XB:
Synced with `main`. The only conflict was `RELEASE_MANIFEST.txt`; I took main's version and regenerated it with `scripts/reanchor_citations.py --apply` followed by `scripts/check_release_manifest.py --write`, so the only rows this branch changes are the three files it touches (`desktop/electron-builder.config.cjs`, `desktop/signing.cjs`, `desktop/signing.test.mjs`).

`node --test desktop/signing.test.mjs` — 19/19 pass. `pytest tests/test_readme_claims.py` — 135 passed, 12 skipped.

The merge commit dismissed the approval; re-approval is needed before it can go in.

eyalgolan:
Third time on the manifest, and still nothing to do with your change:

```
conflicts against main (`56775e42`):  RELEASE_MANIFEST.txt
non-manifest conflicts:             0
desktop/ vs the sha I approved:     IDENTICAL
```

I am not going to keep asking you to re-pin a generated file every time somebody lands. The content approval stands and has not moved since `4d99bfb0`; when this reaches the front of the queue I will resolve the manifest at land time, since that is mechanical and it is our churn, not yours.

If you would rather clear it yourself so the PR shows green, the recipe is unchanged — take main's copy wholesale, never merge the two sides, then regenerate:

```
git merge origin/main
git checkout origin/main -- RELEASE_MANIFEST.txt
git add RELEASE_MANIFEST.txt
uv run python scripts/check_release_manifest.py --strict
```

Either way is fine. I filed #358 about how often this is happening — five landings in one evening, every one touching the manifest — because making contributors do this by hand is the actual defect.

eyalgolan (dismissed):
Re-approving so the state matches what I said in the comment above — the content approval had lapsed only because the head moved for the manifest.

Verified on `fa30168f`:

```
desktop/ vs the sha I reviewed   IDENTICAL
--strict alone                   OK: 1609 file(s)
reanchor_citations --check       VERDICT=OK
node --test signing.test.mjs     # pass 19   # fail 0
conflicts against main           RELEASE_MANIFEST.txt only (0 source)
```

The fix itself is unchanged and still right: `windowsSigningPlan(env)` reads `WIN_CSC_LINK` only, so a macOS `CSC_LINK` can no longer name a Windows artifact `-UNNOTARIZED` and a real `WIN_CSC_LINK` no longer leaves one `-UNSIGNED`. Naming only — no credential path moves.

Approved as of `56775e42`. The manifest is mine to resolve at land time; you do not need to touch this again.

eyalgolan:
Thanks — the central fix is correct and it closes a real hole. Two small things block it.

The claim checks out: `CSC_NAME` is read by electron-builder in exactly one place, `app-builder-lib/out/codeSign/macCodeSign.js:262`, and nowhere on the Windows path. So on main, `CSC_NAME` + notary credentials genuinely named an unsigned `.exe` as a release artifact, and rows 1–2 were worse still — an Apple-notarization word (`-UNNOTARIZED`) stamped on a Windows binary. On both arms of your branch those rows are `-UNSIGNED`.

macOS does not move: `macMode`, `macTag` and `macFatal` are byte-identical across all 16 env combinations on main, your branch, and the merge result — with a positive control (a deliberate one-word macOS mutation on a copy of main) proving the harness can see a macOS change.

---

## B1 — `desktop/signing.cjs:172-177` asserts a fail-safe direction the code does not have

> *"a name that under-claims can be corrected by setting `WIN_CSC_LINK`; a name that over-claims is uploaded to a release page."*

`windowsSigningPlan` clears the tag on the **presence** of `WIN_CSC_LINK`, never on the **outcome** of signing — and electron-builder treats a Windows signing failure as a warning, not a build failure. In the pinned `app-builder-lib@26.15.3`:

- `out/winPackager.js:116-120` — the signing promise is `.catch(…)` → `log.warn("signing failed for file, queue will continue to next file")`
- `out/winPackager.js:129` — `if (!didSignSuccessfully && this.forceCodeSigning) throw …` is the only hard stop
- `out/platformPackager.js:610-612` — `forceCodeSigning` defaults **false**
- `out/codeSign/windowsSignToolManager.js:155-158` — an unresolvable `.pfx` or wrong `WIN_CSC_KEY_PASSWORD` ends in `"no signing info identified, signing is skipped"`

Driving your real config:

```
WIN_CSC_LINK set:   win.artifactName = "${productName}-${version}.${ext}"
                    effective forceCodeSigning = false
empty env (control): win.artifactName = "${productName}-${version}-UNSIGNED.${ext}"
```

So: set `WIN_CSC_LINK`, have signtool fail for any reason, and the build emits a release-shaped `no_human-<v>.exe` whose `Get-AuthenticodeSignature` says `NotSigned` — the exact filename-claims-one-thing/signature-says-another failure this module exists to prevent, in the direction the new sentence says cannot happen. `forceCodeSigning` appears nowhere in `desktop/`, `packaging/` or `.github/` (positive control: `artifactName` returns two hits in the same file).

Either add `forceCodeSigning: winPlan.signed` to the `win` block so the tag and the outcome agree by construction, or rewrite the sentence to say the tag reflects *configuration*, not signing outcome. Don't leave it standing.

## B2 — the empty-string case of the input this PR adds is unpinned

One-line mutation in `windowsSigningPlan`, `present(env, WINDOWS_CERTIFICATE_VAR)` → `env[WINDOWS_CERTIFICATE_VAR] !== undefined`:

```
[M7] exit=0  # pass 19 # fail 0   *** SURVIVOR ***
M7 full desktop suite: # tests 434 / # pass 433 / # fail 0
```

`WIN_CSC_LINK=""` is the canonical GitHub Actions shape for an unconfigured secret, and electron-builder special-cases it (`platformPackager.js:82 // allow to specify as empty string`, then `windowsSignToolManager.js:76-78` returns null → not signed). Under the mutation that build is named `no_human-<v>.exe`. The macOS twin *is* pinned at `desktop/signing.test.mjs:153-154` — the reasoning was applied to the old input and not to the new one. Two asserts closes it; re-run M7 and show it red.

---

Smaller, worth doing while you're here: strike the `+ plan.artifactTag +` suggestion at `electron-builder.config.cjs:173` (it's a written invitation to reintroduce this same bug on Linux); add `WIN_CSC_LINK` to `docs/WINDOWS.md §5.5`, which currently says flatly "The installer is **not signed**" and never mentions the variable that now changes the filename's claim; and consider making the wiring cell behavioural — `require()` the config twice, with and without `WIN_CSC_LINK`, and assert the two `win.artifactName` values. That pins the same thing, survives reformatting (a behaviour-preserving `const win =\n{` currently turns it red), and would also have caught M7.

The `RELEASE_MANIFEST.txt` conflict is the ordinary generated-file class — `scripts/check_release_manifest.py --write` resolves it cleanly (`wrote 1614 row(s)` / `OK: 1614 file(s) match`).

One correction to your PR body: the verification block says `node --test desktop/*.test.mjs` → "431 pass, 3 fail, pre-existing on main". It doesn't reproduce here — from the repo root `# tests 443 / # pass 442 / # fail 0`, and under CI's own invocation (cwd `desktop/`, `uiPages.test.mjs` excluded, `ci.yml:717-729`) `# tests 434 / # pass 433 / # fail 0 / # skipped 1`. All three named tests pass. Looks like a cold Electron binary on your side rather than a code defect — but it's stated as established fact.

Separately, and **not** something to fix in this PR: `electron-builder.config.cjs:327` sets `nhCanAutoUpdate: plan.canAutoUpdate` from the **macOS** plan inside the single shared `extraMetadata`, so it is stamped into the Windows and Linux apps too — while three shipped sentences say it stays false there (`:118`, `:158`, `docs/WINDOWS.md:688`). Pre-existing on main and unchanged by you. I checked how much it matters in practice before saying anything: `grep -c "APPLE_\|CSC_" .github/workflows/ci.yml` is **0**, so every CI-built artifact has no Apple credentials and genuinely gets `canAutoUpdate=false`. It would only bite a *local* build that exports `CSC_LINK` and produces Windows or Linux targets. I'm filing it separately.

eyalgolan (dismissed):
Re-approving at `d0b7c8d9` — the review reset when you merged main, but nothing of yours moved.

```
desktop/ vs my first approval   IDENTICAL (byte-for-byte)
--strict alone                  OK: 1614 file(s)
citations                       VERDICT=OK
node --test signing.test.mjs    # pass 19   # fail 0
```

And it is mergeable again now that the manifest merge driver landed (`0b8c2dc4`) — which is the fix for the churn that made you re-pin this three times. Sorry it took that long; the driver is the thing that should have existed before you ever opened this.

Approved as of `a8a04496`.

eyalgolan (approved):
Re-approving at `93620032`. **I pushed two merge commits to your branch** to clear the manifest conflict — flagging that plainly so a force-push from your side does not surprise either of us. They are merges, not a rebase, so your local clone fast-forwards and nothing of yours was rewritten.

Both were the same mechanical resolution: take main's `RELEASE_MANIFEST.txt` wholesale and regenerate from the merged tree, never interleave two sides of a hash file. Two of them because main moved again between my resolving it and pushing.

Verified on the merge result against `49eaa906`, not on the branch:

```
conflicts                      0
--strict alone                 OK: 1619 file(s)
reanchor_citations --check     VERDICT=OK
node --test signing.test.mjs   # pass 19   # fail 0
manifest rows differing        exactly the 5 files this branch changes
```

That last line is the check that matters for a regenerated lockfile: if my resolution had picked up anything it should not have, the row set would be wider than your diff. It is not.

Your code is untouched from the review I did at `4d99bfb0` — `windowsSigningPlan(env)` reading `WIN_CSC_LINK` only, so a macOS `CSC_LINK` can no longer name a Windows artifact `-UNNOTARIZED` and a real `WIN_CSC_LINK` no longer leaves one `-UNSIGNED`. Naming only, no credential path moves.

Clean as of `49eaa906`.

eyalgolan:
Landed as `676a9cfc`, with you as the commit author.

PR #348, opened 2026-09-13
test(base-tree): drop a precondition that is not one

### LINKED ISSUE #344: A test asserts pytest is on PATH, so the suite is red by default under a normal local invocation
`tests/test_base_tree_gate.py::test_pre_existing_red_test_excused_when_runner_rewrites_the_command` fails whenever the suite is run without the venv's `bin` on `PATH`, which is a legitimate invocation:

```
$ .venv/bin/python -m pytest -q tests/test_base_tree_gate.py
E   AssertionError: pytest must be resolvable via PATH to strip it out
E   assert None
tests/test_base_tree_gate.py:351

$ PATH=".venv/bin:$PATH" .venv/bin/python -m pytest -q tests/test_base_tree_gate.py
11 passed
```

The assertion is `assert pytest_bin, "pytest must be resolvable via PATH to strip it out"`, and `shutil.which("pytest")` is `None` when you invoke the interpreter directly instead of activating the venv. CI activates it, so `main` is green there — this only bites locally.

## Why it is worth fixing rather than documenting

I ran the full suite against four contributor PRs today and reported "1 failed" on each, then told three contributors the failure was "pre-existing on `main`". That was literally true and completely misleading: the failure was **my invocation**, not the tree, and not theirs. With `PATH` set correctly the same suites are `12745 passed, 0 failed`.

A suite that is red-by-default under a normal local invocation teaches people to skim past red. That is the expensive part, not the test itself.

## Suggested fix

Either skip when the binary is genuinely absent:

```python
if not pytest_bin:
    pytest.skip("pytest is not on PATH; this test strips it from a rewritten command")
```

or resolve it the way the rest of the suite runs it — `[sys.executable, "-m", "pytest"]` — so there is nothing to look up. The second is better if the test's subject is the command the runner builds, since it removes the environmental dependency rather than tolerating it.

An `assert` is the wrong shape either way: it reports an environment fact as a product failure.

Measured on `main` (`1d7c28a6`).

L4XB:
Fixes #344

## The assert is not a precondition

```python
pytest_bin = shutil.which("pytest")
assert pytest_bin, "pytest must be resolvable via PATH to strip it out"
```

What the test needs is the bare `pytest` binary being **unresolvable after** it strips PATH, so the fixture repo's detected `pytest -q` fails to launch and the runner's class-3 rewrite fires. Starting with nothing on PATH satisfies that on its own: the strip is a no-op and the scenario is already in place. The real requirement is asserted a few lines further down, and that assertion is untouched:

```python
assert shutil.which("pytest") is None, (
    "PATH stripping did not remove the bare `pytest` binary; ..."
)
```

So the assert only ever fired in the one case where it should not have — `.venv/bin/python -m pytest`, where `which("pytest")` is `None` because the venv was never activated.

## What I verified

Both invocations, on this branch:

```
$ .venv/bin/python -m pytest -q tests/test_base_tree_gate.py          → 11 passed
$ PATH=.venv/bin:$PATH .venv/bin/python -m pytest -q ...same...       → 11 passed
```

**The test's discrimination is unchanged.** I broke `_fix_invocation`'s class-3 rewrite (`return None` instead of `f"{sys.executable} -m pytest{rest}"`) and ran the file with the assert and without it. With the venv on PATH, `test_pre_existing_red_test_excused_when_runner_rewrites_the_command` passes under that mutant **either way** — so this change removes nothing that was grading the rewrite.

## The part that actually improves

With the venv **off** PATH — the invocation the issue is about — that same mutant reddens four tests in this file:

```
FAILED test_coder_introduced_import_breakage_fails_the_attempt
FAILED test_newly_failing_test_still_fails_the_attempt
FAILED test_mixed_failures_fail_only_on_the_newly_failing_test
FAILED test_base_recheck_inconclusive_fails_closed
```

That is a working guard on the rewrite, and it was previously buried under a file that was red before it started. The issue's point — that red-by-default teaches people to skim past red — costs exactly this: a real signal nobody would have looked at twice.

## One observation, not fixed here

The test named for the rewrite does not grade it: the mutant above passes that test in every environment I tried. The four tests listed above are what actually hold the behaviour. Worth its own issue if you want the named test to earn its name; I kept this PR to the reported defect.

`RELEASE_MANIFEST.txt` is restamped in its own commit (`scripts/check_release_manifest.py --write`), and `scripts/reanchor_citations.py --apply` reports `VERDICT=OK` with nothing to rewrite.


eyalgolan (approved):
Approving. The reasoning in the description is right, and it held up when I tried to break it.

## The assert really was not a precondition

I checked the claim rather than taking it: the requirement is asserted **after** the strip, and your diff leaves it untouched.

```python
assert shutil.which("pytest") is None, (
    "PATH stripping did not remove the bare `pytest` binary; the "
    "runner's class-3 rewrite fallback would not be exercised")
```

So the removed line only ever fired in the one case where the scenario was *already* satisfied. Test count is unchanged at 11; the assertion count drops by exactly the one line, which is the right shape for this change.

## The matrix

```
                    main        this PR
venv OFF PATH    1 failed      11 passed
venv ON  PATH   11 passed      11 passed
```

`main`'s failure is at `tests/test_base_tree_gate.py:351` — the assert you removed. So this closes #344 without changing behaviour in the environment that already worked.

## Your mutation evidence reproduces

I restored the class-3 mutant myself (`return None` in place of `return f"{sys.executable} -m pytest{rest}"`, after asserting the target appeared exactly once — a mutation that silently fails to apply looks identical to a caught one):

```
mutant, venv ON  PATH   11 passed          <- the named test does NOT grade the rewrite
mutant, venv OFF PATH    4 failed, 7 passed
```

and the four are exactly the ones you listed. So the assert was hiding a working guard behind a file that was red before it started — which is the real cost #344 was pointing at.

## On the observation you volunteered

You are right, and I am glad you wrote it down rather than leaving it for me to find: `test_pre_existing_red_test_excused_when_runner_rewrites_the_command` does not grade the rewrite it is named for — the mutant passes it in every environment I tried too. The behaviour is actually held by the four tests above. That is worth its own issue, and keeping this PR to the reported defect was the right call. Flagging a weakness next to your own change is the thing that makes a review cheap; thank you for it.

Also confirmed: `--strict` on its own (not after a `--write`, which would only validate my own regeneration) reports `OK: 1607`, citations `VERDICT=OK`, and the full suite is `12748 passed, 225 skipped, 0 failed`.

CI's Python job is red only on `test_unmask_is_one_pass_not_one_per_table_entry`, which is a wall-clock flake unrelated to your branch — it passes here on your head and on `main` alike. That is #349, and the fix is up as #352; nothing for you to do.

Clean as of `e5b07ac8`, expires on the next landing.

PR #352, opened 2026-09-13
Assert _unmask's shape by counting, not by the clock

### LINKED ISSUE #349: Flaky: test_unmask_is_one_pass_not_one_per_table_entry asserts wall-clock <0.4s and fails on loaded CI runners

eyalgolan:
Closes #349.

`test_unmask_is_one_pass_not_one_per_table_entry` asserted wall-clock `elapsed < 0.4` on `guard.evaluate`. Its own docstring said the bound existed to assert *"the SHAPE, not a machine speed"* — but a stopwatch on a shared runner measures the runner. It reddened two unrelated contributor PRs in one afternoon (#342 and #348), and on #348 it was the **only** failure.

That cost is worse than it looks: a red build nobody believes teaches people to re-run CI until it goes green, which is exactly the habit that hides a real failure. It nearly did here — #342 had two genuine citation-drift failures standing next to this flake.

## What replaces it

The same claim, counted instead of timed. `_CountingTable` records how much of the table each call touches, and the test asserts the invariant the docstring always meant:

> cost follows the number of masks in the **string**, never the size of the table

Ten times the table with the same two masks must cost exactly the same, and no call may request a whole-table view (`items`/`keys`/`values`/iteration) — that view is the operation the quadratic version performed once per token.

## Both halves fail on the regression they name

Confirmed by restoring the quadratic implementation (`for key, val in table.items(): tok = tok.replace(key, val)`) and re-running:

```
direct       AssertionError: _unmask scanned the whole 5000-entry table 1 time(s)
via evaluate 8003 of 8003 calls scanned the table
             (0 of 8003 on the current one-pass implementation)
```

The end-to-end half is checked separately rather than assumed: the spy hands the real `_unmask` a counting copy of whatever table `evaluate` built, so a table loop reintroduced anywhere under that entry point still surfaces as a scan. It also asserts `_unmask` was called at all, so the half cannot pass vacuously.

## Verified

```
tests/test_guard.py + tests/test_structural_budget.py   284 passed
scripts/check_release_manifest.py --strict   (alone)    OK: 1607 file(s)
scripts/reanchor_citations.py --check                   VERDICT=OK
manifest diff vs main                                   exactly 1 row: tests/test_guard.py
```

`monkeypatch` rather than `unittest.mock` to match the idiom already in this file — no new import.

## Not addressed here

`tests/test_guard.py` has two other wall-clock assertions (`time.monotonic` around line 590, `time.perf_counter` around line 1461) and `time.process_time` at 441. None of them has been seen failing, so I left them alone rather than widen this past the reported defect — worth a look if either turns up red.


eyalgolan:
Closing in favour of #351, which does the same job and was open twelve minutes before I opened this one.

I filed #349 at 18:51 and started building without checking whether anyone had picked it up; @L4XB had. His is first and it is the better PR: a separate clearly-named test for the flat-in-table-size claim, three mutants rather than one (including a constant-factor one a stopwatch structurally cannot see), and he *measured* the neighbouring timing test's headroom instead of just naming it as unexamined, which is what I did here.

We converged independently on the same `_CountingTable` design, which is mild evidence it is simply the right shape for this.

Nothing here is lost except the `guard.evaluate` spy, which I have offered on #351 as an optional follow-up rather than a condition. Branch `fix/349-unmask-shape-not-clock` deleted.

PR #355, opened 2026-09-13
test(base-tree): grade the rewrite the test is named for

### LINKED ISSUE #353: test_pre_existing_red_test_excused_when_runner_rewrites_the_command does not grade the rewrite it is named for
Raised by @L4XB while fixing #344, and confirmed independently here.

`tests/test_base_tree_gate.py::test_pre_existing_red_test_excused_when_runner_rewrites_the_command` is named for the runner's class-3 rewrite, but it does not grade it.

## Measured

Mutant: `src/no_human/testing/runner.py` class-3 branch returns `None` instead of the rewritten command (target asserted to occur exactly once before writing, so the mutation is known to have applied):

```python
-            return f"{sys.executable} -m pytest{rest}"
+            return None  # MUTANT
```

Result on the whole file:

```
mutant, venv ON  PATH    11 passed                 <- the named test passes
mutant, venv OFF PATH     4 failed, 7 passed       <- the named test STILL passes
```

The four that actually fail:

```
test_coder_introduced_import_breakage_fails_the_attempt
test_newly_failing_test_still_fails_the_attempt
test_mixed_failures_fail_only_on_the_newly_failing_test
test_base_recheck_inconclusive_fails_closed
```

So the behaviour is held by its neighbours, and the test carrying the name contributes nothing to catching a broken rewrite in either environment.

## Why it matters

Someone deleting or weakening the class-3 rewrite would look for the test named after it, see it green, and conclude the change was safe. The real guard is four tests away and named for something else. That is the same trap as a green suite that never exercised the code — the signal exists but not where a reader will look for it.

## Suggested shape

Make the named test assert the rewrite itself: that the command the runner ends up executing is `{sys.executable} -m pytest ...` and not the bare `pytest` it started with, rather than only asserting the outcome (`AWAITING_APPROVAL`) that several other paths can also produce. The outcome assertion is worth keeping — it just should not be the only thing standing between this name and a regression.

Not urgent, and deliberately out of scope for #348 (PR #348), which correctly stayed on the reported defect.

L4XB:
Fixes #353

The follow-up I flagged in #348 and said I would keep out of that PR.

## What the name was not covering

The test asserted only `AWAITING_APPROVAL` and a non-failed attempt, and those are reachable by paths that never rewrite anything. It now records every command that reaches a shell — through `_run_shell`, which `run_tests` aliases at call time precisely so a patch there applies — and asserts three things about the rewrite, in order:

1. the detected bare `pytest` was attempted;
2. a `{sys.executable} -m pytest` run **followed** it;
3. the arguments came through **verbatim**.

## The third one is the point, and I nearly missed it

My first version asserted (1) and (2) only. A mutant that rewrote to `f"{sys.executable} -m pytest"` — dropping `rest`, the node ids — **passed**.

That is the half the original send-back turned on. The guard this test exists to keep deleted was discarding the rewritten verdict because the command *string* differed while the node ids it asked for did not. A rewrite that dropped them is a different request wearing the same name, and a test named for this behaviour that cannot see that is barely better than the one it replaces.

## Mutants

All four red now; all four were green before, in the environment where the test used to run.

| mutant on `_fix_invocation` / its retry | before | after |
| --- | --- | --- |
| class-3 branch returns `None` | passed | **fails** |
| class-3 branch returns `cmd` unchanged | passed | **fails** |
| the retry never executes (`if False and …`) | passed | **fails** |
| the rewrite drops `rest` (the node ids) | passed | **fails** |

The `None` mutant is the one #353 measured, so it is the direct check: `11 passed` before, the named test red now.

## Environments

This branch sits on `main`, so it still carries the `assert pytest_bin` precondition #348 removes — CI activates the venv and passes it; locally I ran with the venv on `PATH`.

I verified the off-`PATH` case too, by simulating #348 (dropping that assert) on top of this branch: clean runner **1 passed**, `None` mutant **1 failed**. So the grading holds in both invocations once #348 lands, and the two PRs touch different lines of the same function — no conflict either way they merge.

`tests/test_base_tree_gate.py`: 11 passed. `-k "runner or base_tree or invocation"`: 226 passed, 2 skipped. `reanchor_citations.py --apply` reports `VERDICT=OK`; the manifest is restamped in the same commit.


L4XB:
CI is red, and the red is the interesting part — so I am leaving it visible while I chase it rather than weakening the assertion.

**What CI recorded:**

```
['pytest -q', 'pytest -q']
```

The bare command **launched**, twice. There was never a rewrite to grade, so the new assertion is correct and the fixture's premise is what is wrong.

That is the same class of finding as #353 itself. The docstring says this test "reproduces the rewrite end-to-end by stripping the bare `pytest` binary's directory from PATH". On that runner it does not. Locally it does — the recording is:

```
pytest -q
<python> -m pytest -q
pytest -q -rA test_preexisting.py::test_preexisting
<python> -m pytest -q -rA test_preexisting.py::test_preexisting
```

— the attempt's own run and the bounded base recheck, each bare then rewritten.

**What I ruled out rather than assumed.** My first guess was `_env_for` prepending a repo venv's `bin` to the child's PATH, since CI has a `.venv` in the checkout and I do not. I tested it: planted a `.venv-probe/bin/pytest` in the checkout and re-ran. The test still passed, because `_env_for` is called with the *task's* repo — the tmp fixture — not with no_human's own. So that is not it.

**What I pushed instead of another guess.** The spy now records the child's `PATH` beside each command, and the failure distinguishes the two situations: a rewrite that did not happen, versus a bare command that launched and therefore never asked for one. `shutil.which` a few lines above measures the *test* process; `_env_for` builds the child's environment separately, and where they disagree is exactly what needs naming.

The next CI run will print where the child resolved `pytest`. I will fix the fixture from that, not from a theory.

Worth saying plainly: whatever the cause, the named test has not been exercising the rewrite on CI either — so #353's "green where a reader would look" is, on that runner, greener than it looked.

L4XB:
The diagnostic answered, and the answer is a product bug rather than a fixture one.

CI reported **no** PATH entry holding a `pytest`, and the bare command still appeared twice with no rewrite. It never ran. It failed to launch, and `_fix_invocation` declined to fix it:

```python
or ("pytest" in out_lower and "command not found" in out_lower)
```

That is bash's and zsh's phrasing. **dash** — the `/bin/sh` on Debian and Ubuntu, and therefore on most Linux CI runners and containers — says `sh: 1: pytest: not found`. busybox says `sh: pytest: not found`. Neither contains "command not found".

Measured against the real messages, before the fix:

```
bash/zsh                  detected=True  rewrite=<python> -m pytest -q
dash (Debian/Ubuntu sh)   detected=True  rewrite=None
busybox sh                detected=True  rewrite=None
```

`_is_invocation_error` never had this problem — it returns True on zero counts — so the launch failure was **detected** everywhere and **repaired** nowhere. That is why it surfaces as an honest "no test evidence" instead of a crash, and why nobody noticed: a project whose detected command is a bare `pytest` gets the retry on a developer's mac and no retry on the runner.

The needle now matches the shell's `<name>: [command ]not found` shape, anchored at end of line. Two negative controls are pinned with it: pytest's own `ERROR: file or directory not found: tests/x.py` has no colon before the phrase, and `sh: 1: vitest: not found` is a different tool. Reverting the needle to the bash-only form reddens the new test.

So #353's finding was even truer than it looked. The named test was green in CI not because the rewrite was held elsewhere, but because **the rewrite never happened there at all** — and the four neighbours that do grade it were, on that runner, grading a path that could not fire.

Locally: `test_runner.py` + `test_base_tree_gate.py` + `test_flaky_rerun_attribution.py` — 120 passed.

Happy to split the runner fix into its own PR if you would rather keep this one test-only — it is one regex and two tests, and it is what makes the test green on CI rather than only on a mac.

eyalgolan (changes requested):
The test work here is excellent and I want to say that first, because what follows is a packaging request rather than a criticism of the code.

Requesting one change: **this PR contains a production behaviour fix that its title and body do not mention.**

## The undisclosed part

`src/no_human/testing/runner.py` replaces the class-3 match

```python
- or ("pytest" in out_lower and "command not found" in out_lower)
+ or _SHELL_NOT_FOUND.search(output) is not None
```

with a new module-level regex. That is a real change to retry behaviour on every host, in a PR titled `test(base-tree): ...` whose body describes only test grading. Nothing in the description mentions dash, busybox, or that production code moved.

I checked whether your new test needed it. It does not — reverting **only** `runner.py` to main and keeping your tests:

```
tests/test_base_tree_gate.py   11 passed
```

So it is an independent second fix, not a dependency of #353.

## To be clear, the fix itself is right, and I verified it

Your new tests against **main's** runner:

```
AssertionError: ('dash', None)
assert None == '<python> -m pytest -q'
```

dash is `/bin/sh` on Debian and Ubuntu, so this was silently costing test evidence on most Linux CI. That is a genuine find and a good one. I measured the net effect both ways:

```
                                 main      this PR
bash    "pytest: command not found"   REWRITE   REWRITE
dash    "sh: 1: pytest: not found"    none      REWRITE
busybox "sh: pytest: not found"       none      REWRITE
trailing CR                           none      REWRITE
noisy multi-line output               none      REWRITE
```

One narrowing, which I report because I went looking for it: the old rule matched `pytest` and `command not found` **anywhere in the output**, the new one requires them on the same line. I could only construct a case that regresses by writing output no real shell emits (`+ pytest -q\nbash: line 5: command not found` — bash always names the command on the same line). So I am satisfied it is strictly better in practice, not just on the cases you chose. The negative test is the part I liked most: `file or directory not found: tests/x.py` and `vitest: not found` both correctly left alone.

## What I am asking for

Split it, or relabel it — your call, and either is fine by me:

- **Split** — `runner.py` + `test_runner.py` as their own PR (it deserves its own line in the log and its own issue; it fixes a bug nobody has filed), leaving this one purely the #353 grading work; or
- **Relabel** — retitle to something like `fix(runner): rewrite bare pytest on dash and busybox too; grade the rewrite in the base-tree test`, and add a short section to the body describing the production change.

The reason this matters more here than in most repos: we squash-merge and the PR title becomes the permanent commit message. A commit reading `test(base-tree): ...` that changed retry behaviour on every Linux host is a commit someone bisecting a retry regression will skip straight past. Five minutes now, and I will re-approve as soon as it is updated — there is nothing else outstanding.

## Everything else checks out

Your four mutants are the right four, and the third one is the interesting one — a rewrite that drops `rest` is exactly the shape the original send-back turned on, and catching that is what makes the test earn its name. I confirmed `main` returns `None` on the `None` mutant case (`11 passed` before, red after), which is the direct check for #353.

I also verified the interaction with #348, since both touch the same function: source files merge clean, only `RELEASE_MANIFEST.txt` conflicts (that is the generated-file churn that hits every concurrent PR here, not anything either of you did), and the merged tree passes `102 passed` across both test files.

L4XB:
Split, as you offered — the log line is the better argument and I should have made it myself when I pushed the second commit rather than only in a thread.

`runner.py` and `test_runner.py` are now **#360**, titled for what they do. This branch is the #353 grading work and nothing else: the net diff against `main` is `tests/test_base_tree_gate.py` plus the manifest restamp.

## One thing to expect on the check here

This branch will be **red on CI until #360 lands**, and for exactly the reason the assertion exists: the test now grades the class-3 rewrite, and on a dash host — which this CI is — the rewrite does not happen at all. Your `11 passed` with main's runner is right on a bash host; the split makes the dependency visible rather than removing it.

Your call which way round: merge #360 first and this goes green on its own, or I rebase this onto #360 now and you get a green check here immediately. Say the word and I will do the rebase.

## Carried across from your review

I moved your finding about the narrowing into #360's body rather than leaving it in this thread — that the old rule matched the two strings anywhere in the output while the new one requires them on one line, and that the only regressing input is output no real shell emits. It belongs with the commit that makes the change.

L4XB:
Synced with `main`. The only conflict was `RELEASE_MANIFEST.txt`; I took main's version and regenerated it with `scripts/reanchor_citations.py --apply` then `scripts/check_release_manifest.py --write`.

The production change you flagged is already out of this branch and lives in #360 — this one is now the test work only.

`pytest tests/test_base_tree_gate.py` — 11 passed.

L4XB:
Closing this and folding it into #360, because the two cannot actually be separated.

You asked me to split the production fix out of this PR, and that was right — a `test(...)` title must not carry a behaviour change. But CI then showed why the halves do not stand alone: the instrumentation this PR adds asserts that the class-3 rewrite HAPPENED, and on your runners it does not happen. The assertion fired with

```
the runner never re-ran through `sys.executable -m pytest`.
Commands: ['pytest -q', 'pytest -q']
```

That is #360's bug, verbatim: `main` only rewrites when the shell says `command not found`, which dash and busybox do not say. So this PR is green on a macOS runner and red on CI, and can only go green once #360 lands.

Rather than leave a PR blocked on another one, the instrumentation now lives in #360, where it is the end-to-end proof of that fix instead of a cell waiting for it. #360 is a `fix(...)`, so the packaging objection that started this does not come back: the title matches what is inside.

One thing worth flagging, since I got it wrong first and pushed it: my initial attempt deleted the whole cell from this branch. The cell is on `main` and passes there — only the instrumentation depends on #360 — so that would have taken a main test with it. Reverted; this branch is now identical to `main`, which is why there is nothing left to merge.

Grading for #353 is therefore in #360: `pytest tests/test_base_tree_gate.py tests/test_runner.py` — 102 passed.

PR #356, opened 2026-09-13
fix(vcs): decide containment on the entry, not on its parent

### LINKED ISSUE #354: commit_paths stages a symlink pointing outside the repo, committing an absolute local path into the public repo
Post-landing audit of #335 (merged as `e5b07ac8`). The core fix in that PR is real and correct — a caller-supplied broken symlink is now staged. These are consequences that were not disclosed and are worth fixing as a follow-up, not a reason to revert.

## 1. An in-repo symlink whose target is OUTSIDE the repo is now committed

`src/no_human/vcs/git.py` — the `relative_to(repo_root)` containment check now constrains **the parent directory only**; the entry's own name is never resolved. Measured A/B against the merge-base, with two independent controls returning IDENTICAL first:

```
=== symlink_escapes_to_outside_file   *** DIFFERS ***
  BEFORE: committed={'app.py': '100644', 'sibling.py': '100644'}
  AFTER:  committed={'app.py': '100644', 'esc.md': '120000', 'sibling.py': '100644'}
          blobs={'esc.md': '<ROOT>/outside_secret.txt'}
```

**The committed blob is an absolute local filesystem path.** That is the part that matters here: this repo's first constraint is that no employer or local-machine trace reaches the published tree, and this path commits one directly — a link whose content is `/Users/<name>/<some>/<local>/<layout>`. A checkout also materialises a link that any path-based reader follows out of the worktree.

Same class, also changed: a symlink to an outside *directory* (now committed, previously skipped), and a symlink to an **ignored** file — which previously failed loudly with `The following paths are ignored by one of your .gitignore files` and now silently succeeds.

There is a defensible argument that this is "more git-correct", since git itself will stage such a link. But it is outside the PR's stated scope, it was not disclosed, and a shipped test docstring explicitly denies it (see 3).

One mitigation already in place, verified: `scripts/check_release_manifest.py` hashes the link target *string* rather than following it.

## 2. A different file than the caller named can be committed

`os.path.abspath` is `normpath(join(cwd, path))`, which collapses `..` **textually, before** any symlink resolution.

```
=== dotdot_through_symlinked_dir   *** DIFFERS ***     # dirlink -> <ROOT>/elsewhere/d
  paths_in: ['<ROOT>/work/dirlink/../victim.md', ...]
  BEFORE: committed={'app.py','sibling.py'}            # correctly skipped
  AFTER:  committed={..., 'victim.md': '100644'}
          blobs={'victim.md': 'I LIVE INSIDE THE REPO\n'}
```

The caller named a file containing `I LIVE OUTSIDE THE REPO`; the commit captured a **different file of the same basename**. This is the only case where commit contents diverge from what the caller asked for, and unlike (1) it has no "more git-correct" defence. In-product reachability is unproven — it needs a `..` traversing a symlinked directory — but the code behaviour is confirmed.

## 3. A shipped test docstring asserts the opposite of (1)

`tests/test_git_commit_paths.py`, verbatim:

> "The membership test still refuses what is genuinely elsewhere, so the fix does not buy broken symlinks by widening what may be staged."

Refuted by (1). False claims in shipped text are this project's named house defect.

## 4. That test cannot detect the property it is named for

```python
assert not [f for f in files if "outside" in f], files
```

`_committed_files()` returns `subprocess.stdout` — a **str**. Iterating a str yields single characters, so `"outside" in f` is never true:

```
first 5 iteration items: ['i', 'n', 's', 'i', 'd']
ASSERTION PASSED even though 'outside.py' IS in the commit  -> VACUOUS
set-based equivalent -> ['outside.py'] (would correctly FAIL)
```

The module already defines the right helper (`_committed_names`, NUL-split into a set) and this test does not use it. The fixture is also structurally incapable of exhibiting the widening class, which needs a symlink *inside* the repo.

Related: the PR body's mutation evidence for this test is misattributed. The reported `1 failed, 18 passed` reproduces, but the kill comes from git, not from the assertion:

```
GitError: git add -- .../outside.py inside.py failed (128):
  fatal: '.../outside.py' is outside repository at '.../work'
```

## 5. Two now-false claims in an untouched consumer

`src/no_human/vcs/manifest_repair.py`:
- "`GitRepo.commit_paths` resolves each entry with `Path(p).resolve()`" — that line no longer exists.
- `continue  # outside the repo — commit_paths will skip it too` — falsified. `_stage_untracked_for_approve` still uses the old `Path(p).resolve()` semantics, so the two now disagree:

```
pin-check staged:        ['control.md']                    # esc.md skipped
commit_paths committed:  app.py, control.md, esc.md (120000)
```

The escaping symlink is committed without the export-guard pin check having seen it. The likely consequence is an extra repair round-trip rather than a bypass, since the guard reads the index at commit time — but the divergence and the false comment are confirmed.

## Also found, pre-existing, in the same function

An unreadable parent directory (`chmod 000`) causes the file to be **silently dropped** while the commit succeeds. Identical in both arms, so not a regression — but it violates the fail-closed rule and it lives here.

## Suggested scope for the fix

Resolve the entry's own name for the containment decision (not just its parent), decide deliberately whether an out-of-repo symlink should be staged or refused, replace the vacuous assertion with the existing `_committed_names` helper and a fixture that actually exercises the class, and reconcile `manifest_repair`'s two claims with what `commit_paths` now does.

L4XB:
Fixes #354

My regression, from #335. Thanks for the audit — both consequences reproduce, and both are worse than the thing they came from.

## What went wrong

**A symlink inside the repo pointing outside was committed.** Containment was tested on the parent directory only, and git stages such a link happily. The committed blob is then an absolute local filesystem path — `/Users/<name>/…` — published in the tree. This repo's first constraint is that no local trace reaches the public tree, and that blob is one.

**`..` was collapsed before anything was resolved.** `os.path.abspath` is `normpath(join(cwd, path))`, so `work/dirlink/../victim.md` became `work/victim.md`: an in-repo file of the same basename, committed **in place of** the one the caller named.

## The fix

The entry's own name is resolved for the containment decision, and a link leading out of the repo is skipped with a warning naming the path and its target. That is the pre-#335 outcome, said out loud instead of falling out of a membership test.

`os.path.realpath`, deliberately, not `Path.resolve(strict=True)`: realpath does not raise on a broken link, it returns the missing target. So the link this method exists to stage still passes whenever that target is in-repo — which is the whole point of #335, and is still pinned by `test_a_caller_supplied_broken_symlink_is_staged_under_its_own_name`.

For the `..` half, the path is joined without normalising, leaving the `..` for `.parent.resolve()` — which walks the link first and the `..` after.

## Measured

A/B on a fixture that carries a real change, so the `stage_all()` empty-commit fallback cannot mask the per-path decision (my first probe had no other change, the fallback fired, and everything was committed — which is worth saying because it would have looked like the defect):

```
before   committed: ['app.py', 'esc.md', 'victim.md']
         esc.md blob = '/private/var/folders/…/outside_secret.txt'
         victim.md blob = 'I LIVE INSIDE THE REPO\n'
after    committed: ['app.py']
         skipping esc.md: it is a symlink to …/outside_secret.txt, outside the repository
```

Two mutants, each caught by the test written for it:

| mutant | test that reddened |
| --- | --- |
| `os.path.abspath` restored | `test_dotdot_through_a_symlinked_directory_does_not_swap_the_file` |
| the entry's own name not re-resolved | `test_a_symlink_pointing_out_of_the_repo_is_not_committed` |

One mutant I could **not** kill, and it is not a hole: replacing `os.path.realpath` with `Path.resolve()` changes nothing, because `resolve()` has been non-strict since 3.6 and also returns the missing target. Equivalent mutant, reported rather than counted.

## The rest of the audit

- **(4) the vacuous assertion.** `[f for f in files if "outside" in f]` iterated a **str**, so it yielded characters and could never be non-empty. It reads `_committed_names` — the set helper the module already had — now.
- **(3) the false docstring.** That test claimed the fix "does not buy broken symlinks by widening what may be staged". It did. Replaced with what the test actually grades, plus the note that an outside path handed to `git add` makes git itself refuse, so the membership test was never what kept `outside.py` out.
- **(5) the two false claims in `manifest_repair`.** Both corrected, including the `continue  # commit_paths will skip it too` comment — still true, but for a different reason than it said, and the divergence it papered over is now named.

## Not in this PR

The `chmod 000` parent directory that silently drops a file. Your audit places it as pre-existing and identical in both arms, and it is a different rule (fail-closed on an unreadable path) from containment. Happy to take it next if you want it here.


L4XB:
CI's only failure here is the stopwatch flake, not this change:

```
FAILED tests/test_guard.py::test_unmask_is_one_pass_not_one_per_table_entry
  guard.evaluate took 0.417s on 4000 quoted tokens
```

That is #349, and #351 replaces the clock with a count. Second sighting today, and the margin is widening rather than holding: **0.40029s** on PR #342's run and **0.417s** here — 100.07% and 104% of a 400 ms budget. It measures 131 ms on a quiet laptop, so the runner is 3x slower than the bound assumes it can be.

Nothing else in the run is red, and this branch touches `vcs/git.py`, `vcs/manifest_repair.py` and two test files — nothing under `guard`.

eyalgolan (approved):
Approving. This is a P1 fix and it is the right one — thank you for taking the audit seriously rather than defensively, and for finding two things in it that I had not.

## I reproduced both halves against `main`, not just read your table

Symlink escape, on a fixture carrying a real change so the `stage_all()` fallback could not mask the per-path decision:

```
BEFORE  committed: ['app.py', 'esc.md']
        esc.md blob = '/var/folders/.../outside_secret.txt'      <- absolute local path, published
AFTER   skipping esc.md: it is a symlink to .../outside_secret.txt, outside the repository
        committed: ['app.py']
```

The `..` swap, on a fixture where the two readings land on genuinely different files (`work/dirlink` -> `repo/other`, so textual collapse and link-first disagree):

```
BEFORE  committed work/victim.md = 'TEXTUAL-COLLAPSE TARGET (wrong file)'
AFTER   committed victim.md      = 'LINK-FIRST TARGET (the one the caller named)'
```

And the property this must not break — #335's whole point — still holds:

```
broken symlink to an in-repo target:  committed ['app.py', 'link.py']   (both arms)
```

## Your mutants, rebuilt here

Each kills exactly the test named for it and nothing else, with the target asserted to occur once before writing:

| mutant | result |
| --- | --- |
| `os.path.abspath` restored | `test_dotdot_through_a_symlinked_directory_does_not_swap_the_file` — 1 failed, 20 passed |
| entry's own name not re-resolved | `test_a_symlink_pointing_out_of_the_repo_is_not_committed` — 1 failed, 20 passed |

Reporting the `Path.resolve()` mutant as equivalent rather than counting it as killed is exactly right, and the kind of thing that makes the rest of the table believable.

## The vacuous assertion is worse than you said, and you were right to rip it out

`_committed_files` returns `.stdout`, a **str**, so the comprehension iterated characters:

```
type: str
[f for f in files if "outside" in f]  ->  []      # no single character contains "outside"
passes even with outside.py present:  True
```

It could not fail for any input. Moving to `_committed_names` is the fix.

## Verified

```
full suite                   12764 passed, 225 skipped, 2 xfailed, 0 failed
--strict alone               OK: 1608 file(s)
reanchor_citations --check   VERDICT=OK
security.md re-anchors       :1222 and :1558 both land on the git fetch calls they describe
```

CI's Python job is red only on `test_unmask_is_one_pass_not_one_per_table_entry`, which is the wall-clock flake (#349) and nothing to do with you — your branch passes it here. #351 fixes it and is approved; yours is the third PR it has reddened today.

## One thing to fix before this lands — it is three lines

`_within`'s docstring says two things the code does not do:

> `Path.is_relative_to` is 3.9+, and this is the same test `relative_to` performs above **without the exception round trip**.

The body *is* an exception round trip (`try: … except ValueError:`), and `requires-python = ">=3.12"` in `pyproject.toml`, so `is_relative_to` is available — I checked on the interpreter this repo runs:

```
python: 3.12.13
Path.is_relative_to available: True
```

So the stated reason for the helper existing does not hold. `return path.is_relative_to(root)` is the whole function, or keep the helper and correct the docstring. Also "string containment" describes `relative_to` a bit loosely — it compares path components, which is what makes it safe here, and that is worth saying accurately since it is the containment decision for a security control.

Behaviourally it is correct either way, which is why this is a nit and not a block — but this repo's most common defect is text asserting something the code does not establish, and a false claim inside the helper that decides containment is the wrong place to leave one. Happy to approve as-is on the understanding it gets tidied before landing.

## On the part you left out

Agreed on deferring the `chmod 000` parent. It is a different rule from containment and the audit placed it as pre-existing in both arms. File it or leave it to me, but do not fold it in here.

Clean as of `ef6a19ea`, expires on the next landing.

Issue #345, opened 2026-09-13
Case-fold probe measures the package's volume, not the worktree's — mixed-volume installs reopen every #328 cell


eyalgolan:
Follow-up to #339 (which is right, and should land — this is the half it doesn't reach).

`host_folds_case()` measures the case-sensitivity of **the volume the `exec_names` module lives on**, then applies that answer to commands that run somewhere else. When the package and the worktree sit on different volumes the two disagree, and one direction fails **open**.

## Measured

Case-sensitive APFS image mounted alongside the normal case-insensitive checkout. Positive control first — the volume really is case-sensitive:

```
$ touch csmnt/probe.txt && ls csmnt/PROBE.TXT
ls: .../csmnt/PROBE.TXT: No such file or directory
```

Same process, same `_folds_case`, two volumes:

```
_folds_case(module on case-INsensitive vol) -> True
_folds_case(module on case-sensitive   vol) -> False
```

End to end through `guard.evaluate(...)`, identical `cwd=/Users/eyalgolan/git/no_human-public` (case-insensitive, where commands actually run). The **only** difference is where the package was imported from:

| command | package on case-insensitive vol | package on case-sensitive vol |
| --- | --- | --- |
| `GH pr merge 7` | DENY | **ALLOW** |
| `RM -rf /` | DENY | **ALLOW** |
| `GIT push origin main` | DENY | **ALLOW** |
| `gh pr merge 7` | DENY | DENY |

So every cell #328 closed reopens on a mixed-volume install.

## Why this is reachable here

Worktrees live under `~/.no_human/worktrees/` while the installed package sits wherever the venv is. A developer keeping a case-sensitive volume for source (not unusual on macOS, where the default is case-insensitive) gets exactly this pairing. The reverse pairing — package case-insensitive, worktree case-sensitive — causes false denials, which is fail-closed and fine.

## The fix is cheap, because the guard already has the answer

`evaluate()` takes `cwd` and documents it as *"the session's worktree — where the backend will actually run the command"*. The probe just isn't asked about it. Either:

- probe `cwd` when it's provided, falling back to the module path when it isn't; or
- probe both and fold if **either** folds (fail-closed, and immune to the pairing question entirely).

I'd lean to the second: it can only over-deny, and over-denial here costs a contributor one confused command while under-denial costs a `RM -rf /`.

Worth a test that pins the property rather than the host — parametrise the probe's answer and assert the decision, so this stays checkable from either kind of machine.

PR #357, opened 2026-09-13
fix(guard): fold names on every host instead of measuring one volume

### LINKED ISSUE #347: venv install guard folds case on Windows only, so PIP install bypasses it on macOS
Surfaced while reviewing an external contributor's PR on the same theme in `agent/guard.py`. This one is in `agent/venv_install_guard.py`, it is live on macOS, and it is a bypass of the rule that a coder session must never install into the shared developer venv.

## Measured

```
  pip install evilpkg        -> DENY
  PIP install evilpkg        -> ALLOW
  Pip install evilpkg        -> ALLOW
```

Same fixture, same PATH (`<foreign venv>/bin:/usr/bin:/bin`), same `VIRTUAL_ENV`. The only difference is the case of the token.

And the shell really does resolve it — this is not a spelling nothing can execute:

```
$ printf '#!/bin/sh\necho RAN\n' > /tmp/foldprobe/bin/pip && chmod +x /tmp/foldprobe/bin/pip
$ PATH=/tmp/foldprobe/bin:$PATH PIP --version
RAN
```

macOS is case-insensitive by default, so `PIP` and `pip` are the same file and the install lands in exactly the venv the guard exists to protect.

## The cause

`_is_installer_name` folds case **on Windows only**, and its comment states the reasoning explicitly: on POSIX `PIP` is "a genuinely different file". That is true of a case-SENSITIVE filesystem and false of the default macOS one. The guard decides from `os.name`, i.e. from the platform class, rather than from the filesystem the command will actually run against.

## Why this is worth fixing properly rather than flipping a flag

The obvious patch — always fold on macOS — trades one assumption for another, and there is a live example of how that goes wrong. An external PR (#339) tried exactly this for the sibling guard and probed the filesystem by case-swapping its own `__file__`. That works in a source checkout and fails in the shipped artifact, because the desktop server is a PyInstaller freeze whose first rule is "NO .py FILES IN THE BUNDLE" — so `__file__` names a path that does not exist, the probe falls back to `os.name == "nt"`, and on macOS the fallback is the PERMISSIVE answer. Built as a real freeze, `RM -rf /` and `GH pr merge 7` both came back ALLOW.

So the constraints a fix here has to satisfy:

1. **Measure a path that exists at runtime.** `denial_reason` already receives `cwd` — the worktree the command will run in, which is on the right volume and exists in both the source tree and the frozen artifact.
2. **An unmeasurable probe must FOLD, not fall back to the platform class.** Folding is the fail-closed direction: it can only produce a false DENY, never a false ALLOW. Falling back to `os.name` on POSIX produces exactly the bypass above.
3. **Remember that command resolution is a PATH question, not a cwd question.** `/usr/bin` may fold when a developer's case-sensitive code volume does not. Measuring one directory and applying the answer to another is the defect in a different costume. If that cannot be resolved cleanly, folding unconditionally on the name path is defensible — the cost is a false DENY on a case-sensitive host for a spelling nobody types.
4. **Deny-direction only.** Verify the change with a before/after corpus through the real `evaluate`, and confirm zero rows move DENY -> ALLOW. The equivalent sweep on the sibling PR moved 10 rows ALLOW -> DENY and 0 the other way; that is the shape to reproduce.

## Testing note, because a source-tree suite structurally cannot see the failure mode above

Whatever probe is chosen, one test must exercise it under the FROZEN layout, not just the checkout. A test that imports the module from `src/` will pass while the shipped binary fails open. If that is impractical, the alternative is to make the probe's failure direction closed so that being wrong is harmless.

## Scope

Not folded into any current PR. The chmod-bypass fix that just landed in this module (`1d7c28a6`) is unrelated and should not be reopened for this.

### LINKED ISSUE #350: Case-folding probe measures __file__, which does not exist in the shipped freeze — guard reverts to assuming, in the open direction
Follow-up to #339, which is now merged. **This is not a request to revert it** — measured against the commit before it, #339 is a strict improvement, and reverting would reopen a live bypass:

```
BEFORE #339 (1d7c28a6)          AFTER (e5b07ac8)
  rm -rf /       -> DENY          rm -rf /       -> DENY
  RM -rf /       -> ALLOW         RM -rf /       -> DENY
  gh pr merge 7  -> DENY          gh pr merge 7  -> DENY
  GH pr merge 7  -> ALLOW         GH pr merge 7  -> DENY
```

Run from source, it closes the class it set out to close. What it does not close is the artifact users actually run.

## The gap

`agent/exec_names.py` measures whether the filesystem folds case by case-swapping **its own source file's name**:

```python
return _folds_case(os.path.realpath(__file__))
```

The shipped desktop server is a PyInstaller onedir bytecode freeze whose first stated rule is:

> `packaging/nh-server.spec:6` — `1. NO .py FILES IN THE BUNDLE.`

So at runtime `__file__` names a path that does not exist, `os.path.exists(swapped)` is False, and `_folds_case` falls back to `os.name == "nt"` → **False on macOS and Linux** → no folding.

Built as a real PyInstaller onedir freeze of `guard` + `exec_names` at that head, run on macOS:

```
FROZEN: True
exec_names.__file__: .../_internal/no_human/agent/exec_names.py   exists: False
host_folds_case(): False

  'RM -rf /'                     -> ALLOW
  'GH pr merge 7'                -> ALLOW
  'FIND / -delete'               -> ALLOW
  'sh -c "GIT push origin main"' -> ALLOW
  'rm -rf /'                     -> DENY    <- control
```

The bypass is live, not theoretical — macOS is case-insensitive by default and the shell resolves the capitalised name:

```
$ touch victim.txt && RM -f victim.txt
  RM DID run (file gone)
```

`GH pr merge 7` is the one to care about: `guard.evaluate` is the real PreToolUse veto (`agent/claude_backend.py`), and "the agent never merges" is a non-negotiable constraint.

## Two smaller things in the same area

**The fallback is the permissive answer, and the comment says the opposite.**

```python
except OSError:
    # An unreadable or vanished path proves nothing; fall back to the
    # host class rather than guessing the permissive answer.
    return os.name == "nt"
```

On macOS `os.name == "nt"` is False — don't fold — which *is* the permissive answer. That `except OSError` branch is also nearly dead: `os.path.exists` swallows `OSError` and returns False, so an unreadable path exits through an undocumented `exists()==False` path with the same permissive value and no comment. And `test_an_unmeasurable_path_falls_back_to_the_host_class` asserts `is (os.name == "nt")` while its docstring says "rather than guessing the permissive one" — the suite currently certifies the wrong behaviour on POSIX.

**The probe measures the wrong filesystem.** It measures the volume holding the *source*, not the volume the *command* runs against. On a real case-sensitive APFS volume holding the package, with cwd on the case-insensitive repo, all four rows reopen. Developers who keep code on a case-sensitive volume are exactly the people who care about this. Command resolution is also a PATH question — `/usr/bin` may fold when `/Volumes/dev` does not.

## What a fix needs to satisfy

1. Measure a path that **exists at runtime**. `evaluate` already receives the `cwd` the command will run in — it exists in both the checkout and the freeze, and it is on a relevant volume.
2. An unmeasurable probe must **fold** (fail closed). Folding can only produce a false DENY on a spelling nobody types; falling back to `os.name` produces the bypass above.
3. **Test under the frozen layout, not just the checkout.** This is the crux: a source-tree suite structurally cannot observe this failure. Every test passed while the shipped binary closed nothing. If a frozen-layout test is impractical, make the failure direction closed so being wrong is harmless.

## Related, and the reason this is worth doing once, properly

The sibling guard has the same defect from the opposite direction — see the issue on `venv_install_guard._is_installer_name`, where `pip install` denies and `PIP install` allows on macOS today. Both stem from deciding case behaviour from `os.name` rather than from the filesystem in play. Whatever probe is chosen here should be the one that fixes both.

L4XB:
Fixes #350
Fixes #347

My #339, and the sibling guard #347 names. Both come out the same way, so they are one change.

## What #339 got wrong

It answered "does this filesystem fold case" by **measuring** — case-swap this module's own filename, ask the OS whether it is the same file. Right in a checkout. Wrong in the artifact users run.

The desktop server is a PyInstaller freeze whose first rule is `NO .py FILES IN THE BUNDLE`, so `__file__` names a path that is not there, the probe proves nothing, and its fallback was `os.name == "nt"` — **False on macOS, the permissive answer**. Simulating that layout reproduces your table exactly:

```
host_folds_case(): False
  'RM -rf /'                     -> ALLOW
  'GH pr merge 7'                -> ALLOW
  'FIND / -delete'               -> ALLOW
  'sh -c "GIT push origin main"' -> ALLOW
  'rm -rf /'                     -> DENY    <- control

with this change:
host_folds_case(): True
  all five -> DENY
```

## Why the fix is "delete the probe", not "fix the probe"

Your constraint 3 is the one that settles it. Command resolution follows **PATH**, not cwd: `/usr/bin` may fold when a developer's case-sensitive code volume does not, so measuring one directory and applying the answer to another is the same defect wearing different clothes. `cwd` is a better path than `__file__` and still the wrong question.

So there is nothing left worth measuring, and folding always is the only position whose failure direction is closed:

- case-insensitive host: denies nothing that could not already run;
- case-sensitive host: can refuse `RM` and `GH` — spellings nobody types, each costing one message with a stated alternative.

Being wrong the other way costs a merge. That is the polarity this file already argues for itself.

`venv_install_guard._is_installer_name` had the identical Windows-only fold and the identical comment ("on POSIX `PIP` is a genuinely different file"), live on macOS. Same rule, same reason.

## Deny-direction sweep

Your constraint 4, through the real `evaluate`, 66 rows of capitalised spellings:

```
66 rows, 4 changed
  PIP install requests    ALLOW -> DENY
  PIP3 install requests   ALLOW -> DENY
  Pip install requests    ALLOW -> DENY
  Pip3 install requests   ALLOW -> DENY
DENY->ALLOW rows: NONE
```

The `RM`/`GH` rows do not move in that sweep because on this host the probe worked — which is exactly why the frozen simulation above is the evidence that matters, and why a source-tree suite structurally cannot see the failure mode. Your testing note called that; it is why one of the new tests sets `__file__` to a path that does not exist and asserts the closed answer comes back anyway.

## Tests

Three tests that pinned the probe are replaced, since they pinned the thing being removed:

- `test_names_fold_on_every_host` — asserts True whichever way the runner's filesystem answers, and derives what the volume actually does as a control so the assertion is independent of it rather than agreeing with it. (The old `test_the_probe_matches_this_host` would now fail on a case-sensitive runner, correctly: it asserted the probe, not the policy.)
- `test_folding_survives_a_file_that_is_not_on_disk` — the frozen shape.
- `test_the_probe_that_read_its_own_path_is_gone` — a shape guard: `ast.unparse` of the function body, docstring dropped, must be exactly `return True`. The docstring explains what the old probe read, so a textual grep would match its own prose; that caught me on the first run.

Plus `test_an_installer_name_folds_case_on_every_host` for #347, with a control that the rule still refuses names that are not installers, so it cannot pass by having collapsed into "always True".

Two mutants, both caught: `host_folds_case` back to a `__file__` probe (2 fail), and the venv guard folding on Windows only again (1 fail).

`-k "guard or exec_names or venv or install or case"`: 1136 passed, 23 skipped, 1 failed — `test_lane_conformance.py::test_the_js_implementation_agrees_on_every_shared_case`, which fails identically on a clean tree here (node reports `pass=None fail=None`).


eyalgolan (changes requested):
You've convinced me on the design, and you were right where I was wrong — my #345 asked you to probe `cwd` instead of `__file__`, and your answer is that both are the wrong question because resolution follows **PATH**, not any one directory. That is correct and I should have seen it. Deleting the probe is the right call.

Requesting one change, and it is a genuine defect rather than a style note.

## `@lru_cache` was orphaned onto `command_name`

On `main` that decorator belongs to `_folds_case`, whose input is one path. Deleting the function left the decorator in place, and it now caches `command_name` — whose input is an arbitrary command token:

```
        main:   109 case_flags   123 host_folds_case   150 @lru_cache  151 _folds_case   170 command_name
  this branch:  109 case_flags   123 host_folds_case   156 @lru_cache  157 command_name
```

Measured:

```
MAIN    command_name cached?  False
PR357   command_name cached?  True
        after 5000 distinct tokens -> currsize=5000, maxsize=None
```

The function is pure, so results stay correct — this is a resource leak, not a wrong answer. But `maxsize=None` means every distinct token the guard ever sees is retained for the life of the process, and the desktop server is long-lived and calls this on every guarded command. Unbounded growth keyed on agent-supplied strings, in a security hot path, is not something to ship by accident.

Just delete the stray decorator line. If you ever want `command_name` memoised, that is a deliberate change with a real `maxsize`, and its own reasoning.

## Everything else holds, and I checked it rather than reading the table

**#350 reproduces and is fixed.** Simulating the freeze — `__file__` naming a path that is not on disk:

```
MAIN    host_folds_case(): False    RM -rf / ALLOW   GH pr merge 7 ALLOW   FIND / -delete ALLOW
PR357   host_folds_case(): True     RM -rf / DENY    GH pr merge 7 DENY    FIND / -delete DENY
```

That is the artifact users actually run, and it was open. This is the important half of the PR.

**No DENY -> ALLOW anywhere.** 18 commands through the real `evaluate`, both trees, byte-identical results on this host — which is exactly what you predicted, since the probe happened to work here. Folding can only widen what matches a guarded name, so the direction is closed by construction, and the sweep agrees.

**`.lower()` is safe in the venv guard:** `_EXACT_INSTALLERS` is 6 entries, none non-lowercase, so folding cannot break an existing match.

**`test_the_probe_that_read_its_own_path_is_gone`** is a nice piece of work — a shape guard via `ast.unparse` with the docstring dropped, because the docstring explains the old probe and a textual grep would match its own prose. Noting that it caught you on the first run is the kind of detail that makes the rest of a PR's claims credible.

## On the polarity argument

Agreed, and for the record I think it is the right trade stated the right way: on a case-sensitive host this can refuse `RM` or `PIP` — spellings nobody types, each costing one message with a stated alternative — and being wrong in the other direction costs a merge. That is the polarity this file already argues for itself, and it is better than the position I landed in #339.

Fix the decorator and I will approve immediately; there is nothing else outstanding.

One process note, no action needed: `main` has moved (`859ad6a7`) since you opened this, so you will want to pick that up — `RELEASE_MANIFEST.txt` will conflict, and the recipe is to take main's copy wholesale and regenerate rather than merge the two sides.

eyalgolan:
Heads up before you rebase — `main` moved under this one in a way that touches your exact file, so it is worth reading before you resolve.

`d4b0f47c` ("The host case-fold probe is measured once per process, not per token") landed a `@lru_cache(maxsize=1)` on `host_folds_case` itself. Its reasoning is worth reading, because it is a better root cause than the one I gave for the flake in #349: the inner `_folds_case` memo was returning the right *answer* (4008 hits to 1 miss), but `os.path.realpath(__file__)` ran **before** the memo and lstat'd every path component on every call. A 4000-argument command drove 4007 probes through one `evaluate`, and `guard.evaluate` went 0.132s to 0.036s once cached. I had put that flake down to a loaded runner; the measurement says most of it was our own syscalls.

So `main` now has two decorators in that file:

```
123: @lru_cache(maxsize=1)      host_folds_case
170: @lru_cache(maxsize=None)   _folds_case
190:                            command_name
```

Three consequences for this PR, none of them fatal:

1. **It conflicts now** — `exec_names.py`, `tests/test_exec_names.py` and the manifest.
2. **Your change makes both caches pointless.** `host_folds_case` becomes `return True`; caching a constant is harmless but it is dead weight, and `_folds_case` disappears entirely. I would delete both decorators and drop the `lru_cache` import if nothing else in the file uses it.
3. **The defect I flagged gets easier to trip during the rebase, not harder.** Both decorators sit immediately above a function you are deleting or emptying. Deleting the function under a decorator leaves the decorator attached to whatever follows — which is exactly how `@lru_cache(maxsize=None)` ended up on `command_name` the first time. Worth re-running that check after you resolve:

```
python -c "from no_human.agent import exec_names as e; print(hasattr(e.command_name,'cache_info'))"
```

`False` is the answer you want.

Everything else in my review stands — the frozen-artifact fix is the important half and it is right. Sorry about the moving target; that churn is on us, and I have filed #358 about how often it is happening.

L4XB:
Fixed, and thank you for measuring it rather than reading the diff — I would not have found it, because the decorator was never written, only left behind.

```
before   command_name cached?  True    currsize=5000 after 5000 distinct tokens
after    command_name cached?  False
```

That is the whole class of thing that makes a deletion dangerous: the line I removed was `def _folds_case(...)`, and the line above it stayed. Nothing in the diff draws the eye to a decorator whose function has gone, and nothing would have drawn the next reader's either — so there is now a test for it:

```
✓ test_command_name_is_not_memoised_on_agent_supplied_tokens
```

Putting `lru_cache` back on `command_name` reddens it. It is deliberately phrased as "is not memoised" rather than "the decorator is absent", so if memoisation is ever wanted it has to arrive as a real decision with a real `maxsize` and pass a changed test, not slide in as leftovers.

Everything else is unchanged and re-verified on this head: 131 passed across `test_exec_names.py` + `test_venv_install_guard.py`, and the frozen simulation still gives DENY on all five rows.

On the design: I only got there because #345 made me look at `cwd`, and looking at `cwd` is what made "which directory?" the obvious wrong question. The disagreement did the work.

eyalgolan:
Reviewed properly, and the mechanism is **right**. Both bugs really are fixed — I reproduced the frozen-artifact table independently and watched it close. This should land, but not as it currently stands: it conflicts with `main` in both files under review, and three shipped claims are false. All of it is cheap to fix.

## First, the thing that makes this urgent

I confirmed on current `main` (`680d6889`), simulating the frozen artifact by pointing `__file__` at a nonexistent path *before* `guard` compiles its case-flagged regexes:

```
host_folds_case() = False
  rm -rf /         -> DENY
  RM -rf /         -> ALLOW
  gh pr merge 7    -> DENY
  GH pr merge 7    -> ALLOW
```

`GH pr merge 7 -> ALLOW` in the shipped desktop freeze is a live bypass of a non-negotiable project constraint. Your PR closes it. That matters more than everything below.

## How to resolve the conflict — please do not "keep both"

`main` now has my `d4b0f47c`, which put `@lru_cache(maxsize=1)` on `host_folds_case` and added `test_the_host_probe_is_measured_once_not_once_per_token`. That test sits inside your conflict hunk.

**Both of mine should be deleted in your resolution, not merged.** Here is why, measured:

```
# with host_folds_case() == True (your version)
realpath calls ending in exec_names.py: 0   (my assertion is <= 1)
-> asserts 0 <= 1 : PASSES, and can NEVER fail
```

My fix made a *wrong probe fast*. Yours *removes the probe*. Once `host_folds_case` is a literal `return True` there is no filesystem read left to cache and nothing for my test to observe — the cache is dead weight and the test is a green assertion that pins nothing. Keeping it would reintroduce exactly the defect class that let the original bug ship. Your change supersedes mine; that is the correct outcome and I would rather lose my commit than keep a vacuous test.

## Blocking

**1. The #347 fix is unpinned end to end.** The only new test asserts the private helper `_is_installer_name`, in a file whose own docstring says its tests are verdict-derived and which calls `guard.evaluate` 65 times. One-line mutation at `venv_install_guard.py:679`:

```python
-        if not _is_installer_name(_basename(token)):
+        if not (_is_installer_name(_basename(token)) and _basename(token) in _EXACT_INSTALLERS):
```

Result: `pip install evilpkg -> DENY`, `PIP install evilpkg -> ALLOW` — the bypass fully reopened — and `tests/test_exec_names.py tests/test_venv_install_guard.py` → **132 passed**. Nothing observes the POSIX `PIP install evilpkg` verdict, which is the literal subject of #347. Add one end-to-end test driving it through `denial_reason` with the real venv fixture.

**2. False closed set in shipped source.** `exec_names.py:147-149` says that on a case-sensitive host it "can **only refuse `RM` and `GH`** — spellings nobody types". The case-sensitive arm also flips `FIND / -delete`, `GIT push origin main`, `sh -c "GIT push origin main"`, and `rm -Rf` / `rm -RF` / `rm -fR`. Six more members than the enumeration admits. The flips are *correct* — `rm -Rf build` was a live bypass on case-sensitive hosts — the sentence is not. Writing a closed set into a comment is a documented repeat defect here; either drop the enumeration or make it true.

**3. Stale prose left standing.** You rewrote `host_folds_case`'s docstring and left `case_flags()` ten lines above untouched: *"`re.IGNORECASE` where the host folds case, else no flag … Same measurement, same reasoning: folding where the filesystem folds…"*. There is no measurement any more, folding no longer tracks the filesystem, and the `else 0` branch is unreachable. This is the same shape as #339's "fall back to the host class rather than guessing the permissive answer".

**4. The new test's "Control" is a tautology.** In `test_names_fold_on_every_host`:

```python
assert volume_folds in (True, False)
```

`volume_folds` is the result of `Path.exists() and os.path.samefile(...)` — always a bool, so this can never fail, and the assertions above never reference it. The PR body repeats the claim that it makes the assertion independent. Either assert something real about the volume or drop it and the sentence.

## Non-blocking

- `import os` at `exec_names.py:89` is now dead — an AST scan finds no `os.` use outside docstrings. You removed the other dead import (`lru_cache`) and missed this one.
- `test_the_probe_that_read_its_own_path_is_gone` is an AST-over-source guard that requires the body to unparse to exactly `return True`. A behaviour-preserving refactor (`folds = True; return folds`) turns it red. Safer than a regex, still the class of guard that has cost this repo a lot of review rounds.

## Verified and clean, so nobody re-litigates it

Regression direction is genuinely safe: with a **control arm run first** (96 rows, 0 diffs, so the arms agree where expected), then 139 rows across two corpora — **9 changed, all ALLOW→DENY, zero DENY→ALLOW**. None of 30+ legitimate capitalised binaries flipped (`R`, `Rscript`, `GCC`, `CMake`, `Cargo`, `Xvfb`, `Node`, `Python3`, `Git status`, `LS`, `git clean -Xdf`, `./scripts/Git-Push-Helper.sh`). Tamper-clean: tests 130→132, assertions 168→173, and all 87 removed lines accounted for. Fail-closed confirmed — no `__file__` on the decision path. Whole-tree gates pass, manifest `OK: 1608`, citations `VERDICT=OK`.

One note: `agent/guard.py` carries a stale-high frozen pin (2925 actual vs 2926 frozen) from an earlier change. You neither introduced nor corrected it — leave it; it has its own issue.

## One request

Please don't self-merge this one. Five PRs went in tonight without review; one of them (#351) removed the wall-clock test that was catching a live 2× regression, so trunk went green while the regression stayed. This PR fixes a real security bypass and I want it in — but the resolution of the conflict above is a judgement call that needs a second pair of eyes, and every commit here has to carry `eyalgolan` as committer, which a GitHub squash cannot do. Rebase, resolve as described, and I will review and land it by hand with you as author.

L4XB:
Synced with `main`, which landed the caching fix for this same function while this was open. Three real conflicts, and two of them are decisions rather than mechanics, so here is what I did with each.

**`@lru_cache(maxsize=1)` on `host_folds_case` — removed.** `main` added it so the `realpath` would not run per token. With the probe deleted the function returns a constant, so there is no syscall left to spare, and a memo on a constant is the same orphaned decorator you asked me to remove one of in this PR. Removing it also keeps the module importable: this branch had already dropped the `from functools import lru_cache` import, so carrying the decorator across would have been a `NameError` at import.

**`test_the_host_probe_is_measured_once_not_once_per_token` — kept.** It grades the property, not the mechanism, so it passes here for a stronger reason: the count is 0, not 1. I kept it as the guard against anyone reintroducing a per-token filesystem question. I did edit its docstring, because it named `_folds_case` and the memo, both of which this PR deletes — the measurement history you recorded (0.132s vs 0.036s, the 0.4s bound on a shared runner) is preserved verbatim.

**`test_the_probe_matches_this_host` — deleted.** It asserts `host_folds_case() is really_folds`, which is the position this PR argues against and you agreed to drop. `test_the_probe_measures_the_volume_it_is_asked_about` already covers the same ground correctly: it still measures what the volume does, and then asserts the answer is `True` regardless — with the measurement kept as an explicit control so the assertion cannot pass by agreeing with the host.

`pytest tests/ -k "guard or exec_names or structural"` — 975 passed, 13 skipped.

eyalgolan (approved):
Approving. The decorator is gone and the check comes back the way it should:

```
decorators in exec_names.py:  none
command_name cached?          False
lru_cache still used at all?  False   (import dropped too)
```

Dropping the import rather than leaving it unused is the right finish.

## Verified, not read

**#350 — the artifact users actually run.** Simulating the freeze, `__file__` naming a path that is not on disk:

```
MAIN    folds=False   RM -rf /=ALLOW  GH pr merge 7=ALLOW  FIND / -delete=ALLOW  sh -c "GIT push origin main"=ALLOW
PR357   folds=True    RM -rf /=DENY   GH pr merge 7=DENY   FIND / -delete=DENY   sh -c "GIT push origin main"=DENY
```

**#347 — the venv guard**, through `denial_reason` with a real foreign venv on PATH and `VIRTUAL_ENV` set, with both controls:

```
                      MAIN      PR357
pip install evilpkg   DENY      DENY     <- control: the guard works at all
PIP install evilpkg   allow     DENY
Pip install evilpkg   allow     DENY
git status            allow     allow    <- control: not everything got denied
```

**No DENY -> ALLOW.** 15 commands through the real `evaluate` on both trees, no verdict moved in the permissive direction.

```
full suite   12776 passed, 225 skipped, 2 xfailed, 0 failed
guard suites 400 passed
--strict     OK: 1609
citations    VERDICT=OK
merge        0 conflicts
```

## One thing I chased and want to record as a non-finding

A sweep of mine showed `pip install requests` flipping to DENY on your branch once. It does not reproduce — 17 consecutive ALLOWs in-process and across fresh processes — and I am confident it was my own instrument: I ran that sweep while a full suite was creating and tearing down venvs on the same machine, and `_resolve_installer` walks the real PATH. Mentioning it so that if you ever see a stray verdict from this guard you know it can be environmental rather than logical.

## On the design

You were right and I was wrong. My #345 asked for a `cwd` probe; resolution follows PATH, so any single-directory measurement answers a question nobody asked. Folding always is the position whose failure direction is closed, and the `ast.unparse` shape guard against the probe coming back is a nice touch — especially the detail that a textual grep would have matched the docstring's own prose.

eyalgolan (changes requested):
I have to walk back my approval here, and the reason is entirely on our side rather than yours.

While this sat approved, `c176f291` ("The case-fold probe answers from a path that exists at runtime") landed on main. It solves the same two issues by the other route — keeping the probe but making it read paths that exist at runtime (PATH entries and cwd, with `_MAX_ENTRY_PROBES` / `_MAX_PATH_PROBES` caps) instead of `__file__`. Measured on current main:

```
frozen simulation (__file__ absent):  host_folds_case() -> True
  RM -rf /         DENY
  GH pr merge 7    DENY
  NH approve 7     DENY

venv guard, foreign venv on PATH:
  pip install evilpkg   DENY
  PIP install evilpkg   DENY     <- #347 closed too
  Pip install evilpkg   DENY
  git status            allow    (control)
```

So both halves of this PR are already delivered, and merging it now would not add a fix — it would **replace** a landed implementation with a different one. That is a decision that deserves to be made deliberately, not as a side effect of an approval I gave before the other change existed. It also conflicts on four files now, source included:

```
RELEASE_MANIFEST.txt  src/no_human/agent/exec_names.py
tests/test_exec_names.py  src/no_human/agent/venv_install_guard.py
```

## For the record, because I think you should get the credit

Your diagnosis is what made this visible: `__file__` does not exist in a PyInstaller freeze, the fallback was `os.name == "nt"`, and that is the permissive answer on macOS. You built the freeze and measured it. Whatever implementation ends up in the tree, that finding is yours, and #350 exists because you wrote it up properly.

You also changed my mind on the design question — my #345 asked for a `cwd` probe and you were right that resolution follows PATH, so any single-directory measurement answers the wrong question. The landed fix ended up probing PATH entries, which is your argument taken seriously.

## Where that leaves this PR

My honest read is that it should close, and I would rather you hear the reasoning than just get a "superseded" label:

- **"Always fold"** (yours) is simpler and has no probe to get wrong. Its cost is over-denial on a genuinely case-sensitive host — `RM` and `PIP`, spellings nobody types.
- **"Probe PATH and cwd"** (landed) preserves exact behaviour on case-sensitive hosts at the cost of real machinery, including two probe caps and a documented residual where `guard.py`'s regexes compile at import time before any `cwd` is known.

I argued for yours in my review and I still think it is the better trade. But it is in the operator's gift to reopen that, not something to settle by merging a conflicting branch, and the landed version is not wrong — it is just more expensive.

If you would rather make the case for replacing it, say so and I will put it in front of him properly with both measurements side by side. Otherwise I will close this with the credit recorded on #350. Either way, nothing here was wasted: the freeze measurement drove the fix that landed.

L4XB:
Closing it. You are right that this is not mine to settle by merging, and I would rather not spend your operator's attention on a swap that buys simplicity and costs exact behaviour on case-sensitive hosts. `c176f291` closes both issues — you measured the frozen case and the foreign-venv case, and that is the part that mattered.

For what it is worth I no longer think "always fold" is clearly the better trade, now that the landed version exists rather than being hypothetical. My argument for it was partly that the probe was hard to get right, and the PATH-entry version answers that: it probes where resolution actually looks, with both caps bounded, and `host_folds_case` cached per `(cwd, path_env)` so a long PATH is paid once. The over-denial I was willing to accept is a real cost that this avoids.

One thing I would rather hand over than let sit, since you named it yourself: the residual is executable, not just documented.

```python
# guard.py:160, at import time
_RM_RF = re.compile(r"\brm\s+(-[a-z]*r[a-z]*f|...)\b", exec_names.case_flags())
```

`case_flags()` resolves `host_folds_case(cwd, path_env)`, so these module-level patterns are frozen with the answer for whatever directory the process happened to start in. A long-lived desktop server that starts on a case-sensitive volume and later guards a command whose `cwd` is a case-insensitive one keeps the import-time flags — `venv_install_guard` asks per call (`:407`) and gets the right answer, while `guard.py`'s regexes cannot. It is narrow and it is not what #350 was about, but it is the same failure direction: the permissive answer on the host where the fold matters.

If that is worth closing, say the word and I will open a separate PR that compiles those patterns lazily per `(cwd, path_env)` and pins it with a cell that flips the fold between two evaluations in one process. Small, additive, and it does not touch the implementation you just landed. If you would rather it stay a register entry, that is fine too — you have the context I do not.

Thanks for the write-up on #350, and for arguing the design on its merits rather than on who filed what.

PR #360, opened 2026-09-13
fix(runner): rewrite a bare pytest on dash and busybox, not only on bash

L4XB:
Split out of #355 at the maintainer's request — it is an independent production fix that was riding in a `test(...)`-titled PR, and it deserves its own line in a squash-merged log.

## The defect

`_fix_invocation`'s class-3 branch matched bash's phrasing only:

```python
or ("pytest" in out_lower and "command not found" in out_lower)
```

**dash** is `/bin/sh` on Debian and Ubuntu — so on most Linux CI runners and in most containers — and it says `sh: 1: pytest: not found`. busybox says `sh: pytest: not found`. Neither contains "command not found", so the rewrite never fired there.

Measured against the real messages, before the fix:

```
bash/zsh                  detected=True  rewrite=<python> -m pytest -q
dash (Debian/Ubuntu sh)   detected=True  rewrite=None
busybox sh                detected=True  rewrite=None
```

`_is_invocation_error` never had this problem — it returns True on zero counts — so the launch failure was **detected** everywhere and **repaired** nowhere. That is why it surfaces as an honest "no test evidence" rather than a crash, and why nobody noticed: a project whose detected command is a bare `pytest` gets the retry on a developer's mac and no retry on the runner.

## How it was found

By #353's new assertion. The base-tree test's own docstring says it "reproduces the rewrite end-to-end", and on CI it did not — the recorded commands were `['pytest -q', 'pytest -q']`, two bare invocations with no rewrite between them. That is what sent me looking here.

## The rule

Anchored on the shell's `<name>: [command ]not found` shape at end of line. Two negative controls are pinned with it: pytest's own `ERROR: file or directory not found: tests/x.py` has no colon before the phrase, and `sh: 1: vitest: not found` is a different tool. Reverting to the bash-only form reddens the new test.

One narrowing, which the review found and I want carried across rather than left in a thread: the old rule matched `pytest` and `command not found` **anywhere** in the output; this one requires them on the same line. The only regressing input is one no real shell emits — bash always names the command on the same line as the phrase.

`tests/test_runner.py`: 91 passed.


eyalgolan (dismissed):
Approving. Thank you for taking the relabel seriously — `fix(runner):` is the honest title, and the body now leads with the production defect instead of burying it.

## Both halves verified by mutation, not by reading

**The runner fix is load-bearing.** Your tests against *main's* runner:

```
AssertionError: ('dash', None)
assert None == '<python> -m pytest -q'
```

dash is `/bin/sh` on Debian and Ubuntu, so this was costing test evidence on most Linux CI and nowhere else — which is exactly why it went unnoticed.

**The base-tree test now grades what it is named for.** Both mutants red, including the one your first attempt missed:

```
class-3 returns None            -> test_pre_existing_red_test_excused... FAILED
rewrite drops `rest` (node ids) -> test_pre_existing_red_test_excused... FAILED
```

That second one is the one that matters. A rewrite dropping the node ids is a different request wearing the same name, and it is the shape the original send-back turned on.

```
full suite   12776 passed, 225 skipped, 2 xfailed, 0 failed
--strict     OK: 1609
citations    VERDICT=OK
merge        0 conflicts
```

## The discovery is the best part of this PR

> By #353's new assertion. The recorded commands were `['pytest -q', 'pytest -q']`, two bare invocations with no rewrite between them.

A test written to grade a behaviour immediately caught a real production bug on a platform none of us runs locally. That is the whole argument for #353 having been worth doing, and it is worth saying out loud.

## One bookkeeping ask, not a blocker

This PR still carries the `test_base_tree_gate.py` work, which *is* the fix for #353 — I confirmed the mutants above against this branch, and #353 is still open with no `Fixes` link. So the same thing that bit #355 is half-present in reverse: the title is right for the runner fix, and the test work rides along unnamed.

Nothing to change in the code. Either add `Fixes #353` to the body, or say the word and I will close #353 against this when it lands. I would rather the log recorded both, since someone looking for "when did that test start grading the rewrite" should land here.

Also carried across correctly: the same-line narrowing I raised on #355. You are right that the only regressing input is one no real shell emits, and I like that you put it in the body rather than leaving it in a review thread where it would have been lost.

eyalgolan:
Thanks for this — the dash/busybox fix is real and it is properly pinned. I reproduced it against messages captured from live `debian:stable-slim`, `ubuntu:24.04`, `alpine:3` and from `/bin/dash` and `/bin/ksh` on macOS, running the two trees side by side and printing `no_human.__file__` in each arm so I know the arms are not reading the same tree:

```
label                   | main    | this PR | verdict
CONTROL-must-rewrite    | REWRITE | REWRITE | same
CONTROL-must-be-None    | None    | None    | same
macOS /bin/sh           | REWRITE | REWRITE | same
macOS /bin/bash         | REWRITE | REWRITE | same
debian+ubuntu /bin/sh   | None    | REWRITE | FIXED
macOS /bin/dash         | None    | REWRITE | FIXED
alpine busybox /bin/sh  | None    | REWRITE | FIXED
busybox 'line 0' form   | None    | REWRITE | FIXED
macOS /bin/ksh          | None    | REWRITE | FIXED
zsh                     | REWRITE | None    | REGRESSION
csh / tcsh              | REWRITE | None    | REGRESSION
```

Full suite on your head is green (`12777 passed, 222 skipped, 2 xfailed`, exit 0), the whole-tree gates pass, the CLA ledger entry is in place and untouched, and there is no tamper (`test_base_tree_gate.py` 11→11 tests / 39→44 asserts, `test_runner.py` 89→91 / 172→174). The stale base is harmless — `git merge-tree --write-tree` against current main is clean and the gates pass on the merged tree.

Two things need changing before it can land.

**1. zsh is named as covered, and it was covered before this change.**

`src/no_human/testing/runner.py:901` says "bash and zsh say `command not found`", and `tests/test_runner.py:418-421` carries the same claim in a test named `test_bare_pytest_is_rewritten_whatever_the_shell_calls_it`. The new regex requires `pytest` to appear *before* the `: … not found` at end of line. zsh puts the command name last, and csh/tcsh append a period. Real output from this machine:

```
sh      => /bin/sh: pytest: command not found
bash    => /bin/bash: pytest: command not found
dash    => /bin/dash: 1: pytest: not found
zsh     => zsh:1: command not found: pytest
ksh     => /bin/ksh: pytest: not found
csh     => pytest: Command not found.
```

The PR body says "the only regressing input is one no real shell emits — bash always names the command on the same line as the phrase." Two real shells on this host regress, and one of them is the shell the source comment claims to cover. Severity is limited because `_run_shell` uses `Popen(shell=True)`, which always uses `/bin/sh` regardless of `$SHELL` (I checked — with `SHELL=/bin/zsh` the wording is still `/bin/sh: …`), so zsh text only reaches `_fix_invocation` if a project's own test command spawns zsh. But the comment, the test name and the PR body each assert coverage the code does not have.

A pattern that covers all ten captured real messages and still rejects the negative controls:

```python
_SHELL_NOT_FOUND = re.compile(
    r"^(?:.*\bpytest\b.*?:\s*(?:command\s+)?not found"   # sh/bash/dash/ksh/busybox
    r"|.*:\s*command not found:\s*\S*\bpytest\b)"        # zsh puts the name last
    r"[.\s]*$",                                          # csh/tcsh append a period
    re.IGNORECASE | re.MULTILINE,
)
```

Either widen it like that, or drop zsh from the comment and rename the test to something the code satisfies.

**2. The negative controls do not exercise what they are credited with.**

`runner.py:908-910` credits the end-anchor with excluding pytest's own `file or directory not found: tests/x.py`, and the PR body says "two negative controls are pinned with it". All three rows at `tests/test_runner.py:449-451` contain no `pytest` token at all, so `\bpytest\b` rejects them before the anchor or the colon is ever consulted. Mutation, one change each, against `tests/test_runner.py`:

| change | result |
|---|---|
| drop the end anchor | `91 passed` — survives |
| drop the colon | `91 passed` — survives |
| drop `(?:command\s+)?` (positive control) | `1 failed` — caught |

The control going red is what proves the harness can see a broken regex, so those two greens are real survivals. Rows that would actually discriminate: `pytest: plugin pytest-xdist: not found, install it` (the anchor excludes it) and `collected 0 items / pytest plugins not found` (the colon excludes it). Worth fixing the comment too — `\bpytest\b` is what excludes the cited example, not the anchor.

Smaller: `tests/test_base_tree_gate.py:467` lost the two blank lines between two top-level functions, and `RELEASE_MANIFEST.txt` will need regenerating after the above.

One note for whoever picks it up next, not something to fix here: the same bash-only assumption survives four lines below, at `runner.py:920` (`cmd.startswith("python ") and "command not found" in output`), and in `_INVOCATION_ERROR_PATTERNS` at `runner.py:873`. Both are dash-blind in exactly the same way. Same root cause, separate change.

eyalgolan (approved):
Re-approving at `fb330950`. The reset was the main merge; your own work is untouched:

```
src/no_human/testing/runner.py   IDENTICAL
tests/test_base_tree_gate.py     IDENTICAL
tests/test_runner.py             IDENTICAL
--strict alone                   OK: 1614 file(s)
citations                        VERDICT=OK
```

Everything I verified the first time stands: the dash/busybox rewrite is load-bearing (your tests return `AssertionError: ('dash', None)` against main's runner), and both mutants on the base-tree test go red, including the one that drops the node ids.

Still the one open thread from my last review: this carries #353's fix without a `Fixes #353` line, so that issue would stay open when this lands. Say the word and I will close it against this commit — no need to push again just for that.

Approved as of `a8a04496`.

eyalgolan (approved):
Re-verified at `c6f3680c` and still good — the delta since my last review is purely your merge of main, and your three files are byte-identical:

```
src/no_human/testing/runner.py   IDENTICAL
tests/test_base_tree_gate.py     IDENTICAL
tests/test_runner.py             IDENTICAL
```

I re-ran the evidence rather than assuming it carried, because main landed guard and case-fold work that the base-tree test exercises through the orchestrator:

```
full suite                     12865 passed, 227 skipped, 2 xfailed, 0 failed
--strict alone                 OK: 1619 file(s)
citations                      VERDICT=OK
main is an ancestor            YES

mutant: class-3 returns None            1 failed   (caught)
mutant: rewrite drops `rest`            1 failed   (caught)
control: clean branch                   1 passed
your tests vs CURRENT main's runner     1 failed   (still load-bearing)
```

That last line is the one worth re-checking after a rebase: if main had independently fixed the dash phrasing, your change would have quietly become a no-op and the PR would be landing dead code. It has not — `AssertionError: ('dash', None)` against main's runner, same as before.

Still the one open thread, and it is bookkeeping rather than code: this carries #353's fix (the base-tree grading) with no `Fixes #353` line, so that issue stays open when this lands. Say the word and I will close it against this commit — no need to push again just for that.

Clean as of `49eaa906`.

eyalgolan:
Landed as `c388aaf5`, with you as the commit author.

PR #379, opened 2026-09-14
fix(db): give task config its own writer so a stale handle cannot revert it

### LINKED ISSUE #343: A watcher tick with nothing to do silently reverts a human's budget raise
`nh task config <id> lifetime_tokens=N` is the documented human-only escape from a `BUDGET_EXHAUSTED` park. A watcher tick can silently undo it, leaving no trace, and the operator finds out only when the task dies at the old cap.

## What happened

I raised task `9b6e928a` from 10,000,000 to 14,000,000 weighted and the CLI confirmed it (`applied lifetime_tokens=14000000`). Some minutes later the stored config read:

```
{"attempt_tokens": 4000000, "lifetime_tokens": 10000000, "budget_unit": "weighted"}   updated=2026-09-13T16:13:01Z
```

The raise is gone. The only task event at that exact second is:

```
19:13:01  wake_tick: watcher checked (awaiting_approval): nothing to do
```

Two other tasks I raised in the same window (`0ff9125c` → 16,500,000; `4135165f` → 14,000,000 + `lifetime_attempts`) kept their raises, so this is not the CLI failing to write.

## The mechanism — confirmed in code, not inferred from the incident

`src/no_human/core/db.py:2195` `update_task_columns` writes:

```sql
UPDATE tasks SET
  ...
  title = CASE WHEN COALESCE(json_extract(COALESCE(context,'{}'),'$.title_updated_at'),'')
               > :handle_title_marker THEN title ELSE :title END,
  ...
  priority=:priority, plan=:plan, config=:config,
  updated_at=:updated_at
WHERE id=:id
```

Its own docstring states the problem it was written to solve:

> Persist the task's mutable columns EXCEPT context and status. Multi-writer zones (watcher, CLI, gate) must write context only via merge_context/append_context_list — this companion writes the rest **without clobbering concurrent context merges with a stale blob.**

and for `title`:

> a stale handle must not revert a `nh task retitle` that landed while this handle was in flight

So three columns are explicitly defended against a stale handle — `context` and `status` by exclusion, `title` by a marker-keyed CASE. **`config` is written wholesale from whatever the handle loaded.** The watcher is named in that docstring as one of the multi-writer zones, and `blockers/wake.py` calls this method from 14 sites, most of which have no interest in `config` at all — including paths that conclude "nothing to do".

The race needs no unusual timing: the watcher loads a task, a human runs `nh task config` while the tick is in flight, and the tick's save writes the pre-raise config back.

## Why this one matters more than an ordinary lost update

- The budget cap is the ONE setting the design deliberately reserves to a human. `BUDGET_EXHAUSTED` parks specifically so a person can decide whether to spend more. Silently reverting that decision converts an explicit human judgement into a no-op.
- It is invisible. The tick logs "nothing to do", the CLI already reported success, and nothing records that a value changed. The operator's next signal is the task dying at a cap they believed they had lifted.
- It is self-concealing on retry: raise again, and the next tick can revert it again.

## Fix shape

`title` already demonstrates the pattern in this very statement. Either:

1. **Exclude `config`** from `update_task_columns` and give it a dedicated writer, the way `status` has `set_status` and `context` has `merge_context`. Nothing that calls this method from the watcher appears to intend a config write.
2. **Guard it like `title`** — stamp a `config_updated_at` marker when the CLI writes config, and make the CASE keep the row's value when the row's marker is newer than the handle's.

(1) is cleaner and matches the docstring's own stated architecture: `config` belongs in the "multi-writer zone" list beside `context` and `status`, not in the "rest".

## Also worth checking while in here

Whether any other multi-writer column has the same exposure. The audit question is not "does this method write X" but "is X written by more than one zone, and does the CASE/exclusion list name it".

L4XB:
Closes #343.

## The fix shape

Option (1) from the issue, and for the reason the issue gives: the docstring already names `config` as belonging in the multi-writer zone beside `context` and `status`, and nothing in the watcher intends a config write.

`config` leaves the column list of `update_task_columns` and gets one writer, `update_task_config` — a targeted single-column `UPDATE`, shaped like `update_task_title`.

**It leaves `update_task` too**, which the issue does not ask for. The same wholesale write is there, and an orchestrator handle lives for an entire attempt — a far longer window than a tick — so a cap raised mid-run was reverted the same way. Leaving it would also have made the new docstring false: `update_task_config` is the only writer, or it is not. If you would rather have that as a separate PR, say so and I will split it.

## The audit the issue asks for

> The audit question is not "does this method write X" but "is X written by more than one zone, and does the CASE/exclusion list name it".

For `config` the complete set of in-memory writers in `src/` is:

| site | persisted by | affected |
| --- | --- | --- |
| `cli/commands.py` task create, `api/app.py` task create, the four intake pollers, slack intake | `create_task` (INSERT) | no |
| `blockers/actions.py::apply_action` — `set_task_config`, the only verb that writes config | `update_task_columns` | **yes**, 3 call sites |

Found by grepping every `.config =`, `.config[...] =`, `.config.update/setdefault/pop/clear` and `del …config[…]` in `src/`. The three `apply_action` call sites — `nh task config`, the CLI blocker reply, the API blocker answer — are wired to the new writer.

Each write sits **beside the `apply_action` that produced the mutation**, before the branch below it, rather than inside one of the branches — so a branch added later cannot silently drop the raise. (My first cut put the API one inside the terminal "keep parked" branch. That branch is unreachable for a config write — `apply_action` refuses `{park, set_task_config}` together — so it was dead code, and the live path lost the write. `test_api.py::test_reply_choose_applies_the_action_and_resumes_from_checkpoint` caught it.)

## Tests

`tests/test_config_clobber.py`, 8 cells: the reported incident, the same race through `update_task`, repeated stale write-backs, accept controls that both writers still write the columns they own, and that the new writer neither carries a stale row back nor leaves an empty config unreadable.

Every mutation is caught, each by a different cell:

| mutation | red |
| --- | --- |
| put `config` back into `update_task_columns` | 2 |
| put `config` back into `update_task` | 2 |
| make `update_task_config` a no-op | 6 |
| drop its `updated_at` bump | 1 |

And the three call sites are covered by suites that already existed — each reds when its wiring is removed: `test_task_config_cli.py` (5 red), `test_cli_commands.py::test_reply_choose_applies_the_options_action`, `test_api.py::test_reply_choose_applies_the_action_and_resumes_from_checkpoint`.

`test_send_back_refused.py` used `update_task` to persist a budget raise in its setup and now uses the new writer. It was the only caller in the tree that relied on config travelling with the row.

## Ratchets

- The three frozen file budgets are re-anchored on this tree, each with the reason, per this file's own convention.
- `cli/commands.py` grew by 11 lines, which moved six doc citations; re-anchored with `uv run python scripts/reanchor_citations.py --apply`, plus one `docs/WINDOWS.md` `file:line` the script does not cover (`signal.SIGKILL`, 7360 → 7371).

## Full suite

```
12364 passed, 642 skipped, 2 xfailed, 1 failed in 37:18
FAILED tests/test_lane_conformance.py::test_the_js_implementation_agrees_on_every_shared_case
```

That one fails identically on `main` with the branch stashed, and this change does not touch it.


eyalgolan (dismissed):
Approving. This is the right shape for the fix — a targeted single-column writer, exactly like `update_task_title` and `set_status`, rather than trying to make a read-modify-write safe.

## #343 reproduces on main and is fixed here

Built the race directly: a long-lived handle loads the row, a human raises the cap, then the handle's owner saves because it had "nothing to do".

```
                                    MAIN            PR379
stale handle loaded at              10,000,000      10,000,000
after the human's raise             14,000,000      14,000,000
after a 'nothing to do' tick        10,000,000      14,000,000
                                    REVERTED        SURVIVED
```

## The thing I went looking for, and did not find

Removing `config=:config` from two UPDATE statements is the kind of change that silently drops writes somewhere else. So I enumerated every site that mutates `task.config` rather than trusting the diff — eight of them — and checked how each persists:

```
cli/commands.py:1150,1155     -> create_task   (INSERT, writes config directly)
api/app.py:1003,1008          -> create_task   (INSERT)
eval/funnel_eval.py:270       -> create_task   (INSERT)
blockers/actions.py:184,214,216 (apply_action, pure) -> its three callers:
    cli/commands.py:1345  -> update_task_config at :1365
    cli/commands.py:3827  -> update_task_config at :3849
    api/app.py:4061       -> update_task_config at :4077
```

Every mutation is covered, and the three `apply_action` sites — the budget-raise path this issue is about — each write through the new method a few lines later. Putting that write beside the `apply_action` that produced it, rather than burying it in a generic save, is the part that makes this hold up.

```
full suite   12782 passed, 225 skipped, 2 xfailed, 0 failed
```

## Two notes, neither blocking

The docstrings on both changed writers now say what is excluded and why, with the issue number — that matters more than usual here, because the next person to add a column to `update_task_columns` needs to know config is deliberately absent rather than forgotten. Good.

`update_task_config` takes the whole blob rather than merging keys, so two concurrent writers to *different* config keys would still clobber each other. That is fine as shipped — config is human-only and the three call sites are all human-initiated — but it is worth knowing the guarantee is "one human writer at a time", not "per-key safe". Worth a line in the docstring if you touch it again; not worth a round trip now.

eyalgolan:
Thanks for this — the db-layer change is correct, and I can confirm it closes a defect that bit us for real today. Two things need fixing before it lands.

Context for the acceptance bar I used: earlier today a task's `lifetime_tokens` was raised 8,000,000 → 13,000,000, read back from sqlite and **confirmed persisted**, and was later found silently reverted to 8,000,000 while the task was mid-attempt — 238,016 weighted tokens left, carrying review feedback that would have been lost to BUDGET_EXHAUSTED. So the question I measured is not "does a config write land" but "can a concurrent writer holding a stale in-memory task still clobber it afterwards".

**Your change closes exactly that.** A probe driving the real code with a concurrent raise landing between the handler's `find_task` and its write, same instrument on both arms, with a positive control proving the raise is visible:

```
ARM main (0b8c2dc4)          ARM head (2aceb5e1)
FAILED probeA                PASSED probeA   <-- the live incident: CLOSED
FAILED probeC (free text)    PASSED probeC
FAILED probeB (park)         FAILED probeB
FAILED probeB2 (approve_plan) FAILED probeB2
4 failed, 2 passed           2 failed, 4 passed
```

Same result on the merge result against current main.

---

## B1 — the reply path still reverts on two verbs

`src/no_human/api/app.py:4076-4077` and `src/no_human/cli/commands.py:3848-3849`:

```python
if applied:
    await store.update_task_config(task.id, task.config)
```

`applied` is the return of `apply_action`, which returns a truthy summary for `{"park": true}` (`blockers/actions.py:133`) and `{"approve_plan": true}` (`:128`) — and **neither verb touches `task.config`**; it is only assigned at `actions.py:184`, inside the `set_task_config` branch below both early returns. So on those two options the handler writes the handle's pre-raise config wholesale over the row. Observed:

```
AssertionError: reverted to {'lifetime_tokens': 8000000, ...} — output: applied parked — work kept as-is
assert 8000000 == 13000000
```

This is **not a regression** — on main all four probes fail, so your PR strictly shrinks the exposure and free text is now safe. But `approve_plan` is the damaging one, because it *resumes* the task, so the reverted cap is then spent against.

The comments at `app.py:4072-4073` and `commands.py:3843-3844` state:

> "An option's `set_task_config` (a budget raise, usually) mutated `task.config` above, and config has its own writer since #343."

That is false for the two verbs that reach that line — the guard's own justification describes an invariant it does not enforce.

A verb-agnostic fix avoids re-enumerating the verb set (a closed set in a comment is how this class recurs). Snapshot before `apply_action`, and before the `choose` branch so the free-text path is defined:

```python
config_before = dict(task.config or {})
...
if (task.config or {}) != config_before:
    await store.update_task_config(task.id, task.config)
```

With that applied: all 6 probes pass, and `test_task_config_cli.py + test_config_clobber.py + test_send_back_refused.py` plus the three reply tests → `37 passed in 17.67s`. `nh task config` at `commands.py:1364` already uses `if settings:` and is correct — leave it.

Worth adding a held-out cell to `tests/test_config_clobber.py` for a park/approve_plan reply under a concurrent raise. The current 8 cells are DB-layer only, so nothing in the PR would have caught this.

## B2 — the three ratchets and four citations are stale; the gate goes RED on rebase

Main moved `api/app.py` 6332→6340 and `core/db.py` 5193→5285 in the four commits since your base `680d6889`. With your frozen values on the rebase result:

```
AssertionError: api/app.py: frozen 6338, now 6346 (+8); this budget only ratchets down
  core/db.py: frozen 5230, now 5322 (+92); this budget only ratchets down
FAILED tests/test_structural_budget.py::test_no_frozen_entry_has_grown
```

Correct post-rebase values, measured with the file's own scanner: **`cli/commands.py: 8945`, `api/app.py: 6346`, `core/db.py: 5322`**.

Citations drifted the same way (main shrank `commands.py` by 1) — `reanchor_citations.py --check` on the rebase result wants `approve:5442` (PR has 5443), `bench_run:8009 / 8160 / 8038` (PR has 8010 / 8161 / 8039), and `docs/WINDOWS.md` needs **7370**, not 7371. With those six corrected the resolved rebase runs `166 passed, 12 skipped` and `VERDICT=OK`.

---

## What I attacked and could not refute

The core is right, and I re-derived the audit rather than taking the PR body's word:

- **Exactly one `UPDATE` in the tree writes `tasks.config`** — `db.py:2289` inside `update_task_config`. `grep -rE "UPDATE\s+tasks" src/ | grep -v core/db.py` → empty, positive control 12 hits inside `db.py`. `create_task` is a plain INSERT with no `ON CONFLICT`/`OR REPLACE`. The docstring's "its only writer, as `set_status` is for status" is true.
- **Consumer audit reproduces.** 26 `.config` mutation sites swept and each read (positive control `\.title\s*=` → 1); all are create/INSERT paths except `actions.py:184/214/216`, whose three call sites are wired. 50 non-test `update_task` and 32 non-test `update_task_columns` call sites, none intending a config write. The aliasing class (`x = task.config` then mutate) is read-only lookups only.
- **No surviving mutations** on the db layer: config back into `update_task_columns` → 2 failed; into `update_task` → 2 failed; `update_task_config` → `return` → 6 failed; drop the `updated_at` bump → 1 failed.
- **The wiring is pinned by the suites you name, with no "rode on another gate" shape**: removing each of the three call sites kills exactly its own named test.
- **No `config_updated_at` marker, and none is needed** — `title` needs one because it is still *in* both column lists behind a CASE; `config` is excluded outright, which is strictly stronger.
- **`docs/WINDOWS.md` really was required**: reverting only that number fails `test_windows_md_code_line_citations_resolve`.
- **No tamper**: tests 10101 → 10109, assertions 28727 → 28744.
- Branch-head gates all green; the 11 failures + 18 errors in my full-suite arm were artefacts of a checkout with no `.git` (ablation in a git-initialised copy of the same head: `276 passed, 9 skipped`).

Two notes, neither blocking: your PR body's full-suite evidence cites `test_lane_conformance.py::test_the_js_implementation_agrees_on_every_shared_case` failing identically on main — it passes on all three arms here (base 81, head 81, main 81), so the conclusion is right but the evidence looks environment-specific. And `nh task config` now commits priority and config in two transactions where main committed once; a kill between them leaves priority applied and the cap not, which fails loudly rather than silently, but the PR doesn't mention the change.

eyalgolan (approved):
Approving. **I pushed a merge commit to your branch** to clear six conflicts — a merge, not a rebase, so nothing of yours was rewritten. All six were generated or derived; I re-derived from the merged source rather than merging either side's stale numbers. Details in the commit message, but three things you should know because they carry judgement rather than mechanics:

1. **Three structural budgets re-measured on the merge result**, each with its reasoning inline — `cli/commands.py` 9048→9059 (+11), `api/app.py` 6340→6346 (+6), `core/db.py` 5285→5322 (+37). Each matches your own diff exactly; the +37 is mostly the docstrings explaining why `config` is now excluded from the two generic writers, which is the part a future reader most needs.
2. **`docs/WINDOWS.md` re-anchored by hand** (7494→7484), located by content rather than by arithmetic. That doc is not in `_CITATION_DOC_PATHS`, so `reanchor_citations.py --apply` cannot fix it — that is #259, which I hit for real here.
3. Everything else fell out of `--apply` plus a manifest regenerate.

## Verified on the merge result

```
full suite                   12871 passed, 227 skipped, 2 xfailed, 0 failed
--strict alone               OK: 1620 file(s)
citations                    VERDICT=OK
structural budget            18 passed
```

## The review itself stands

#343 reproduces on main and is fixed here — I built the race rather than reading for it:

```
                               MAIN         PR379
stale handle loaded at         10,000,000   10,000,000
after the human's raise        14,000,000   14,000,000
after a 'nothing to do' tick   10,000,000   14,000,000
                               REVERTED     SURVIVED
```

And the thing I went looking for and did not find: removing `config=:config` from two UPDATE statements is exactly the change that silently drops writes elsewhere, so I enumerated all eight sites that mutate `task.config` instead of trusting the diff. Five reach `create_task` (an INSERT, unaffected); the three `apply_action` callers each pair with `update_task_config` within twenty lines. Complete coverage.

One note I will repeat from the first review because it is worth writing down somewhere permanent: `update_task_config` takes the whole blob, so the guarantee is "one human writer at a time", not "per-key safe". Correct as shipped — config is human-only and all three call sites are human-initiated — but if a future caller writes a different key concurrently, that is where to look.

Clean as of `49eaa906`.

eyalgolan:
Landed as `8dccde9f`, with you as the commit author.

PR #381, opened 2026-09-14
docs(guard): record the renamed-installer gap in the residual-risk register

### LINKED ISSUE #341: venv guard identifies installers by resolved basename, so a renamed copy bypasses it entirely
Found by an independent review of the chmod-bypass fix, and verified separately with a positive control. Pre-existing and by construction — not a regression, and not something to fold into that fix. Filing it because it is absent from the module's residual-risk register, which is otherwise exhaustive, and an undocumented gap in a security guard reads as a closed one.

## The gap

`venv_install_guard` decides whether a command is an install by the **resolved basename** of the binary (`_basename` → `_is_installer_name`). An installer invoked under any other name is invisible to every check in the module.

```
  control    DENY   <- <shared venv>/bin/pip install evilpkg
  renamed    ALLOW  <- cp <shared venv>/bin/pip /tmp/notpip && /tmp/notpip install evilpkg
```

The copied file keeps its shebang, so it runs the same interpreter and installs into the same shared venv. The control line is what makes this meaningful: the identical operation under its own name is correctly refused.

The same shape appears through a symlink — a `<venv>/bin/pip` symlinked to a target whose basename is not an installer name resolves to `None` and is allowed, because the discard happens on the basename of the **realpath**, not of the name the command actually typed.

## Why this is not simply a bug to fix

The module is deliberately built to avoid lexical rules — matching on command text rather than on filesystem structure is exactly what it was written to get away from, and the project has a standing finding that a lexical guard cannot enforce a capability. Closing this by pattern-matching command strings would reintroduce the class the module exists to escape, and would be trivially evadable in its own right.

So the realistic options are:

1. **Document it.** Add one line to the residual-risk register beside the existing entries (shell expansion, stdin, `bash script.sh`, runtime PATH mutation, `sudo -u`, inherited `VIRTUAL_ENV`). Cheapest, honest, and consistent with how the module already treats what it cannot see.
2. **Structural detection.** Decide "is this an installer" from something other than the name — the shebang's interpreter plus the argv shape, or the resolved target's inode matching a known installer in a venv's `bin/`. More faithful to the module's own philosophy, considerably more work, and needs its own false-positive analysis.

My recommendation is (1) now and (2) only if the threat model ever includes a coder deliberately evading the guard. Today's failures are accidental — a shared venv on PATH, an inherited `VIRTUAL_ENV` — not adversarial renames.

## Scope note

Deliberately NOT fixed in the chmod-bypass diff. That change is unblocking a red trunk and has already been through three review rounds; widening it to a detection-strategy question would be scope creep on a security guard under time pressure. Related: the resolver-contract rewrite proposed in the companion issue about electing one representative from an indeterminate set.

L4XB:
Closes #341 — option (1), which is what the issue recommends.

## What is added

One entry in `venv_install_guard`'s residual-risk register, beside shell expansion, stdin, `bash script.sh`, runtime `PATH` mutation, `sudo -u` and the inherited `VIRTUAL_ENV`.

## Measured, not restated

The issue's table reproduced here before writing it down, because the register is the module's evidence, not its intentions:

```
control    DENY   <shared venv>/bin/pip install evilpkg
renamed    ALLOW  cp <shared venv>/bin/pip /tmp/notpip && /tmp/notpip install evilpkg
symlink    ALLOW  <shared venv>/bin/pip-link -> /tmp/notpip
```

The control is load-bearing and it earned its place: my first fixture put the shared venv **inside** the worktree, so all three read `ALLOW` — which without a control would have looked like "no gap" rather than "the fixture is wrong". Moving the venv outside `cwd` produced the split above.

The symlink half is the same defect seen from the other side: the discard happens on the basename of the **realpath**, not on the name the command actually typed, so a `bin/pip` pointing at a non-installer name resolves to `None` and is allowed.

## Why documented rather than fixed

The issue says it and the module says it: matching on command TEXT is the class this module exists to escape (*a lexical guard cannot enforce a capability*), and a name-pattern rule would be trivially evadable in its own right. The structural alternative — deciding "is this an installer" from the shebang's interpreter plus the argv shape, or from the resolved target's inode matching a known installer in a venv's `bin/` — is real work with its own false-positive analysis, and is worth it only if the threat model ever includes a coder deliberately evading the guard. Today's failures are accidental: a shared venv on `PATH`, an inherited `VIRTUAL_ENV`.

An undocumented gap in a security guard reads as a closed one. This makes it read as what it is.

## Checks

Comment only, no behaviour change.

```
pytest -k "venv_install_guard or venv_guard or structural_budget or readme_claims or reanchor"
  245 passed, 13 skipped
check_release_manifest.py --strict                OK (re-inventoried; the file's hash moved)
```

`agent/venv_install_guard.py` carries no `FROZEN_FILE_LINES` entry, so the structural ratchet has nothing to re-anchor here.


eyalgolan (dismissed):
Approving. An undocumented gap in a security guard reads as a closed one, and this register was exhaustive everywhere except here — so filling it in is worth a PR on its own.

Since this PR is entirely a factual claim about what the guard does **not** stop, I reproduced all three rows rather than taking the table on faith:

```
control    DENY   <shared venv>/bin/pip install evilpkg
renamed    ALLOW  /tmp/notpip install evilpkg          (cp of that pip)
symlink    ALLOW  <shared venv>/bin/pip-link -> /tmp/notpip
```

Exactly as written, control included — and the control is what makes the other two mean anything, so I am glad it is in the text and not just in your head.

```
--strict alone   OK: 1610 file(s)
citations        VERDICT=OK
guard + budget + claims   216 passed, 12 skipped
merge into main  0 conflicts
```

## On the judgement call

The part I want to endorse explicitly is the decision **not** to fix it with a smarter name rule:

> matching on command TEXT is the class this module exists to escape … and a text rule would be evadable in its own right

That is right, and it is the reasoning that keeps this file honest. A basename allowlist that also matched `notpip` would look like a fix, close nothing, and cost the next reader an hour working out why the module has two philosophies. Naming the structural options (shebang interpreter plus argv shape, or inode identity against a known installer) and saying they are only worth it if the threat model turns adversarial is the correct disposition of a gap you are not closing today.

Worth noting for anyone reading later: this is accidental-misuse territory, not defence against a coder deliberately evading the guard, and the text says so in those words. If that threat model ever changes, this register entry is where the work starts.

Clean as of `a8a04496`, expires on the next landing.

eyalgolan (approved):
Re-approving at `6c50eb5c`. **I pushed a merge commit to your branch** to clear the manifest conflict — a merge, not a rebase, so nothing of yours was rewritten.

Verified on the merge result against `49eaa906`:

```
conflicts                   0
--strict alone              OK: 1619 file(s)
manifest rows differing     exactly 1: src/no_human/agent/venv_install_guard.py
guard + budget + claims     224 passed, 12 skipped
```

One row, matching the one file you change. Worth saying why I checked that rather than trusting a green `--strict`: a hash manifest can merge *cleanly and wrongly*, so "no conflict" proves nothing on its own — the row set has to match the diff.

The review itself stands: I reproduced all three rows of your residual-risk table (control DENY, renamed ALLOW, symlink ALLOW), and the control is what makes the other two mean anything. Declining to paper over it with a smarter name rule is the right call and the reasoning is in the text.

Clean as of `49eaa906`.

eyalgolan:
Landed as `0888c30b`, with you as the commit author.

PR #382, opened 2026-09-14
fix(guard): stop a denial claiming a resolution the shell would not make

### LINKED ISSUE #338: venv guard resolves to a binary the shell will not execute, instead of reporting an indeterminate target
Raised by an independent review of the fix for the chmod bypass (the change that took trunk from red to green). The bypass fix is correct and lands separately; this is the design objection it leaves open, recorded rather than folded in, because re-architecting the resolver while trunk was red was the wrong trade.

## The objection

`_resolve_installer`'s bare-token PATH walk justifies its own ordering with this, ~90 lines above the new code:

> a POSIX shell skips an EACCES entry and keeps walking, so `command -v`/`type -p` resolve past it to the same later match

That is true, and it was measured:

```
--- readable: which entry wins? ---          # positive control
.../venvA/bin/pip
RAN: venvA pip
--- venvA chmod 0600 ---
.../sysbin/pip
RAN: sysbin pip          (sh exit=0; bash -lc identical)
```

The fix makes `_resolve_installer` return `<venvA>/bin/pip` in exactly the case where the module has already established that binary never runs. In a module whose thesis is "resolve what actually executes and where it writes", that is a proxy for the answer rather than the answer.

## Why it is still the right verdict, and why it is still worth fixing

The DENY is correct — fail-closed on an indeterminate read is this module's standing rule for a tri-state probe. But the mechanism achieves a right verdict through a claim that is not true, and that shows up in two user-visible places:

1. The denial text tells the user the install "resolves to `<foreign venv>`". It does not resolve there; the shell would run the system binary. The user is given a false fact to act on.
2. A non-venv directory whose own parent is unreadable reads as a venv (`_venv_root_of` treats an undetermined `pyvenv.cfg` probe as a venv root), so it produces a DENY naming a path no install would ever reach. Realistic triggers: a non-root process with `/root/.local/bin` on PATH, a PATH entry under another user's home, and — unverified, not simulated — a stale NFS mount raising `ESTALE`/`EIO` through the same `except (OSError, ValueError)` in `_probe_is_file`.

## The shape a fix should take

Do not replace the resolved installer with a file that cannot execute. Either:

- keep the determinate winner as the resolved installer and add the undetermined entry's venv as an ADDITIONAL prefix candidate, so the verdict is fail-closed without the resolver asserting something false; or
- return an explicit indeterminate result and let `denial_reason` say what is actually true — that the target could not be established because a PATH entry could not be read — which is the wording the `_UNRESOLVABLE_CHARS` branch already uses for the analogous case.

The second is closer to how the module already talks about targets it cannot prove.

## Constraint for whoever takes this

Two existing tests pin OPPOSITE outcomes for a superficially identical shape and both are right; any change must keep both green:

- `test_an_unreadable_path_entry_ahead_of_the_sessions_own_venv_is_skipped_not_denied` — unreadable decoy first, session's OWN venv behind it, must ALLOW.
- `test_an_unreadable_venv_is_not_escaped_by_a_system_installer_later_on_path` — unreadable foreign venv first, system pip behind it, must DENY.
- `test_a_non_venv_undetermined_entry_does_not_shadow_a_later_undetermined_venv` — two undetermined entries, non-venv first, must DENY.

Also note the accompanying false-premise correction already landed with the bypass fix: `pip` does NOT read `VIRTUAL_ENV` (`grep -rn VIRTUAL_ENV pip/_internal` → 0 matches against a 9-file positive control; it decides on `sys.prefix != sys.base_prefix`), and `uv sync` prints `does not match the project environment path .venv and will be ignored`. Only `uv pip` honours the variable. Do not reintroduce that reasoning.

L4XB:
Closes #338.

## The objection, and what actually moves

The verdict was already right — fail-closed on a tri-state read is this module's standing rule. What was wrong is the sentence. A POSIX shell skips an EACCES entry and keeps walking, so the binary the denial named is one the shell would never execute, and the message said the install **resolves to** it.

Measured on the shape from the issue:

```
readable    install blocked: resolves to <foreign venv>, outside this session's worktree …
unreadable  cannot verify this install's target: <foreign venv>/bin/pip could not be read
            (permission denied?), so the effective install location cannot be resolved
            before the command runs — and what could be read points outside this
            session's worktree (<wt>). Spell the literal path instead …
```

Both DENY. Only the claim changed, to the wording the `_UNRESOLVABLE_CHARS` branch already uses for the analogous case. The path is still named, as the thing that could not be read — which is what the user can act on.

This also covers the register's known imprecision that the issue lists as its second user-visible symptom: a non-venv directory whose own parent is unreadable reads as a venv (`_venv_root_of` treats an undetermined `pyvenv.cfg` probe as a venv root), which used to produce a DENY naming a path no install would reach. It now produces the honest wording for the same reason.

## Why not option (2) as written

> return an explicit indeterminate result and let `denial_reason` say what is actually true

That is the right *statement*, and the wording above is it. But changing `_resolve_installer`'s return type to carry it moves a contract that a dozen tests pin directly (`resolved == os.path.join(primary_venv, "bin", "pip")`, in the two cases the issue itself names as must-stay-green) and that `_effective_prefixes` consumes — a large diff across a security guard to fix a sentence.

So the indeterminacy is derived at the message site instead: the same probe the resolver asked, about the same file. Nothing else moves. **The three tests the issue names are untouched and green**, including their `resolved ==` assertions and their `primary_venv in inside` message assertions — the path is still in the message.

If you would rather have the typed result anyway, say so and I will do it as its own change with the test migration it implies; I did not want to spend that budget inside a fix whose user-visible content is one sentence.

## Tests

Two cells in `tests/test_venv_install_guard.py`, and the control inside the first is load-bearing — the **readable** reading must KEEP `resolves to`, because there the claim is established:

- `test_a_denial_resting_on_an_unreadable_entry_does_not_claim_it_resolves_there`
- `test_the_honest_wording_does_not_leak_into_a_fully_readable_denial` — an unreadable determinate **non-venv** entry on PATH must not re-word a denial that does not rest on it

| mutation | red |
| --- | --- |
| remove the branch (i.e. `main` today) | the #338 cell |
| fire the honest wording unconditionally | **both** cells |

```
pytest tests/test_venv_install_guard.py                        65 passed
pytest <the six files that name the guard>                     494 passed, 2 xfailed
pytest test_structural_budget/test_readme_claims/test_reanchor  158 passed, 12 skipped
check_release_manifest.py --strict                              OK (re-inventoried)
```

`agent/venv_install_guard.py` carries no `FROZEN_FILE_LINES` entry, so there is no structural ratchet to re-anchor.

## Not reintroduced

The DO-NOT-RESTATE paragraph stands: nothing here reasons about `VIRTUAL_ENV`. `pip` decides venv membership by `sys.prefix != sys.base_prefix` and never reads the variable; `uv sync` ignores it with a warning; only `uv pip` honours it.


eyalgolan:
Status note: this now conflicts on source, not just the manifest.

```
conflicts against main (c176f291):
  RELEASE_MANIFEST.txt
  tests/test_venv_install_guard.py
```

`c176f291` ("The case-fold probe answers from a path that exists at runtime") touched that test file, so your two additions land in the same region. Nothing wrong with your change — it just needs a rebase.

The review itself is done and positive; I will post it as soon as this merges clean, so you are not waiting on me. One heads-up while you are in there: `46927373` bumped the version to 0.2.3, so if you run the suite in a worktree against a shared venv you may see `test_the_declared_version_and_the_installed_metadata_agree` fail — that is the venv's installed metadata (0.2.3) disagreeing with a pre-bump tree, not your change. It passes on the merge result.

eyalgolan (approved):
Approving. **I pushed a merge commit to your branch** to clear the conflicts — a merge, not a rebase, so nothing of yours was rewritten. One of the two was a real source conflict, so here is exactly what I did with it.

## The source conflict, and how I resolved it

`tests/test_venv_install_guard.py`: one hunk, your two tests against main's three from `c176f291`, both appended at the end of the file. The trap is that the shared trailing line after the `>>>>>>>` marker is a single `)` that closes the final `assert` on **each** side — so the obvious "keep both" truncates a statement. My first two attempts did exactly that and produced `SyntaxError: '(' was never closed`.

Resolved by keeping both blocks with that closing line duplicated, then checking the things that could still be wrong:

```
conflict markers remaining   0
file parses (ast.parse)      OK
all five contested tests     present exactly once
tests/test_venv_install_guard.py   73 passed
```

No name collision between your two and main's three, which I checked before deciding both could coexist rather than assuming it.

## The review

```
full suite                   12865 passed, 227 skipped, 2 xfailed, 0 failed
--strict alone               OK: 1619 file(s)
citations                    VERDICT=OK
```

Your tests are load-bearing — against main's guard, `test_a_denial_resting_on_an_unreadable_entry_does_not_claim_it_resolves_there` fails, and both pass here. And the scoping is pinned: mutating your condition to `if True` reddens 31 tests including your own `test_the_honest_wording_does_not_leak_into_a_fully_readable_denial`, which is the control that stops the honest wording leaking into a denial that does not rest on an unreadable entry.

The verdict never moves — the new branch returns a string in every case it fires, so a DENY cannot become an ALLOW. That was the first thing I checked, because a message-only change in a guard is exactly where a stray `None` would hide.

## What I liked

Deriving the unreadable set at the point of use rather than threading it out of `_resolve_installer` is the right call — that function's contract (a path, or `None`) is what a dozen tests pin, and widening it to carry provenance would have been a refactor wearing a bugfix's clothes.

Clean as of `49eaa906`.

eyalgolan:
Landed as `736cab6e`, with you as the commit author.

PR #404, opened 2026-09-15
Wiring evidence reads JS/TS and class bodies, not just Python top level

PrinceXDev:
## What this changes

Wiring evidence now reads the JS/TS family and class bodies, not only Python
module top level, so the "implemented but never called by the production path"
check can see a method and a JS module. Part of #114 (phase 4).

The issue makes phase 4 contingent on infrastructure phase 3 justified, and
PR #229's PROCEED authorises the A/B and nothing more. This deliberately
builds on none of it: the reference half of the check was always
language-agnostic (`git grep -w` reads bytes), so only the declaration half
needed widening, and it is widened with stdlib `ast` and a line scanner — no
symbol server, no language server, no new dependency, nothing that waits on
the A/B's verdict.

New `src/no_human/review/symbols.py` is that declaration half. Python at
module level and inside class bodies, reported by qualified name
(`Store.recompute_totals`); `export`ed module-level declarations in
`.js/.jsx/.mjs/.cjs/.ts/.tsx`, read from source whose comments and string
literals have been blanked first. Three things are deliberately NOT collected,
and each has a test: a `def` inside a function and an unexported JS binding
(file-private by construction — listing them would bury the real findings),
and a dunder method (the language calls it, so no reference search could find
the call that does exist). `collect_wiring_evidence` also moves from a
per-subprocess timeout to one deadline over the whole pass, because with a
`git grep` per name a per-call timeout bounded nothing in aggregate, and it
memoizes the search by name.

Why it was worth doing: over the 300 most recent non-merge commits on `main`,
25% of the commits that touch code touch something other than Python
(`.mjs` 63, `.js` 44, `.jsx` 38) and got no wiring evidence at all, and 20%
of the Python symbols those commits add are declared inside a class.

## Evidence

Old vs new collector replayed over the 60 most recent non-merge commits on
`main`. The widening is a strict superset — no finding lost:

```
commits examined       : 60
findings, old collector: 79 across 20 commits
findings, new collector: 88 across 22 commits
  already seen by the old collector: 79
  net-new: JS/TS: 3
  net-new: class body: 6

   e85feb29 src/no_human/core/orchestrator.py Orchestrator._citation_drift_preflight
   a8a04496 web/src/replayScrub.js API_BODY_CLASSIFICATION
   1e42e2f6 web/src/onboardingEmail.js EMAIL_MAX_LOCAL_LEN
   1e42e2f6 web/src/replayScrub.js REPLAY_EXCLUDED_PATHS
   5774bab6 src/no_human/core/orchestrator.py Orchestrator._handle_pre_review_red
   5774bab6 src/no_human/core/orchestrator.py Orchestrator._round_failure_attribution
   83b0b364 src/no_human/core/scheduler.py Scheduler._cas_heartbeat_with_retry
   93de0329 src/no_human/vcs/git.py GitRepo._null_paths
   93de0329 src/no_human/vcs/git.py GitRepo._run_null
```

The new tests against the UNFIXED collector, in a worktree at `main`
(`tests/test_symbols.py` cannot import there at all, so only the wiring file
is shown):

```
$ python -m pytest tests/test_wiring_evidence.py -q -p no:randomly
FAILED tests/test_wiring_evidence.py::test_a_method_no_caller_reaches_is_listed_by_qualified_name
FAILED tests/test_wiring_evidence.py::test_an_exported_component_nothing_renders_is_listed
FAILED tests/test_wiring_evidence.py::test_a_js_test_file_is_neither_a_definer_nor_production_wiring
FAILED tests/test_wiring_evidence.py::test_an_unexported_js_binding_is_not_listed
FAILED tests/test_wiring_evidence.py::test_format_block_is_labeled_hedged_and_empty_on_no_findings
5 failed, 11 passed in 17.84s
```

Single-change mutation over `review/symbols.py`, each mutation deleting one
rule the module claims. The first pass had two survivors: both were fixtures
that did not exercise the rule they named, and chasing the second surfaced a
guard that `re.match` already made unreachable, now deleted.

```
KILLED  dunder rule removed  (test_python_skips_dunders_because_the_language_calls_them)
KILLED  function bodies descended  (test_python_skips_definitions_inside_a_function_body)
KILLED  export no longer required  (test_javascript_skips_an_unexported_binding)
KILLED  match relaxed to search  (test_javascript_skips_an_indented_declaration)
KILLED  literals no longer blanked (block comment)  (test_javascript_does_not_mine_a_commented_out_declaration)
KILLED  quote state not closed at the newline  (test_an_apostrophe_in_jsx_text_costs_one_line_not_the_rest_of_the_file)
KILLED  literals no longer blanked (template)  (test_javascript_does_not_mine_a_template_literal)
KILLED  d.ts no longer excluded  (test_a_typescript_stub_is_not_a_declaration_source)

8 of 8 killed, 0 survivors
```

Wall-clock of the collector on 12 real commits of this repo. It stays far
inside its deadline, so it keeps the "a few git reads" cost
`reviewer._collect_gate_evidence` assumes and still does not need a thread:

```
84146eee    0.33s    1 finding(s)
3b937c45    0.27s    0 finding(s)
e85feb29    4.09s    5 finding(s)
8dccde9f    1.41s    0 finding(s)
676a9cfc    0.20s    0 finding(s)
c388aaf5    0.17s    0 finding(s)
736cab6e    0.12s    0 finding(s)
0888c30b    0.14s    0 finding(s)
31845063    2.00s   10 finding(s)
e157cee0    0.05s    0 finding(s)
49eaa906    1.72s    5 finding(s)
c176f291    0.53s    3 finding(s)

worst case over 12 commits: 4.09s (deadline is 30s)
```

Full suite, this tree and a pristine `main` worktree, same command and same
machine:

```
$ python -m pytest -q -n 4 --tb=no -rf
this tree : 633 failed, 11837 passed, 679 skipped, 2 xfailed, 65 errors in 4453.24s
main      : 632 failed, 11807 passed, 685 skipped, 2 xfailed, 65 errors in 4588.15s

$ comm -23 failures_this_tree failures_main     # fails only here
FAILED tests/test_scheduler.py::test_events_persisted_to_store_on_task_finish
$ comm -13 failures_this_tree failures_main     # fails only on main
(none)
```

That single differing test is order-dependent, not a regression: it fails the
same way on pristine `main` when run alone.

```
$ python -m pytest tests/test_scheduler.py::test_events_persisted_to_store_on_task_finish -q   # in the main worktree
tests/test_scheduler.py:917: AssertionError: assert 'test_seed' == 'tool_use'
1 failed in 2.63s
```

The 632 shared failures and 65 errors are this checkout, not this branch. The
first one traced to `scripts/export_guard.py`, which the public export does
not ship — `scripts/precommit_manifest_gate.py:78` records that it and
`EXPORT_CLASSIFICATION.txt` are classified `drop` — so every fixture that
shells out to the guard errors by construction here. Please read the
`comm` diff above rather than the absolute numbers; a Linux CI checkout with
the dropped files present is the run that produces a meaningful total.

## Checklist

      checks for the file.)

Notes on the checklist. No dependency changed, so no lockfile moved. Test
count rises by 25: `tests/test_symbols.py` is new (17) and
`tests/test_wiring_evidence.py` goes from 8 to 16. Nothing was deleted; one
test was renamed and strengthened
(`test_non_python_diffs_contribute_nothing_and_do_not_crash` ->
`test_a_language_no_reader_claims_contributes_nothing_and_does_not_crash`,
which now also passes a `.go` file through). The passed counts reconcile
exactly: 11807 + 25 new + 6 that skipped on main and ran here - 1 flaky =
11837; I did not root-cause the 6 skip-to-pass, and note only that it moves
coverage up, not down. `contributors/princexdev.md` already exists from an
earlier PR.


eyalgolan (approved):
Approving. The evidence in the description is the kind that makes a review cheap, and it held up when I re-ran it myself rather than reading it.

## Verified independently

**The widening is a strict superset.** I replayed the old and new collectors over the 25 most recent non-merge commits on `main`:

```
findings, old        : 45
findings, new        : 47
LOST by the widening : 0
net-new              : 2
```

Smaller sample than your 60, same direction, and zero losses — which is the number that matters, because a widening that quietly *swaps* findings would look like an improvement in a raw count.

**The JS blanker holds against the cases that usually break hand-rolled scanners.** I threw six at it:

```
export in a line comment      -> ['REAL']   OK
export in a string            -> ['REAL']   OK
export in a template literal  -> ['REAL']   OK
export in a block comment     -> ['REAL']   OK
unexported binding skipped    -> ['REAL']   OK
regex literal with slashes    -> ['REAL']   OK
```

The regex-literal case is the one I expected to leak, since `/export const FAKE/` looks like the start of a comment to a naive scanner. It does not.

```
tests/test_symbols.py + test_wiring_evidence.py   33 passed
full suite                                        12987 passed, 0 failed
--strict alone                                    OK: 1626 file(s)
citations                                         VERDICT=OK
structural budget                                 18 passed
```

## The docstring is doing real work

Naming the ceilings where the prompt block repeats them is right, and the `BY BARE NAME` paragraph is the one I would have asked for if it were missing:

> a method is searched for as `close`, not `Store.close`, so an unrelated `other.close()` elsewhere in the tree reads as a reference … a listed method is stronger evidence than an unlisted one is.

That is the honest statement of an asymmetric instrument, and it tells a reviewer exactly how to weigh a null result. Same for the three deliberate exclusions each carrying a test — a nested `def`, an unexported binding, a dunder. The dunder reasoning in particular ("the language calls it, so no reference search could find the call that does exist") is the sort of thing that would otherwise get "fixed" by a future contributor.

## One asymmetry I measured, not a blocker

The whole-pass deadline behaves differently depending on which phase it expires in:

```
normal timeout        -> 6 findings
timeout=0.35 (search) -> 0, partial-by-break
timeout=-1  (collect) -> 0, everything discarded
```

The search loop guards with `if remaining() <= 0: break` and keeps what it decided, which is what the docstring promises. The earlier collect loop calls `_show(..., remaining())` unguarded, so once the deadline passes a negative timeout raises and the handler returns `[]`. Both fail toward silence, both log (`wiring evidence timed out after -1s`), and the docstring's claim is specifically about running out *mid-search*, so nothing is misstated. But a reader could reasonably expect the same partial behaviour in both halves. A `max(0.0, ...)` clamp in `remaining()` plus the same guard before the `_show` pair would make the two phases behave alike. Worth doing when you are next in here; not worth another round now.

## North star and security

Yes on both. Your measurement — 25% of code-touching commits are non-Python and got no wiring evidence at all, 20% of added Python symbols are inside a class — is the argument for this being worth the code, and it is a measurement rather than an intuition.

No new execution surface: `git show` and `git grep` were already the mechanism, the new module is pure string and `ast` work, and nothing reaches the network. The move from a per-subprocess timeout to one deadline is a genuine improvement rather than a wash, since a per-call bound over N searches bounded nothing in aggregate.

Clean as of `84146eee`, expires on the next landing. I will land this shortly.

eyalgolan:
Landed as `d3bb252d`, with you as the commit author.

Squashed onto main and verified on the squashed tree — the tree that actually lands — with the full suite green (`12987 passed, 0 failed`) and your A/B re-run there.

Two things I want to call out, because they made this an easy review:

The evidence in your description was reproducible. I replayed old-vs-new myself over 25 commits and got the same shape you did — **0 findings lost**, net-new on top. A widening that quietly swapped findings would show a healthy count and be a regression; proving the superset is the check that rules that out, and you did it before anyone asked.

And the ceilings are written where they get read. The `BY BARE NAME` paragraph — a method searched as `close`, so an unrelated `other.close()` reads as a reference, therefore a listed method is stronger evidence than an unlisted one — is exactly the kind of thing that usually lives in someone's head and gets lost. Same for the dunder exclusion carrying its reasoning, which will stop a future contributor "fixing" it.

The one item from my review is non-blocking and noted in the commit: the whole-pass deadline keeps partial results when it expires mid-search but discards them in the earlier collect loop, where `_show` gets a negative timeout. Both fail toward silence and both log, so nothing is misreported — a `max(0.0, ...)` clamp in `remaining()` plus the same guard before the `_show` pair would make the phases behave alike. Worth picking up next time you are in that file.

Thanks — this is phase 4 of #114 done without waiting on the infrastructure it was nominally gated behind, which was the right read.

PR #407, opened 2026-09-15
Strip the executable suffix in the raw-text guard matchers (#305 residual)

### LINKED ISSUE #305: A trailing .exe bypasses the command guard's name matching, including the never-merge and never-push-to-main gates

### LINKED ISSUE #406: A trailing .exe still bypasses the raw-text guard matchers (rm.exe -rf, gh.exe pr merge, nh.exe serve): the residual #305 records but #320/#339 did not reach
Split out from #305, which closed after #320 (argv half) and #339 (casing half) landed. This is the third and last axis of the same class, and the one #305's own recording test still asserts open on `main` today.

The argv resolver `exec_names.command_name` strips `.exe`/`.cmd`/`.bat`/`.ps1`/`.com` for `argv[0]`. But the guard also decides through RAW-TEXT matchers that name a binary literally before a space (`_RM_RF`, `_FORGE_MERGE`, `_GIT_DESTRUCTIVE`, `_GIT_WRITE`, `_FORGE_WRITE`, `_FORGE_MENTION`, `_GIT_MENTION`, `_LEXICAL_MERGE_STACK`, `_LEXICAL_LIVE_SERVER`, plus the whole-string `\bgit\s+push\b` fallback in `evaluate`). Those never see a basename, so the suffix walks straight past them.

## Measured on `main` at `84146ee`, Windows reading

```
rm.exe -rf /                         ALLOW   _RM_RF
gh.exe pr merge 7                     ALLOW   _FORGE_MERGE
echo "gh.exe pr merge 7" | sh         ALLOW   _FORGE_MERGE (unanchored mention)
git.exe push --force origin main      ALLOW   _GIT_DESTRUCTIVE
timeout 30 git.exe push origin main   ALLOW   whole-string never-push fallback
nh.exe merge-stack run                ALLOW   _LEXICAL_MERGE_STACK
nh.exe serve                          ALLOW   _LEXICAL_LIVE_SERVER
```

The bare spellings all DENY. Casing was closed by #339; this is the orthogonal suffix axis, which #339 did not touch. `tests/test_exec_names.py::test_the_raw_text_matchers_are_still_open_and_that_is_recorded` still asserts three of these ALLOW on `main`, so the residual is green and live.

## Why a name resolver cannot reach it

`command_name` resolves `argv[0]`; these matchers read the command as text and never call it. Closing them means widening the patterns themselves, the same way #339 widened them for casing. Your own #307 note applies: enumerate the matcher set by script, not by eye (that PR widened 12 by hand and still missed `_LEXICAL_MERGE_STACK` and `_LEXICAL_LIVE_SERVER`), and prefer a shared fragment over an inline suffix at each site.

## Shape I have ready (branch `fix-305-raw-text-exe`, PR to follow)

- One shared constant `exec_names.EXE_SUFFIX_RE`, derived from the same `_EXECUTABLE_SUFFIXES` tuple the argv side strips, so the two cannot drift. Optional, non-capturing, carries no case logic (`case_flags()` already handles that at each site).
- Spliced after each binary name at 15 sites: 14 named `re.compile` patterns enumerated by script, plus one inline `re.search` the compile-scan missed. The inline one is exactly why enumerating by eye is unsafe.
- `agent/guard.py` stays at its frozen 3036: every splice is in-place, the fragment lives in the unfrozen `exec_names.py`.
- Ungated like the argv strip (#107): a POSIX file literally named `rm.exe` is unusual, and reading it as `rm` errs toward denial.

Measured side effect, pinned rather than claimed: the raw-text matchers read the ORIGINAL string, so `\bgit` finds `git` inside an unquoted `C:\tools\git.exe push` that `shlex` would have mangled. That closes the raw-text-reachable half of #105's unquoted-backslash residual too (push, reset --hard, forge merge, rm -rf, merge-stack). The argv-only gates (`git stash` working-tree clobber, `nh approve`) still leak an unquoted backslash path, so those stay asserted open and the PR is `Refs #305`, not a claim that the backslash class is done.

Filing this as the tight follow-up I offered on #305 rather than reopening it, since closing #305 was your call.

Sreekant13:
Closes #406. Refs #305, #105.

The argv resolver `exec_names.command_name` strips `.exe`/`.cmd`/`.bat`/`.ps1`/`.com` for `argv[0]`, but the guard also decides through RAW-TEXT matchers that name a binary literally before a space, and those never see a basename. So after #320 (argv half) and #339 (casing half) closed, these still walked past on Windows:

```
rm.exe -rf /                         _RM_RF
gh.exe pr merge 7                     _FORGE_MERGE
echo "gh.exe pr merge 7" | sh         _FORGE_MERGE (unanchored mention)
git.exe push --force origin main      _GIT_DESTRUCTIVE
timeout 30 git.exe push origin main   whole-string never-push fallback
nh.exe merge-stack run                _LEXICAL_MERGE_STACK
nh.exe serve                          _LEXICAL_LIVE_SERVER
```

## What this does

One shared fragment `exec_names.EXE_SUFFIX_RE`, derived from the same `_EXECUTABLE_SUFFIXES` tuple the argv side strips so the two cannot drift, spliced after each binary name at 15 sites: 14 named `re.compile` patterns enumerated by script, plus one inline `re.search` fallback the compile-scan missed. The inline one is why the count is 15 and not 14, and why the set was enumerated by script rather than by eye (your #307 note: that PR widened 12 by hand and still missed `_LEXICAL_MERGE_STACK` and `_LEXICAL_LIVE_SERVER`).

The fragment is optional and non-capturing, so it splices cleanly into any pattern, and it carries no case logic: `case_flags()` already handles casing at each site from #339, so this is the orthogonal suffix axis only. Ungated like the argv strip (#107): a POSIX file literally named `rm.exe` is unusual, and reading it as `rm` errs toward denial.

`agent/guard.py` stays at its frozen 3036: every splice is in place, the fragment lives in the unfrozen `exec_names.py`.

## Side effect, measured and pinned rather than claimed

The raw-text matchers read the ORIGINAL string, so `\bgit` finds `git` inside an unquoted `C:\tools\git.exe push` that `shlex` would have mangled. That closes the raw-text-reachable half of #105's unquoted-backslash residual too (push, reset --hard, forge merge, rm -rf, merge-stack). The argv-only gates (`git stash` working-tree clobber, `nh approve`) still leak an unquoted backslash path, so those stay asserted open and this is `Refs #305`/`#105`, not a claim that the backslash class is done. That remainder wants a name resolver on the win-readings path, which is out of scope here.

## Tests

- `tests/test_raw_text_exe_coverage.py` (new): behavioral. Every executable spelling denies wherever the bare one does, for each raw-text matcher; a scheduling wrapper plus `.exe` still denies the push; a false-denial control set (a `.exe` operand, a file named like a binary, viewing a PR) stays allowed; and a guard that `EXE_SUFFIX_RE` is derived from the one suffix source. Revert the `guard.py` splices and six rows go red.
- `tests/test_exec_names.py`: the two prior "still open" assertions were updated. The raw-text rows are promoted to DENY, and the unquoted-backslash boundary is retargeted to the argv-only cases (`git stash`, `nh approve`) that genuinely remain, so the test still records exactly what is open.

Guard suites show zero new failures against the same checkout's baseline. Manifest repinned (`--strict` clean), citations reanchored (VERDICT=OK), no CRLF introduced.

Note on the auto-close keyword: the closing keyword is here in the description for #406 only. The single commit says `Refs #305`, never `Closes`, so a squash merge cannot re-close #305 the way it did once before.


eyalgolan (approved):
Approving. Good find — #320 closed the argv half and #339 the casing half, and it would have been easy to assume the class was finished.

## Measured

```
                                MAIN    PR407
rm -rf /                        DENY    DENY
rm.exe -rf /                    ALLOW   DENY    <- the hole
gh pr merge 7                   DENY    DENY
gh.exe pr merge 7               DENY    DENY
git.cmd push --force            DENY    DENY
nh.exe merge-stack run          DENY    DENY
git status                      ALLOW   ALLOW   (control)
github --help                   ALLOW   ALLOW   (control: no over-match)
```

`rm.exe` is the sharp one: the others already denied through a second path, so it is the spelling with no net under it, on the command where a miss costs the most.

## What makes the test worth having

Each row asserts the **bare** form denies before checking the suffixed spellings — "control: bare ... must deny or the row is vacuous". That is the difference between a table that measures something and one that agrees with itself. And `test_the_fragment_is_derived_from_the_one_suffix_source` pins `EXE_SUFFIX_RE` against `_EXECUTABLE_SUFFIXES`, so the argv side and the raw-text side cannot drift apart later — with `not ... search("github ")` as the over-match control.

Choosing a behavioural guard over a scan of the regex source, and saying why — "a source scan of the patterns turned into a regex-of-regex that mis-read grouped alternations, so it was dropped for this" — is the right disposition. You tried the stronger instrument, it lied, you wrote down that it lied. That belongs in the file.

## One residual, stated accurately by you already

The coverage list is per known binary, so a raw-text matcher added later for a NEW binary would not be covered until someone adds a row. Your docstring says exactly that ("A new raw-text matcher for a known binary that forgets the suffix reopens one of these and goes red"), so nothing is overclaimed — worth keeping in view if the matcher set grows.

```
full suite   12991 passed, 1 failed
```

That one failure is `test_repo_discovery.py::test_untracked_file_counts_as_dirty`, and it is not yours: it passes on clean `main`, passes on your tree when run directly (`53 passed`), and your diff touches only `guard.py`/`exec_names.py` and their tests. A flake under a loaded parallel run.

Clean as of `5032e74c`.

eyalgolan:
Landed as `e02dfebb`, with you as the commit author, closing #406.

Sreekant13:
Thanks for landing it, and for re-measuring on the squashed tree rather than the branch.

On the residual you flagged: agreed, the coverage table is per known binary, so a raw-text matcher added later for a NEW binary stays uncovered until someone adds a row. I kept it behavioral on purpose after the source-scan instrument lied (it read the grouped alternations `(?:nh|no-human)\s` as ending at the group, not the name, and went vacuously green), so the enumeration is by hand rather than from the patterns.

If the matcher set grows enough that hand-maintained rows become the weak link, the right follow-up is a guard that derives the binary set FROM the raw-text patterns themselves and asserts each carries `EXE_SUFFIX_RE` before its `\s` — the source-scan done correctly, walking the alternation groups instead of assuming a bare name. Happy to pick that up as its own small PR whenever the set moves; until then the behavioral table is the honest instrument. Will keep it in view.

PR #409, opened 2026-09-15
Make the exec_names fold tests host-independent, not POSIX-only

### LINKED ISSUE #408: Four exec_names fold tests pass only on a POSIX runner: _verdicts_with_fold never pins _IS_WINDOWS, and the removed-cwd repro cannot run on Windows
Four tests in `tests/test_exec_names.py` pass on the Linux CI runner and fail on a Windows one. The Windows CI job runs only three named test files (`test_windows_portability.py`, `test_windows_credential_file.py`, `test_check_release_manifest.py`), so CI never exercises this module on Windows and nothing catches it. A contributor who runs the full suite on Windows sees four reds that are test artifacts, not guard bugs, which is the opposite of the property #345 asks for: the fold tests should stay checkable from either kind of machine.

Two distinct causes, both in the test harness. No source is at fault.

## Cause 1: `_verdicts_with_fold` pins the fold probe but not `_IS_WINDOWS`

The helper simulates a case-sensitive host by patching `host_folds_case` to `False`. But `command_name` folds executable case whenever `is_windows` is set, independent of the volume probe:

```python
if fold_case is None:
    fold_case = is_windows or host_folds_case()
```

That is correct: `GIT.EXE` and `git.exe` are one file on Windows regardless of the volume's case flag. The consequence is that a `fold=False` row asserting a case-sensitive host allows an uppercase spelling only holds when the runner's own `_IS_WINDOWS` is `False`.

Measured on a Windows host, faithful subprocess order (patch `host_folds_case` before importing guard):

```
fold=False, inherit _IS_WINDOWS=True     fold=False, pin _IS_WINDOWS=False
  DENY   GH pr merge 7                      ALLOW  GH pr merge 7
  DENY   GIT push origin main               ALLOW  GIT push origin main
  DENY   FIND / -delete                     ALLOW  FIND / -delete
  DENY   gh pr merge 7                       DENY   gh pr merge 7   (correct)
```

The right column is the POSIX CI reading. Three tests depend on it: `test_a_case_sensitive_host_is_not_punished`, `test_a_case_sensitive_host_keeps_the_documented_table`, and `test_the_runner_recursion_folds_a_wrapped_name_for_git_too`.

## Cause 2: `test_the_probe_survives_a_removed_process_cwd` needs to delete the process cwd

The test does `os.chdir(doomed); os.rmdir(doomed)` to reproduce "the worktree was cleaned up while the agent is still chdir'd into it", then asserts `guard.evaluate` still returns a verdict instead of crashing. On Windows a directory that is a process's current working directory is locked, so:

```
os.rmdir(doomed)
PermissionError: [WinError 32] The process cannot access the file because it
is being used by another process: '...\\doomed'
```

The precondition cannot be produced with real filesystem calls on Windows at all, so the assertion is never reached.

## Fix in #<PR>

Test-only.

1. `_verdicts_with_fold` pins `_IS_WINDOWS = False` in its subprocess, after the guard import (the patterns are already compiled; `_IS_WINDOWS` is read per call). On a POSIX runner `_IS_WINDOWS` is already `False`, so this changes nothing there and CI stays green; it only brings a Windows runner in line.
2. `test_the_probe_survives_a_removed_process_cwd` is skipped on `win32` with a reason. The guard it pins (`_candidate_anchors` catching `OSError`) is not OS-specific and the POSIX run exercises it. A mock-`getcwd` repro is deliberately not substituted, because this test's stated contract is that a REAL removed cwd, not a patched `os.getcwd`, is the repro. `sys.platform` is used rather than `os.name`, so the skip does not trip `test_no_test_asserts_the_permissive_fallback`, whose needle is the deleted probe fallback (a different concern from a portability skip).

After: `tests/test_exec_names.py` is 79 passed, 4 skipped on Windows (was 4 failed), and unchanged on POSIX.

If you would rather the removed-cwd test keep running on Windows through a patched `os.getcwd` that raises `FileNotFoundError`, rather than skip, say so and I will switch it; I chose the skip to honor the docstring's "not a mock" stance.

Sreekant13:
Closes #408. Refs #345.

Four tests in `tests/test_exec_names.py` pass on the Linux runner and fail on a Windows one. The Windows CI job runs only three named test files, not this module, so nothing catches it; a contributor running the full suite on Windows sees four reds that are test artifacts, not guard bugs. Test-only, no source change.

## Cause 1: the fold helper pins the probe but not the OS axis

`_verdicts_with_fold` simulates a case-sensitive host by patching `host_folds_case` to `False`, but `command_name` folds executable case whenever `is_windows` is set:

```python
if fold_case is None:
    fold_case = is_windows or host_folds_case()
```

That is correct (`GIT.EXE` and `git.exe` are one file on Windows), so a `fold=False` row asserting a case-sensitive host allows an uppercase spelling only holds when the runner's `_IS_WINDOWS` is `False`. The subprocess now pins it, after the guard import (patterns are already compiled; `_IS_WINDOWS` is read per call). On POSIX `_IS_WINDOWS` is already `False`, so this changes nothing there and CI stays green.

Fixes `test_a_case_sensitive_host_is_not_punished`, `test_a_case_sensitive_host_keeps_the_documented_table`, `test_the_runner_recursion_folds_a_wrapped_name_for_git_too`.

## Cause 2: the removed-cwd repro cannot run on Windows

`test_the_probe_survives_a_removed_process_cwd` does `os.chdir(doomed); os.rmdir(doomed)`. Windows locks a directory that is a process cwd, so `os.rmdir` raises `WinError 32` and the precondition is never reached. Skipped on `win32` with a reason: the guard it pins (`_candidate_anchors` catching `OSError`) is not OS-specific and the POSIX run exercises it, and the test's own contract is a REAL removed cwd rather than a patched `os.getcwd`, so a mock is deliberately not substituted. `sys.platform` is used, not `os.name`, so the skip does not trip `test_no_test_asserts_the_permissive_fallback`.

## Result

`tests/test_exec_names.py`: 79 passed, 4 skipped on Windows (was 4 failed), unchanged on POSIX. Manifest repinned, `--strict` clean, no CRLF.

If you would prefer the removed-cwd test keep running on Windows through a patched `os.getcwd` that raises `FileNotFoundError` rather than skipping, say so and I will switch it. The skip was chosen to honor the docstring's "not a mock" stance.


eyalgolan (approved):
Approving. Test-only, and the second hunk is the one that matters.

Pinning `_guard._IS_WINDOWS = False` when the row asserts a case-SENSITIVE host is a real correctness fix, not a platform workaround. `command_name` folds unconditionally when `is_windows` — `fold_case = is_windows or host_folds_case()` — so a `fold=False` row was only ever asserting anything on a POSIX runner. The fold probe alone does not simulate the whole host, and the comment says exactly that. Without the pin the test was silently host-dependent while looking like it pinned a property.

I checked the mutation cannot leak: `_verdicts_with_fold` runs the rows in a subprocess, not a monkeypatch, and its own docstring explains why (the raw-text patterns are compiled at import). So setting `_IS_WINDOWS` there stays in that process.

The skip on the removed-cwd test is argued rather than asserted, which I appreciate — Windows genuinely locks a directory that is a process's cwd, the guard being pinned (`_candidate_anchors` catching `OSError`) is not OS-specific, and the POSIX run still exercises it. Declining the mock-getcwd repro because "this test's contract is that a REAL removed cwd, not a patched os.getcwd, is what raised" is the right call; a mocked version would pass while proving something weaker.

Using `sys.platform` rather than `os.name` to avoid tripping `test_no_test_asserts_the_permissive_fallback` is a nice detail — that is the kind of interaction that usually gets discovered by a red build three weeks later.

```
tests/test_exec_names.py + tests/test_guard.py   350 passed, 2 skipped
```

Clean as of `7bfae1e6`.

PR #411, opened 2026-09-15
error_signature: strip Windows paths too, so StuckDetector stops under-firing on Windows

### LINKED ISSUE #410: error_signature strips POSIX paths but not Windows ones, so StuckDetector under-fires on Windows (same error twice fingerprints differently)
`core/bounds.py::error_signature` reduces an error blob to a stable fingerprint by stripping volatile tokens, and its docstring names "paths" among them. The path rule is POSIX-only:

```python
norm = re.sub(r"/[^\s'\"]+", "<path>", norm)        # absolute paths
```

A native Windows absolute path (`C:\Users\...`, `\\server\share\...`) has no leading `/` and uses backslashes, so it is never stripped. The same error seen on two attempts then fingerprints DIFFERENTLY on Windows whenever its text carries a Windows path with a per-run-varying, non-hex segment (a `tempfile` dir, a pytest tmp counter, a build output dir, a PID). On POSIX the whole path collapses to `<path>` and the two fingerprints match.

## Measured (on `main`)

Same error, two runs, differing only in the volatile path segment:

```
POSIX  : same error, two temp dirs      -> signatures equal? True
Windows: same error, two temp dirs      -> signatures equal? False
POSIX  : pytest tmp run counter differs -> equal? True
Windows: pytest tmp run counter differs -> equal? False
```

```python
from no_human.core.bounds import error_signature as es
# tempfile names: Python mkdtemp uses ascii_letters+digits, so not pure hex
w1 = r'AssertionError: mismatch' '\n  File "C:\\Users\\u\\AppData\\Local\\Temp\\tmpv5k2p9qz\\case.py", line 8'
w2 = r'AssertionError: mismatch' '\n  File "C:\\Users\\u\\AppData\\Local\\Temp\\tmpq8w1e7rt\\case.py", line 8'
assert es(w1) != es(w2)   # they should be equal; the error is identical
```

(A worktree id that is pure hex, `aaaa1111`, is caught by the separate `\b[0-9a-f]{8,}\b` -> `<hash>` rule regardless of separators, so that particular case masks the gap. The gap shows the moment the volatile segment is not pure hex, which `tempfile`/pytest/PID segments are not.)

## Why it matters

`StuckDetector` (§3.5) keys "the same error twice means zero progress, reset in a fresh session" on this signature: `StuckDetector.record` -> `error_signature(error_text)` (bounds.py), fed raw `result.final_text` at `core/orchestrator.py` (`stuck.record(result.final_text or reason)`). On Windows, an identical failure whose traceback points into a temp/pytest path does not register as a repeat, so the detector under-fires and the attempt keeps retrying the identical failure instead of resetting: burned turns and budget, on the platform no_human ships a desktop app for. It is the same POSIX-only-assumption class as #105 / #106 / #305 / #345, in the stuck-detection layer rather than the guard.

Two call sites already document the assumption this breaks: `_fingerprint_detail`'s docstring ("`error_signature` also collapses paths and numbers") and the attempt-dedup fingerprint comment ("error_signature ... modulo paths and numbers"). Both read as true only on POSIX. The dedup fingerprint's own payload is a git diff, whose headers are forward-slash even on Windows, so that site is largely shielded; the `StuckDetector` raw-text path is not.

## Fix (small, `bounds.py` is not frozen)

Add Windows absolute-path stripping alongside the POSIX rule, before it so a drive-letter prefix collapses with the rest rather than being left behind:

```python
norm = re.sub(r"(?<![a-z0-9_])[a-z]:[\\/][^\s'\"]*", "<path>", norm)  # C:\... or C:/...
norm = re.sub(r"\\{2}[^\s'\"]+", "<path>", norm)                       # \\server\share\...
norm = re.sub(r"/[^\s'\"]+", "<path>", norm)                           # POSIX absolute paths
```

The negative lookbehind keeps a real drive letter (`c:\`) matching while a trailing colon inside a longer token (`hklm:\`, `note:/`) does not. `norm` is already lowercased, hence `[a-z]`. Relative backslash paths stay unstripped, exactly as relative POSIX paths already are, so the two platforms match rather than one over-collapsing.

I can open a PR with this plus a Windows analogue of `test_signature_stable_across_volatile_tokens`, if you would like it.

Sreekant13:
Closes #410.

`error_signature` reduces an error blob to a stable fingerprint by stripping volatile tokens, and its docstring names "paths" among them. The path rule was POSIX-only:

```python
norm = re.sub(r"/[^\s'\"]+", "<path>", norm)        # absolute paths
```

A native Windows path (`C:\Users\...`, `\\server\share\...`) has no leading `/` and uses backslashes, so it was never stripped. The SAME error seen on two attempts then fingerprints differently on Windows whenever the path carries a per-run-varying, non-hex segment (a `tempfile` dir, a pytest tmp counter, a PID). A pure-hex worktree id is caught by the separate `\b[0-9a-f]{8,}\b` -> `<hash>` rule regardless of separators, which is why the gap only surfaces for non-hex segments.

## Measured

```
POSIX  : same error, two temp dirs -> equal? True
Windows: same error, two temp dirs -> equal? False   (before)
```

## Why it matters

`StuckDetector` (§3.5) keys "the same error twice means zero progress, reset in a fresh session" on this signature, fed raw `result.final_text` through `stuck.record`. On Windows an identical failure whose traceback points into a temp/pytest path did not register as a repeat, so the detector under-fired and the attempt kept retrying the identical failure instead of resetting: burned turns and budget, on the platform the desktop app ships for. Same POSIX-only-assumption class as #105 / #106 / #305 / #345, in the stuck-detection layer rather than the guard. Two call sites already documented the assumption this breaks (`_fingerprint_detail`'s docstring and the attempt-dedup fingerprint comment, both "collapses paths").

## Fix

Add drive-letter and UNC rules before the POSIX one, so a drive or UNC prefix collapses WITH the rest instead of being left behind:

```python
norm = re.sub(r"(?<![a-z0-9_])[a-z]:[\\/][^\s'\"]*", "<path>", norm)  # C:\ or C:/
norm = re.sub(r"\\{2}[^\s'\"]+", "<path>", norm)                      # UNC \\srv\share
norm = re.sub(r"/[^\s'\"]+", "<path>", norm)        # POSIX absolute paths
```

`norm` is already lowercased, hence `[a-z]`. The negative lookbehind keeps a real drive letter (`c:\`) matching while a trailing colon inside a longer token (`hklm:\`, `note:/`) does not. Relative backslash paths stay unstripped, exactly as relative POSIX paths already are, so neither platform over-collapses. `bounds.py` is not frozen, so the rules land in `error_signature` directly.

## Tests

`test_signature_stable_across_windows_paths` (backslash tempfile, UNC share, forward-slash drive path) and `test_signature_does_not_overcollapse_a_trailing_colon_token`, next to the existing POSIX volatile-token test. All signature tests green; no new failures against the same-checkout baseline (the pre-existing `parse_quota_reset` and `failed_restoration` failures on this host are unrelated and identical before and after). Manifest repinned, `--strict` clean, no CRLF.


eyalgolan (approved):
Approving. The thing I most wanted to know about a change to a fingerprint function is whether existing fingerprints move, and they do not.

I ran the same blobs through `error_signature` on `main` and on this branch:

```
                     MAIN               PR411
posix traceback      7404c2ea8240cd29   7404c2ea8240cd29   unchanged
posix tmp varying    f89881cdee1f6444   f89881cdee1f6444   unchanged
no path at all       014f79990f7e22dd   014f79990f7e22dd   unchanged
url in message       46473d36718dfdef   46473d36718dfdef   unchanged
registry HKLM:\      53324d964038f1d6   53324d964038f1d6   unchanged
windows drive        4c2f1876049ce8cb   f89881cdee1f6444   now == posix tmp varying
windows unc          0bce33203d877094   7404c2ea8240cd29   now == posix traceback
```

That last pair is the result worth stating plainly: the same logical error now fingerprints **identically** on Windows and POSIX, which is the property `StuckDetector` needs. And nothing that already had a signature moved, so no history is invalidated.

The `HKLM:\` row is the one I went looking for a bug in. `(?<![a-z0-9_])[a-z]:[\\/]` is doing real work there — a naive `[a-z]:[\\/]` would have collapsed a registry path into `<path>` and quietly merged unrelated errors. Good that the lookbehind is there and better that the comment says why.

Ordering the Windows rules before the POSIX one is also correct rather than incidental: the POSIX rule would otherwise eat only the run after the first `/` and leave the drive prefix behind, producing a third distinct signature. The comment covers that too.

```
tests/test_bounds.py                      72 passed
test_signature_stable_across_windows_paths  fails on main, passes here
```

Clean as of `7bfae1e6`.

Issue #332, opened 2026-09-13
The version a user is running is readable in exactly one place, and it is not About


eyalgolan:
Observed on the installed Windows app at 0.2.1, during release-walk baselining
(screenshots taken and read, not inferred):

- **Settings → Updates** displays `no_human 0.2.1`. This is the ONLY surface in
  the app that shows the running version.
- **About** shows no version string at all. Its sections are: *What no_human is*,
  *Documentation*, *Keyboard shortcuts*, *Contact*.

The binary agrees with the Updates card (`FileVersion 0.2.1`,
`ProductVersion 0.2.1.0`, `BUILD_STAMP commit=a5641fe6 dirty=no`), so nothing is
misreporting. The problem is placement, not correctness.

**Why this is more than a UX nit.** About is where a user looks for "what am I
running" — it is the conventional home for it, and it is the page a person opens
when filing a bug report. Putting the version only in Settings → Updates means:

1. A user reporting a problem usually cannot tell us their version without being
   walked to a pane they had no reason to open.
2. Our own release verification has a single point of failure. Confirming that an
   upgrade actually took effect currently depends on one card in one pane. A
   release where the binary is 0.2.3 and a surface still renders 0.2.2 is
   invisible to a checksum, and right now there is only one place that would
   reveal it.

**Acceptance (properties, not mechanisms):**

1. A user can read the running version from the About surface without opening
   Settings.
2. Every surface that displays a version derives it from one source, so two
   surfaces cannot disagree after an upgrade.
3. A test fails if a displayed version can diverge from the packaged build's
   version.

Not a release blocker. Found while establishing the 0.2.1 baseline for the
v0.2.3 three-platform release walk.

PR #412, opened 2026-09-15
Wire nh task add --follows and warn nh approve off superseded tasks

### LINKED ISSUE #232: follows_id is plumbed end to end and NULL on every task; nh task add cannot set it
## What is wrong

`follows_id` exists to record "this task supersedes that one". It is wired
through the whole stack and has never held a value.

```
sqlite> select count(*) total, sum(case when follows_id is not null then 1 else 0 end) with_follows from tasks;
996|0
```

The plumbing is real, not vestigial:

* `core/task.py:518,526,536,560,586` — field, constructor, serialisation, row hydration
* `core/db.py:915` — column, commented "sibling link, NOT parent_id's compound-child relation"
* `core/db.py:2000,2165` — persisted in both update statements
* `web/src/TaskComposer.jsx:467` — the web follow-up composer sets `followsId`

But `nh task add --help` has no `--follows` / `--supersedes` flag. Every task
filed from the CLI — which is how the supervising session files all of them —
must leave it null.

## Why it matters, measured

Three live tasks supersede an earlier one, and all three relationships exist
only as English inside `description`:

| superseded | superseded by |
| --- | --- |
| 86b5bf3d (has an open PR) | 3bc56444 |
| 717d7da5 | 90549239 |
| fcbd9f3c | bae495e4 |

I nearly landed the superseded task's pull request. The reasoning that led
there was sound on its face — that task has an attempt with `review_passed=1`
and an open PR, so the coder is done — and it is wrong only because a *later*
task replaced it. No query can find that. `select ... where follows_id = ?`
returns nothing for every task in the database, so the only way to discover it
is to read the prose of every sibling's description and notice a task id.

The near-miss is the point: this is a correctness trap that scales with the
size of the board, and the field designed to prevent it is inert.

## Suggested direction

Add `--follows <task-id>` to `nh task add` and populate the column, then make
the supersede relationship visible where a landing decision is actually made —
minimally, refuse or warn when approving/landing a task that something else
declares it follows. The web composer already produces the value, so the model
and storage need no change; this is a CLI flag plus a read at the decision
point.

Filed after a near-miss, not a failure — nothing was landed wrongly.

PrinceXDev:
## What this changes

Adds --follows <task-id> to nh task add (the CLI gap issue [#232](https://github.com/no-human-ai/no_human/issues/232) names directly — follows_id was already wired through the DB/model/API, but nothing could set it) and makes nh approve refuse — with --force-superseded as the explicit override — a task that a later task's follows_id already names as followed-up-on, in both the single-task and --ready --yes paths.

Fixes [#232](https://github.com/no-human-ai/no_human/issues/232)

## Evidence

> $ ./.venv/Scripts/python.exe -m pytest -q tests/test_task_follows.py tests/test_api_task_follows.py tests/test_task_lifecycle.py tests/test_cli_commands.py tests/test_approve_ready_cli.py tests/test_approve.py -m "not slow and not nightly" --deselect tests/test_task_lifecycle.py::test_config_path --deselect tests/test_cli_commands.py::test_logs_shows_and_tails_the_verification_artifact_when_present --deselect tests/test_cli_commands.py::test_agents_table_shows_BURN_not_non_cache_coder_tokens --deselect tests/test_cli_commands.py::test_stop_happy_path --deselect tests/test_cli_commands.py::test_stop_wedged_escalates_to_sigkill --deselect tests/test_cli_commands.py::test_stop_only_targets_pidfile_pid --deselect tests/test_cli_commands.py::test_stop_race_exits_before_sigterm_delivered --deselect tests/test_cli_commands.py::test_stop_sigkill_exhausted_keeps_pidfile --deselect tests/test_cli_commands.py::test_stop_keeps_pidfile_while_process_alive
> 294 passed, 9 deselected in 234.51s (0:03:54)

The 9 deselected tests fail identically on unmodified main (verified via git stash/re-run) — pre-existing Windows-environment noise (pidfile/signal mocking, console line-wrap), not caused by this change.

## Checklist

- [ ] Lockfiles committed if a dependency changed.
      checks for the file.)


eyalgolan (approved):
Approving. This does both halves of #232, and the second half is the one that actually prevents the near-miss the issue was written about.

`--follows` resolving to a real task before filing is right — `find_task` first, `print_no_task_matching` on a miss — so a typo cannot store a dangling id in a column whose whole value is that it can be queried.

The refusal is the part I checked hardest, because a new gate in the approve path can block legitimate landings. Two things make it safe:

`_superseding_successors` excludes a successor that was itself cancelled (FAILED with a `cancel_reason`) and counts everything else including DONE — which is the right polarity, since DONE is the strongest evidence the predecessor's work was superseded, not the weakest.

And the blast radius on existing tasks is zero by construction: the refusal can only fire where `follows_id` is set, and nothing could set it before this PR. That is why the issue could say "996 tasks, 0 with follows" — every task on the board today passes the new gate untouched.

`--force-superseded` is named in the refusal text itself rather than only in `--help`, so the operator reads the way out at the moment they are stopped, and it is wired through both the single-task and `--ready --yes` paths rather than just the one a test happened to cover.

```
tests/test_task_follows.py + cli + approve_ready   251 passed
full suite on the squashed tree                    13063 passed, 0 failed
--strict alone                                     OK: 1629 file(s)
citations                                          VERDICT=OK
```

Fixes the CLI gap #232 named directly. Clean as of `e02dfebb`.

eyalgolan:
Landed as `b6e98517`, with you as the commit author, closing #232.

Verified on the squashed tree, not just the branch — full suite green there, and the behaviour re-checked on that tree.

Thank you, and specifically for taking two issues that were labelled good-first-issue and treating them as real design problems rather than minimum patches. #232's value was never the flag; it was making the supersede relationship something a landing decision can see. #332's was never the About line; it was the second surface resolving through the same source so the two cannot drift after an upgrade. Both of you went for the part that mattered.

PR #413, opened 2026-09-15
The version is readable from About, and from one source (#332)

### LINKED ISSUE #332: The version a user is running is readable in exactly one place, and it is not About

PrinceXDev:
## What this changes

About now shows the running version, so it can be read without opening Settings > Updates — and both surfaces resolve it through one module (`runningVersion.js` + `useRunningVersion.js`) instead of two copies of the shell-then-server precedence that could disagree after an upgrade. Fixes #332.

The chain from the packaged build to the screen is now pinned end to end. `tests/test_api_version.py` already tied `__version__`, `pyproject.toml` and `/openapi.json` together, and `desktop/packagedFiles.test.mjs` already tied `desktop/package.json` to `pyproject.toml`; nothing tied the UI to either. `web/src/runningVersion.test.mjs` does, and fails if a surface grows a version literal or starts reading `window.nhDesktop.version` / calling `fetchVersion` on its own.

image

Two behaviour changes beyond the new block. The old precedence was `desktop?.version ?? versionInfo?.version`, and `??` falls through on null/undefined only — an empty string beat a server that had actually answered, which is what a blank `--nh-app version=` sends the preload. Empty, blank and non-string values are rejected now. And when neither source answers, About says "Version unavailable" rather than naming one; it never prints "unknown", matching what `updateNotice.js` already decided about that word.

`AboutVersion` is `React.createElement` in a `.js` module rather than JSX, following `drainChip.js`'s `PausedIndicator`: `node --test` cannot parse JSX, and that split is what lets the markup be asserted in a renderer instead of by regexing `About.jsx`.

No CSS was added — the block reuses `.nh-about-block` and its existing `code` rule. No dependency was added.

## Evidence

Web board suite, this branch:

    $ cd web && npm run build && npm test
    ℹ tests 1745
    ℹ pass 1745
    ℹ fail 0

Same command on `main` (HEAD~1), same machine, for the count comparison:

    ℹ tests 1732
    ℹ pass 1732
    ℹ fail 0

+13 tests, no failures either side.

The new module and the existing Updates tests together:

    $ cd web && node --test src/runningVersion.test.mjs src/updateNotice.test.mjs
    ℹ tests 35
    ℹ pass 35
    ℹ fail 0

Server-side version literals (the other end of the chain this PR pins to):

    $ .venv/Scripts/python.exe -m pytest -q tests/test_api_version.py
    7 passed in 3.33s

Desktop version parity, unchanged by this PR but cited by it:

    $ cd desktop && node --test packagedFiles.test.mjs
    ✔ the version handed to the board is the real one, not npm_package_version
    ✔ parity: one version source, with no platform-specific override
    ✔ parity: the app version and the frozen server's version cannot drift

`RELEASE_MANIFEST.txt` re-pinned by hand for the six files (never `--write`).
Verified the way the `File inventory` job verifies — hashing the committed
blobs, not the working tree:

    tracked files: 1630
    manifest rows: 1629
    hash mismatches     : 0
    tracked but unlisted: 0
    listed but missing  : 0

End-to-end against the real server (`nh start`, board on:8420), not a mock:

    $ curl http://127.0.0.1:8420/api/version
    {"version": "0.2.3", "dist_name": "no-human", "published": true}

    About            → "Version / no_human 0.2.3 / Include this when you
                        report a bug."
    Settings>Updates → "You are running no_human 0.2.3 in a browser. Upgrade
                        the command line with: pip install --upgrade no-human"

Both surfaces, one page load, same number.

## Checklist

- [x] AI-assisted portions are disclosed above, and I can defend every line (see "AI-assisted contributions" in `CONTRIBUTING.md`).
- [x] First PR only: I have read `CLA.md` and added my `contributors/<handle>.md` entry to this PR. (Ticking this box is not the record — the file is. CI checks for the file.)


eyalgolan (approved):
Approving. #332's three acceptance criteria, each met by something I could check rather than take on trust.

About shows the running version, so it is readable without opening Settings → Updates — which was the actual complaint, since About is where a person looks when filing a bug report.

Both surfaces resolve through one module (`runningVersion.js` + `useRunningVersion.js`) instead of two copies of the shell-then-server precedence. That is the criterion that mattered most: two copies that agree today are a release away from disagreeing.

And the precedence is pinned by tests that read like someone thought about the failure modes rather than the happy path — an empty or blank shell version losing to a server that really answered, a non-string from either source ignored rather than rendered, whitespace trimmed rather than printed, and "neither source answered: null, never an invented version". That last one is the right instinct: a version string a user might quote in a bug report must never be fabricated.

```
web: runningVersion + updateNotice   35 pass, 0 fail
full suite on the squashed tree      13052 passed, 0 failed
--strict alone                       OK: 1632 file(s)
citations                            VERDICT=OK
```

One note on my own process rather than your code: my first run of your web tests failed with `Cannot find package 'react'`, which was my worktree having no `node_modules` — not your bug. Mentioning it because if you ever see me report a JS failure, it is worth asking whether I installed deps first.

Fixes #332. Clean as of `e02dfebb`.

eyalgolan:
Landed as `ee6e31f2`, with you as the commit author, closing #332.

Verified on the squashed tree, not just the branch — full suite green there, and the behaviour re-checked on that tree.

Thank you, and specifically for taking two issues that were labelled good-first-issue and treating them as real design problems rather than minimum patches. #232's value was never the flag; it was making the supersede relationship something a landing decision can see. #332's was never the About line; it was the second surface resolving through the same source so the two cannot drift after an upgrade. Both of you went for the part that mattered.

PR #414, opened 2026-09-15
manifest_repair resolves a real interpreter for its five script spawns

### LINKED ISSUE #402: manifest_repair.py has 5 more [sys.executable, ...] sites that break in the frozen desktop build (4th instance of the same scar)
Split out of the `nh approve` P0 hotfix so that change stays scoped to the landing path.

`src/no_human/vcs/manifest_repair.py` has five `[sys.executable, ...]` shell-outs. In the PyInstaller desktop build `sys.executable` is the frozen `nh` binary, not a Python, so each one re-enters the click CLI and dies in its own argument parser:

```
$ /Applications/no_human.app/Contents/Resources/nh-server/nh -m pytest -q
Error: No such option '-m'.
$ ... nh scripts/check_release_manifest.py --strict
Error: No such command 'scripts/check_release_manifest.py'.
```

The five sites, on `main`:

```
manifest_repair.py:242   approve_pending_pins        [sys.executable, str(guard), "approve", "--all", "--prune"]
manifest_repair.py:537   write_pending_manifest      [sys.executable, str(script), "--write"]
manifest_repair.py:683   _repair_by_manifest_write   [sys.executable, str(guard), "approve", *pinned]
manifest_repair.py:696   _repair_by_manifest_write   [sys.executable, str(guard), "approve", *pinned]
manifest_repair.py:791   commit_with_manifest_repair [sys.executable, str(script), "--write"]
```

**Reachable from the frozen server**, which is why this is not theoretical:

```
src/no_human/core/orchestrator.py:140   imports commit_with_manifest_repair
src/no_human/core/worktree.py:655,714   imports and calls it
src/no_human/core/evidence_ledger.py:30,89
```

These run at COMMIT time inside an attempt, not at land time.

## Why it was scoped out of the hotfix rather than folded in

The landing-path sites break **every** land for **every** repo — the pytest gate runs unconditionally. These five are gated on the repo carrying `RELEASE_MANIFEST.txt` and `scripts/check_release_manifest.py`, i.e. this repo's own shape, so a customer repo never reaches them. Different blast radius, different urgency, and a landing-path hotfix should not carry a commit-path change it cannot test in the same breath.

## The fix already exists, twice over

This is the **fourth** instance of one scar. Three call sites already carry the fallback written out inline:

- `testing/repro_gate.py::_pytest_python` (~339-368) — where it produced "a confident but FALSE fail for every bugfix"
- `core/worktree.py::_builder_python` (~479-501)
- `vcs/approve_merge.py::_real_python` — added by the hotfix

None was generalised, which is precisely why it keeps recurring. The right fix here is probably not a fourth copy: promote one of them to a shared helper and have all four call it. That is a small refactor with a real test surface, and worth doing deliberately rather than inside an incident.

## What a test has to do to catch it

The existing suites could not see this class: they assert the **argv list** through a seam without executing it, and an argv assertion cannot tell that `argv[0]` is not an interpreter. A test has to force the frozen shape (`sys.frozen` / `sys.executable` monkeypatched to the bundle path) and assert `argv[0]` is not that binary — which is what the hotfix's five new tests do for the landing path.

L4XB:
## What this changes

Fixes #402.

`vcs/manifest_repair.py` spawns a Python script five times. In the PyInstaller
desktop build `sys.executable` is the frozen `nh` binary, not a Python, so each
one re-entered the click CLI and died in its own argument parser — at COMMIT
time, inside an attempt, on any repo carrying `RELEASE_MANIFEST.txt`.

This is the fourth instance of one scar, so rather than write the fallback out
a fourth time it is promoted to `proc.real_python`, and the three existing
copies (`repro_gate._pytest_python`, `core/worktree.py::_builder_python`,
`approve_merge._real_python`) now delegate to it. Each keeps its name,
docstring and test seam — the merge gate's tests pin `_real_python`, and the
incident docstrings are where the next reader of that incident looks.

`proc.py` was chosen as the home because it already is the "how nh spawns a
child" module and has no `no_human` imports of its own, so every layer reaches
it without a cycle. Its module docstring is widened by one paragraph.

### How I enumerated the five sites

Not by trusting the list in the issue. I walked the module's AST and printed
argv[0] for every subprocess-family call in it:

```
$ python3 - <<'EOF'
import ast, pathlib
p = pathlib.Path("src/no_human/vcs/manifest_repair.py")
tree = ast.parse(p.read_text(), filename=str(p))
for n in ast.walk(tree):
    if isinstance(n, ast.Call) and ast.unparse(n.func).startswith("subprocess."):
        a0 = n.args[0].elts[0] if isinstance(n.args[0], ast.List) else n.args[0]
        print(f"  line {n.lineno:4d}  {ast.unparse(n.func)}  argv[0] = {ast.unparse(a0)}")
EOF
  line  536  subprocess.run  argv[0] = sys.executable
  line  790  subprocess.run  argv[0] = sys.executable
  line  241  subprocess.run  argv[0] = sys.executable
  line  682  subprocess.run  argv[0] = sys.executable
  line  695  subprocess.run  argv[0] = sys.executable
```

Five calls in the file, five offenders — the module spawns nothing else, so
there is no sixth site hiding behind an alias or an indirect argv. One
correction to the issue's table: 683 and 696 are in `commit_with_manifest_repair`
(the initial approve and the count-drift retry), and 791 is
`_repair_by_manifest_write`; the issue has those two swapped.

`test_no_spawn_in_manifest_repair_uses_sys_executable_as_the_interpreter` is
that same walk, kept as a test. It is the only check in the diff that cannot be
satisfied by fixing four sites out of five, which given the history of this
scar seemed worth having.

### Sibling modules with the same shape — found, deliberately not fixed here

I ran the same AST sweep over all of `src/no_human`. Everything below is the
same shape and none of it is in this issue's scope; listing it so the next
person does not have to re-derive it:

| Site | argv[0] | Frozen-aware today? |
|---|---|---|
| `vcs/derived_conflict.py::_inventory_argv` | `sys.executable` | yes — a fourth inline copy, PATH-only |
| `vcs/budget_conflict.py:356` (`run_budget_test`) | `sys.executable` | **no** |
| `testing/citation_drift.py::_interpreter` | `sys.executable` when the repo ships no venv | **no** |
| `testing/runner.py:945` (the `pytest` repair suggestion) | `f"{sys.executable} -m pytest"` | **no** |
| `eval/northstar.py:1086`, `eval/replay.py:274`, `eval/sandbox_selftest.py:94`, `eval/funnel_corpus.py:113` | `sys.executable` | **no** |
| `testing/ui_evidence.py:863` | `sys.executable` | yes |
| `doctor.py:597-598` | `sys.executable` | n/a — its own docstring says the DMG has no provisioning path wired to it |

`budget_conflict.run_budget_test` is the closest relative: same package, reached
from `derived_conflict._resolve_in_worktree`, and gated on this repo's own shape
exactly as these five are. It is a two-line change now that `proc.real_python`
exists — say the word and I will fold it in, or open it as its own issue.
I left it out because widening a fix past its issue is how the diff stops being
reviewable, and because the eval ones want a separate judgement about whether
the eval harness should fail closed at all.

### Behaviour when there is no interpreter at all

`real_python` returns `None` only in a frozen build with no venv and no
`python3`/`python` on PATH. Each of the four resolution points keeps the
failure contract that site already had:

- `approve_pending_pins` and `write_pending_manifest` log and no-op. Both are
  advisory by contract — the file stays unpinned and fails later, honestly, by
  name — and that must not change just because the interpreter is the missing
  thing.
- `commit_with_manifest_repair` and `_repair_by_manifest_write` raise `GitError`
  reusing the `"manifest re-approve failed"` substring verbatim, so
  `is_gate_refusal` still classifies it and the checkpoint bypass seam needs no
  change.

`tests/test_egress_allowlist.py`'s `vcs/manifest_repair.py` entry is updated in
prose only: it still declares one `exec:<dynamic>` channel, and now says that
`<interpreter>` is what `proc.real_python` resolves. What runs is unchanged;
only what runs it is.

## Evidence

Merge base `d3bb252`. The base tree was checked out with `git worktree add`
and the new tests copied onto it, so the base run reads an unfixed source, not
a half-switched working tree.

**The five sites, driven through the real commit path on the unfixed base:**

```
$ cd <base worktree at d3bb252> && PYTHONPATH=$PWD/src python -m pytest -q \
    tests/test_vcs.py -k "frozen or no_spawn" -p no:randomly
FAILED tests/test_vcs.py::test_frozen_pin_maintenance_runs_the_guard_instead_of_the_nh_binary
FAILED tests/test_vcs.py::test_frozen_proactive_write_regenerates_the_manifest
FAILED tests/test_vcs.py::test_frozen_reactive_approve_repairs_the_refusal_and_commits
FAILED tests/test_vcs.py::test_frozen_reactive_manifest_write_repairs_the_refusal_and_commits
FAILED tests/test_vcs.py::test_a_frozen_build_with_no_interpreter_fails_the_commit_by_name
FAILED tests/test_vcs.py::test_no_spawn_in_manifest_repair_uses_sys_executable_as_the_interpreter
6 failed, 1 passed, 103 deselected in 7.50s
```

The base failure is the production symptom, not a proxy for it:

```
E   no_human.vcs.git.GitError: manifest re-approve failed (2): Error: No such
    command '/…/work/scripts/export_guard.py'.
```

and the AST ratchet names all five:

```
E   AssertionError: 5 of 5 spawns still pass `sys.executable` as the interpreter,
    which is the frozen `nh` binary in the desktop build:
    ['manifest_repair.py:536', 'manifest_repair.py:790', 'manifest_repair.py:241',
     'manifest_repair.py:682', 'manifest_repair.py:695']
```

The seventh test —
`test_a_frozen_build_with_no_interpreter_keeps_the_proactive_pass_advisory` —
passes on the base too, and I am not claiming it as a discriminator. It pins the
contract that the proactive pair never raise; the base reaches the same outcome
by a different route (the spawn fails), and the point of the test is that the
route change does not alter the contract.

`tests/test_proc.py` cannot run on the base at all, which is the honest report
for a file that tests a function that does not exist there:

```
E   ImportError: cannot import name 'real_python' from 'no_human.proc'
```

**On this branch:**

```
$ python -m pytest -q tests/test_vcs.py -k "frozen or no_spawn" -p no:randomly
7 passed, 103 deselected in 7.48s
```

Every test file that names a symbol this diff touches:

```
$ python -m pytest -q tests/test_vcs.py tests/test_proc.py tests/test_approve_merge.py \
    tests/test_repro_gate.py tests/test_attempt_venv_isolation.py \
    tests/test_egress_allowlist.py tests/test_structural_budget.py
335 passed in 227.96s (0:03:47)

$ python -m pytest -q tests/test_orchestrator_pr_conflict.py tests/test_checkpoint_commit_seam.py \
    tests/test_guard.py tests/test_citation_drift_preflight.py tests/test_hard_kill_salvage.py \
    tests/test_attempt_timeout.py tests/test_readme_claims.py tests/test_windows_portability.py \
    tests/test_ui_evidence.py tests/test_cancel_kills_processes.py tests/test_test_lanes.py
709 passed, 12 skipped in 216.25s (0:03:36)
```

Test count went up, not down:

```
$ python -m pytest --collect-only -q | tail -1
13215 tests collected      # base d3bb252
13229 tests collected      # this branch  (+14)
```

Manifest and citations:

```
OK: 1626 file(s) match RELEASE_MANIFEST.txt

$ python scripts/reanchor_citations.py --check
VERDICT=OK
```

The manifest re-pin is in the same commit, and its changed rows are exactly the
nine files this diff changes — no more, no fewer:

```
$ git show -U0 RELEASE_MANIFEST.txt | grep '^[+-][^+-]' | awk '{print $NF}' | sort -u
src/no_human/core/worktree.py
src/no_human/proc.py
src/no_human/testing/repro_gate.py
src/no_human/vcs/approve_merge.py
src/no_human/vcs/manifest_repair.py
tests/test_attempt_venv_isolation.py
tests/test_egress_allowlist.py
tests/test_proc.py
tests/test_vcs.py
```

**Not run:** the whole suite in one go, and the web/desktop/Playwright suites —
this diff touches no JS. I ran every Python file that names a symbol it touches
plus the two guard suites (`test_egress_allowlist.py`, `test_structural_budget.py`)
that a change of this shape can trip.

**One deliberate test edit:** `tests/test_attempt_venv_isolation.py` patched
`worktree.sys`, and `sys` is no longer imported by `core/worktree.py` once
`_builder_python` delegates. The two call sites now patch `sys` directly. It is
the same module object, so the patch is identical; the alternative was keeping a
dead import alive purely as a test handle.

## Checklist

- [x] Lockfiles committed if a dependency changed. *(no dependency change)*
      entry to this PR. *(not first PR — `contributors/l4xb.md` is already in `main`)*


eyalgolan (approved):
Approving. This took the recommendation in #402 and did the harder, right version of it — the resolution lives once in `proc.real_python` and each site keeps a thin named binding with its own reasoning, rather than a fourth copy of the fallback.

## Verified

All five `manifest_repair` call sites converted; the three `sys.executable` strings left in that file are a docstring, a comment and an error message, not call sites:

```
manifest_repair.py:261,561,710,723,821   -> [python, ...]
```

The frozen path through every binding, with the non-frozen control first:

```
NOT frozen: approve_merge -> sys.executable            (unchanged)
frozen:     approve_merge -> <repo>/.venv/bin/python
frozen:     repro_gate    -> <repo>/.venv/bin/python
frozen:     worktree      -> <repo>/.venv/bin/python
```

None of them returns the frozen `nh` binary, which is the whole point.

**The tests are load-bearing.** Against `main`'s code, `tests/test_proc.py` cannot even import (`real_python` does not exist yet) and six `test_vcs.py` tests fail, including the three that matter:

```
test_frozen_reactive_manifest_write_repairs_the_refusal_and_commits
test_a_frozen_build_with_no_interpreter_fails_the_commit_by_name
test_no_spawn_in_manifest_repair_uses_sys_executable_as_the_interpreter
```

```
full suite   13001 passed, 227 skipped, 2 xfailed, 0 failed
```

## What I liked

Keeping the wrappers instead of replacing every call with `real_python(...)` is the judgement call I would have argued for. `_real_python`'s docstring is the one a reader lands on after the incident, `_pytest_python` carries the "confident but FALSE fail" history, and `_builder_python` carries the venv-creation reasoning. Deleting them to save three functions would have scattered that history into a helper that cannot hold three different call-site rationales at once. The docstring saying so explicitly — "this wrapper stays because it is the name the merge gate's tests pin and the docstring the reviewer of that incident reads" — is the right note to leave.

`test_a_frozen_build_with_no_interpreter_fails_the_commit_by_name` is the test I most wanted to see. Fail-closed is easy to claim and easy to get wrong in the direction where the caller swallows `None` and proceeds; pinning that the commit fails *and names the condition* is what makes the fallback trustworthy.

Also right that `repro_gate` keeps its frozen fast-path locally rather than pushing it into `proc.real_python` — that path has a different question to answer (a `bin` dir vs a venv root) and folding it in would have made the shared helper carry a caller's shape.

Clean as of `d3bb252d`. I will land this shortly.

eyalgolan:
Landed as `eda10503`, with you as the commit author, closing #402.

PR #415, opened 2026-09-15
The venv guard identifies a renamed installer by what it is, not by its name

### LINKED ISSUE #341: venv guard identifies installers by resolved basename, so a renamed copy bypasses it entirely

L4XB:
## What this changes

Closes #341.

The register entry for this gap landed in #381; this closes the two rows it
measured.

`_resolve_installer` decided what a token IS from the resolved **basename**
(`_basename` -> `_is_installer_name`). A copy of a shared venv's `pip` keeps its
shebang, so it runs the same interpreter and writes into the same shared venv —
under a name no check in the module looked through. A link inside a venv's own
`bin/` went the same way, because the discard happened on the basename of the
**realpath** rather than of the name the command typed.

`_installer_by_content` asks what the resolved file **is**, and only when the
name check has already declined — so it can turn an ALLOW into a DENY and never
the reverse. A generated console script names its interpreter on line 1 and the
package it dispatches to a few lines below, and a rename or a link changes
neither:

```
#!<venv>/bin/python
# -*- coding: utf-8 -*-
from pip._internal.cli.main import main
```

Four gates before anything is read — a regular file, the execute bit, a `#!`
first line, a dispatch import from `pip`/`uv` — which is what keeps it off the
hot path: an ordinary command's path-like tokens are source files, directories
and data, and a stat plus an `X_OK` check discards nearly all of them without
opening anything.

**What comes back is the interpreter, not the script's own path.** That is the
honest answer to the question the caller is actually asking — `pip` installs
into the environment of the Python that runs it, and the copy's own location
(`/tmp`) says nothing about where the write lands — and it needs no new
machinery downstream, because `_effective_prefixes` already turns a resolved
interpreter into its owning venv. The consequence that matters: a renamed
installer whose shebang names the session's **own** worktree venv is still
allowed, exactly as its correctly-named spelling is. The verdict still turns on
where it writes, not on what it is called.

The returned path is deliberately *not* symlink-followed, unlike the rest of
`_resolve_installer`'s contract, for the reason `_effective_prefixes` already
records for `--python` values: a venv's `bin/python` is routinely a symlink to a
base interpreter living elsewhere, and following it answers "where does this
binary physically live" instead of "which venv did this invocation name".

### Why this is not the class the module exists to escape

The issue is right that a smarter *name* rule would be no fix — a basename
allowlist that also matched `notpip` would close nothing and cost the next
reader an hour. The text read here is not the command. It is the artefact that
will execute: it cannot be re-spelled, re-quoted, wrapped in a nested shell, or
laundered through a variable without changing what actually runs. That is the
same direction the module already takes with `_venv_root_of` (a `pyvenv.cfg`
probe, not a path pattern) — one layer further in.

### What it still cannot catch

The register entry is rewritten rather than deleted. An undocumented gap in a
security guard reads as a closed one, and so does a half-closed one:

| Still open | Why not closed here |
|---|---|
| the BARE-token spelling — `notpip install evilpkg` with the copy's directory on `PATH` | the bare-token branch still gates on the name before walking `PATH`. Lifting that gate reads the header of every executable on `PATH` for every token of every command — a cost and a false-positive surface out of proportion to a gap whose realistic spelling is the explicit path |
| a copied **compiled** installer | `uv` from PyPI is an executable, not a script: no shebang, nothing to read. Closing it wants content identity (inode, or a digest against a known venv's `bin/`) — the heavier of the two options the register listed |
| a copy whose shebang is rewritten to `#!/usr/bin/env python` | that form names no path, and what `PATH` resolves at exec time is already something the register says the module cannot see |
| a program that installs without dispatching to `pip`/`uv` | its own re-implementation, or a wrapper that `exec`s one |

The first three are measured against the same fixture and each has a
`test_341_open_*` test asserting it is still ALLOW, with a docstring saying:
closing one is welcome, but delete the test and its register bullet in the same
change, never the test alone. The fourth has no fixture to measure — it follows
from what the identification reads — and is documented only. So the register's
claims are machine-checked in both directions except that last row.

This is still not the capability-level fix. Today's failures remain accidental —
a shared venv on `PATH`, an inherited `VIRTUAL_ENV` — not adversarial renames,
and the last paragraph of the register (run the session with the shared venv not
writable by its UID) is what actually ends the class.

## Evidence

Merge base `d3bb252`. The base tree was checked out with `git worktree add` and
the new tests copied onto it, so the base run reads unfixed source.

**The measured table, before and after, from the same fixture** (a shared venv
with a real generated `pip` console script, a worktree with its own venv, and a
`cp` of the shared `pip` to a name of the copier's choosing):

```
                    BEFORE (d3bb252)   AFTER (this branch)
control             DENY               DENY     <shared venv>/bin/pip install evilpkg
renamed             ALLOW              DENY     cp <shared>/bin/pip .../notpip && notpip install evilpkg
symlink             ALLOW              DENY     <shared venv>/bin/pip-link -> .../notpip
own venv pip        ALLOW              ALLOW    <worktree>/.venv/bin/pip install ok
renamed own pip     ALLOW              ALLOW    cp <worktree>/.venv/bin/pip .../mypip && mypip install ok
renamed pytest      ALLOW              ALLOW    cp <shared>/bin/pytest .../runtests && runtests install hooks
no install verb     ALLOW              ALLOW    notpip --version / list / show pip
non-executable .py  ALLOW              ALLOW    notes.py install evilpkg   (imports pip, is not a program)
copied uv binary    ALLOW              ALLOW    cp <shared>/bin/uv .../notuv && notuv sync   (register row)
```

The control is the row that makes the other two mean anything, and the four
ALLOW rows below it are the false positives this fix would be worthless for
causing.

**On the unfixed base**, the two behavioural discriminators fail:

```
$ cd <base worktree at d3bb252> && PYTHONPATH=$PWD/src python -m pytest -q \
    tests/test_venv_install_guard.py -k "341 and (renamed_copy or link_whose)" \
    -p no:randomly --tb=short
E   AssertionError: a renamed copy of pip installs into the same venv
E   assert None is not None
FAILED tests/test_venv_install_guard.py::test_341_a_renamed_copy_of_the_shared_pip_is_denied
FAILED tests/test_venv_install_guard.py::test_341_a_link_whose_realpath_is_not_an_installer_name_is_denied
2 failed, 1 passed, 101 deselected in 0.71s
```

The one that passes there is the control — as it must. Across the whole `-k 341`
selection on the base: **11 failed, 8 passed**. The eight are the control, the
four false-positive guards and the three `_open_` register pins: every test that
should pass on both sides does.

**On this branch:**

```
$ python -m pytest -q tests/test_venv_install_guard.py -p no:randomly
104 passed in 1.17s
```

Every suite that names a symbol this diff touches, or imports the module:

```
$ python -m pytest -q tests/test_venv_install_guard.py tests/test_guard.py \
    tests/test_case_fold_sweep.py tests/test_windows_command_readings.py \
    tests/test_attempt_venv_isolation.py tests/test_pushed_tip_rewrite_guard.py \
    tests/test_task_spec.py tests/test_structural_budget.py
565 passed, 2 xfailed in 114.50s (0:01:54)

$ python -m pytest -q tests/test_egress_allowlist.py tests/test_readme_claims.py \
    tests/test_test_lanes.py tests/test_exec_names.py
273 passed, 14 skipped in 46.33s
```

Test count went up, not down:

```
$ python -m pytest --collect-only -q | tail -1
13215 tests collected      # base d3bb252
13234 tests collected      # this branch  (+19)
```

Manifest and citations:

```
OK: 1626 file(s) match RELEASE_MANIFEST.txt

$ python scripts/reanchor_citations.py --check
VERDICT=OK
```

The re-pin is in the same commit and changes exactly the two rows this diff
changes (`src/no_human/agent/venv_install_guard.py`,
`tests/test_venv_install_guard.py`).

**No egress change.** The diff adds one `open()` on a path the command already
named, and no exec, so `tests/test_egress_allowlist.py` needs no new channel —
and it is green.

**Not run:** the whole suite in one command, and the web/desktop/Playwright
suites — the diff touches no JS. `-n 4` was not available in the venv I used, so
every run above is serial.

## Checklist

- [x] Lockfiles committed if a dependency changed. *(no dependency change)*
      entry to this PR. *(not first PR — `contributors/l4xb.md` is already in `main`)*


eyalgolan (approved):
Approving. This closes #341 structurally, which the register entry said was the harder and right way to do it, and the way it handles what it *cannot* close is the part I want to praise.

## Measured, with both non-regression controls

I built the fixture and ran it against `main` and this branch:

```
                             MAIN    PR415
control: named pip (shared)  DENY    DENY
renamed copy                 ALLOW   DENY    <- closed
symlink to the copy          ALLOW   DENY    <- closed
session's OWN venv pip       ALLOW   ALLOW   <- not over-denied
shell script mentioning pip  ALLOW   ALLOW   <- no false positive
ordinary python script       ALLOW   ALLOW   <- no false positive
```

The fourth row is the one that would have made this unshippable if it had moved, and it does not: returning the INTERPRETER rather than the script's own path is what keeps a renamed installer pointed at the session's own worktree allowed, exactly as its correctly-named spelling is. The docstring's argument for that choice is right and worth having written down.

A confession, since it nearly cost you a wrong review: my first fixture had the shebang naming a `python` I never created, so `_probe_is_file` short-circuited and both rows still read ALLOW. I went looking for a bug in your code before checking my own instrument. The lesson is yours to keep — the gate at `_probe_is_file(interpreter) is False` is doing real work.

## Narrowing the register instead of deleting it

```
test_341_open_a_copied_compiled_installer_is_not_identified
test_341_open_a_renamed_installer_found_bare_on_path_is_not_identified
test_341_open_an_env_shebang_names_no_path_and_is_not_identified
```

Three of the four still-open sub-cases pinned as tests rather than described in prose, and the fourth marked as following from what the identification reads with no fixture to measure. That is the honest shape: a register entry that shrinks by exactly what was closed, with the remainder executable. If someone later widens the identification, those tests tell them what they just changed.

Declining the bare-token spelling deliberately — because lifting that gate would read the header of every executable on `PATH` for every token of every command — is a cost judgement I agree with, and it is stated as a judgement rather than an oversight.

`_SCRIPT_HEAD_BYTES` caps the read, so there is no unbounded file read on the guard path.

```
full suite   13006 passed, 227 skipped, 2 xfailed, 0 failed
```

Clean as of `5032e74c`.

eyalgolan:
Landed as `f07ca4bb`, with you as the commit author, closing #341.

PR #449, opened 2026-09-15
fix: enforce WINDOWS.md citation coverage

Rahul-pamula:
Summary
WINDOWS.md is now covered by the canonical citation validation. Historical citations remain backticked but are excluded from live validation using the new historical citation fence, while the live _try_kill citation now uses a stable symbol anchor.

The temporary test_windows_md_code_line_citations_resolve workaround has been removed since the canonical citation coverage now handles WINDOWS.md. The historical citation convention is also documented, and reanchor_citations.py now includes WINDOWS.md.

Testing
Testing Context: By registering WINDOWS.md in _CITATION_DOC_PATHS, the bespoke test_windows_md_code_line_citations_resolve test has been deleted because it is now fully superseded by test_the_citation_table_covers_every_line_citation_in_the_four_docs. If the live citation in WINDOWS.md rots or is missing from CITATION_TABLE, that canonical test will fail.

uv run pytest tests/test_readme_claims.py

Result: 164 passed, 12 skipped, 2 pre-existing unrelated failures in security.md citation anchors.

The two failures are unrelated to this change and were intentionally left untouched; they are caused by line drift in vcs/git.py.

No _resolve_source broadening was needed because the historical citations are intentionally excluded from live validation.


eyalgolan (approved):
Approving. You built the contract we agreed and then made it verifiable, which is the part I could not have asked for in advance.

## The payoff, measured both ways

I inserted twelve lines near the top of `cli/commands.py` and ran the re-anchor tool on your branch and on `main`:

```
PR449   DRIFT: WINDOWS.md `cli/commands.py:_try_kill:7595` -> `:7607` (re-anchoring)
        applied 8 re-anchor(s) · VERDICT=OK · second run converges
main    WINDOWS.md still says cli/commands.py:7595
```

That is the whole issue closed. Before this, `--apply` could not see the doc at all, which is why I hand-anchored that one citation three times today by locating the `signal.SIGKILL` comment and counting. It is now mechanical, and the symbol anchor is what makes the re-anchor exact rather than a guess.

## The fence does what it claims, in both directions

```
rot a HISTORICAL citation (server.mjs:96-100 -> 9999-9999)   no new failure
rot the LIVE citation (_try_kill:7595 -> 99999)              FAILS
```

Both controls, not just the one that flatters the change. A fence that only silenced things would have passed the first test and failed to be worth having; this one still bites on the citation that is supposed to be enforced.

The unbalanced-fence assertion is yours, not something I asked for, and it is the right instinct — a malformed fence would silently swallow the rest of a document, which is exactly the failure mode a "skip this region" mechanism invites.

## On the contract

Every point landed as agreed: historical stays out of validation, backticks kept, `WINDOWS.md` registered in `_CITATION_DOC_PATHS`, the bespoke test removed **in the same change** as the canonical coverage that replaces it, and the convention written down in `docs/verification.md` in language a stranger can act on —

> A citation that moves with the code is documentation (a *live citation*); a citation that records a finding at a specific point in time is *evidence*.

That sentence is the thing future contributors will actually need, and it did not exist before.

You also did not broaden `_resolve_source`, which I asked you to leave alone. Declining scope you had already investigated is harder than taking it.

## Two notes, neither blocking

**Your branch shows 2 red on its own**, both `security.md` citations into `vcs/git.py` (`_have_remote_commit:1220`, `GitRepo.fetch:1556`). They are not yours — `main` moved those lines after you branched, and the **merge result is clean**:

```
your branch alone   2 failed, 164 passed
merge result        166 passed
full suite          13074 passed, 0 failed
```

A rebase will clear them, and with this PR landed the tool will re-anchor that class for you rather than making you count.

**One to consider later, not now:** the fence is a region marker, so a live citation accidentally written *inside* a historical block would be silently unchecked. That is the same shape as the gap this PR closes, one level up. Not worth solving today — the fences are few and reviewed — but worth a sentence in `verification.md` if you are ever back in it.

Clean as of `7347088d`. And you were right that it would not take five rounds.

eyalgolan:
Landed as `8d8f8c05`, with you as the commit author, closing #259.

The measurement I care about, run on the landed tree rather than quoted from the PR: twelve lines inserted into `cli/commands.py`, then the re-anchor tool.

```
after this change   DRIFT: WINDOWS.md `...:_try_kill:7595` -> `:7607`, applied, converges
before              WINDOWS.md still says cli/commands.py:7595
```

That is the hand-anchoring gone — I did it three times by hand during landings yesterday, and it is now mechanical.

Two things you did that I want to name, because they are the reason this took one round.

You stopped and asked the semantic question instead of making ten stale references resolve. That question *was* the issue; the code was the easy half. And you tested the fence in **both** directions — a fence that only silences things passes the flattering half of the test and is worth nothing. The unbalanced-fence assertion was yours and it is the right instinct for a skip-this-region mechanism.

The sentence in `docs/verification.md` is the part that outlives the code:

> A citation that moves with the code is documentation (a *live citation*); a citation that records a finding at a specific point in time is *evidence*.

Nobody had written that down. Now it is there for the next person.

For the record, the two `security.md` reds on your branch were never yours — `main` moved those lines after you branched, and the merge result was clean. And you were right: it did not take five rounds.

PR #497, opened 2026-09-16
fix(review): preserve coverage for capped diffs

### LINKED ISSUE #437: Main review truncates large diffs at 60k chars; a changed file can go unseen and still PASS
## Problem
`_DIFF_CAP = 60_000` (`review/reviewer.py`) applies to the diff in the main review, not only the angle passes. Git's path order means `tests/` is usually what gets cut.

| 2026-09-09 to 2026-09-15 | Reviews |
|---|---|
| gate reviews checked | 269 |
| diff truncated (median full size 91k chars) | 141 |
| a changed file with neither patch nor full text in the prompt | 121 |
| of those, no tool call naming that file | 43 |
| of those 43, PASSED | 26 |

## Proposal
- Budget the diff per file, so every changed file's patch appears up to its share.
- List every file whose patch was cut, by path, in the prompt.
- If a non-excluded changed file was never inspected, route the result through the existing `ReviewerUnavailable` path (re-run, then escalate). Never PASS, and never spend one of the coder's attempts on a reviewer omission.
- Exclusions come from a trusted, documented list, not from files the coder can edit.

## Acceptance criteria
- With a diff over the cap, every changed file's patch appears up to a per-file budget and cut files are named. The test fails on main.
- A recorded review that never inspects a cut, non-excluded file does not PASS.
- Diffs under the cap produce a byte-identical prompt to today.

## Before enabling (maintainers)
Run a large clean control through the live reviewer 3 times; one coverage-rule failure kills the rule. Estimate from recent review transcripts how many reviews would newly escalate.

mhmdwaelanwr:
## What this changes

Fixes #437.

Replaces whole-diff prefix truncation in the main review gate with per-file budgeting for oversized diffs. Every changed file keeps visible patch coverage up to its share, cut paths are listed explicitly, and any required cut path that was not inspected routes through the existing `ReviewerUnavailable` retry/escalation path instead of accepting an incomplete verdict.

If even the patch headers plus the cut-path ledger cannot fit inside the cap, the review fails closed rather than silently truncating a second time. Diffs under the cap are returned byte-for-byte unchanged.

The trusted coverage-exclusion set is code-owned and currently empty; target-repository configuration cannot opt paths out of coverage.

### Expected escalation impact

I cannot independently re-derive the production review corpus from this repository. Using the figures reported in #437, 43 of 269 checked gate reviews had a changed file omitted from prompt coverage with no tool call naming that file, so this rule would have escalated up to about 16% of that reported sample.

Of those, 26 had passed, which is about 9.7% of the reported sample. These are issue-reported figures, not independently verified measurements.

### AI-assisted contribution disclosure

The implementation and regression tests were developed with AI assistance. I reviewed the submitted patch, validated the targeted behavior and adjacent git-diff security regression, and can explain and defend the changes in review.

## Evidence

Validated on the fork branch with GitHub Actions:

```text
$ uv run pytest tests/test_diff_coverage.py tests/test_reviewer.py -q
101 passed in 4.85s

$ uv run pytest tests/test_git_config_exec.py -q
7 passed in 1.09s

OK: 1645 file(s) match RELEASE_MANIFEST.txt

### CODE REVIEW

**Review by eyalgolan — DISMISSED** [review]
Welcome, and thanks — this is a strong first PR. You found the real failure mode in #437 and your fix is the one I'd have written. One red test to clear and I'll land it.

## What I verified

I ran your `budget_diff` against the two cases that matter:

```
under cap    identical=True   cut_paths=[]          <- byte-identical to today
over cap     len=60000        cut=['x.py']          <- still inside the cap
```

Byte-identical under the cap is the part I care most about. It means every review that isn't hitting the limit today behaves exactly as it did, so this can't regress the common path.

Then the decisive one — four changed files, cap exceeded:

```
new code   src/a.py  src/b.py  tests/test_z.py  tests/test_y.py   all present
old code   tests/test_z.py  ABSENT
```

That's the issue's thesis demonstrated directly. Git orders paths with `src/` before `tests/`, so blind prefix truncation drops the test files first — precisely the files a reviewer most needs to see to judge whether a change is actually covered. A reviewer was reaching a verdict having never seen them. Good catch.

Raising `ReviewerUnavailable` on `DiffCoverageError` is also the right call rather than degrading silently. A reviewer that can't see the diff should stop, not guess.

## The one blocker

`tests/test_structural_budget.py` is red on your branch:

```
review/reviewer.py: frozen 3147, now 3174 (+27); this budget only ratchets down
```

Clean on `origin/main`, so it's this branch. That budget is deliberately one-directional — `reviewer.py` is already over the file-length threshold and the freeze exists to stop it growing further.

I'd rather you moved code than raised the number, and you've already built the right home for it. Most of the +27 is the tool-input walker:

```python
stack = [payload]
while stack:
    value = stack.pop()
    ...
```

That's a self-contained pure function over a nested payload — it has no reason to live in `reviewer.py` and would sit naturally in `diff_coverage.py` next to `budget_diff`, where it's also far easier to test directly. Moving it should bring `reviewer.py` back under its frozen line and give you a unit test for the traversal itself.

## One thing worth knowing, not a blocker

Your inspection check matches on substring:

```python
inspected.update(path for path in required if path in value)
```

This establishes the path was *named* in a tool input, not that its contents were read. A reviewer that greps for the filename, or merely mentions it in a search string, marks it inspected. I'm not asking you to close that here — doing it properly means matching on tool *kind* and result, which is a bigger change than this PR should carry, and what you have is a large improvement over no check at all.

Please just soften the failure message, which currently overclaims:

```
"reviewer reached a verdict without inspecting truncated changed file(s)"
```

It can only prove the file was never referenced. Something like "without referencing" would be accurate. We have a habit here of text asserting more than the code establishes, and I'd like to not add to it. If you want to open a follow-up issue for the stronger check, I'll take it.

## Also good

`contributors/mhmdwaelanwr.md` is in the PR, so the CLA gate is satisfied — thanks for reading CONTRIBUTING properly. Your tests cover the cap boundary rather than just the happy path, which is the difference between a test that pins behaviour and one that decorates it.

Fix the budget and ping me; I'll re-run the full suite and land it.

eyalgolan:
Quick update — `main` moved 12 commits since my review, so I re-verified against it rather than leave you working from a stale verdict.

**Your branch now conflicts, and it's my doing, not yours.** I landed six task branches this evening and the only conflict is the generated pin file:

```
unmerged: [RELEASE_MANIFEST.txt]
```

Nothing in `reviewer.py` or `diff_coverage.py` conflicts — your actual change still merges clean. To clear it locally:

```
git fetch origin && git merge origin/main
git checkout --ours RELEASE_MANIFEST.txt      # take main's copy wholesale
git add RELEASE_MANIFEST.txt && git commit
```

Regenerate rather than hand-resolve — a hash file merges cleanly and wrongly, so the merged text is never right even when git is happy with it. Please don't use GitHub's "Update branch" button; it rewrites authorship in a way we then have to undo.

**The one blocker is unchanged.** On the merge result against today's `main`:

```
AssertionError: review/reviewer.py: frozen 3147, now 3174 (+27); this budget only ratchets down
```

Same +27, so nothing about my earlier suggestion has gone stale — moving the tool-input walker into `diff_coverage.py` should still bring it back under the freeze. Note the assertion now reports from `tests/test_structural_budget.py:2561` rather than `:2505`; the file shifted under us, which is exactly why I re-ran it instead of quoting my earlier output.

Everything else I verified still holds. No rush, and thanks for a clean first contribution — ping me when it's pushed.

eyalgolan:
Re-verified against today's `main` (`a00adde9`, Release 0.2.4) so you are not working from a stale verdict. Two updates, and one of them is a correction of mine.

**The conflict is unchanged** — still only the generated pin file, still my doing:

```
$ git merge origin/main
Auto-merging src/no_human/review/reviewer.py
$ git diff --name-only --diff-filter=U
RELEASE_MANIFEST.txt
```

`reviewer.py` auto-merges. Take `main`'s copy and regenerate rather than hand-resolving:

```
git checkout --theirs RELEASE_MANIFEST.txt
```

**The structural budget is now two red tests, not one.** I said one; that was accurate when I wrote it and is not any more. On the merge result with today's `main`:

```
FAILED tests/test_structural_budget.py::test_no_new_oversized_functions
  FROZEN_FUNCTION_LINES: review/reviewer.py:AdversarialReviewer.review is 301 (> 300) and is not frozen
FAILED tests/test_structural_budget.py::test_no_frozen_entry_has_grown
  review/reviewer.py: frozen 3272, now 3299 (+27); this budget only ratchets down
2 failed, 16 passed
```

`main` alone is green (`18 passed`), so both come from this branch — but the second one is not your fault, and I want to be precise about why:

```
AdversarialReviewer.review    main at my review (df81bd72)  293 lines
                              main today (a00adde9)         299 lines
                              + your 2 lines                301   <- over the un-frozen 300 cap
```

`main` grew that method by six lines while your PR sat waiting. Your change puts two lines inside it, and those two are what cross the threshold. Nothing about your patch changed.

**The good news is that the remedy I already suggested clears both.** Moving the tool-input walker out of `reviewer.py` and into `diff_coverage.py` next to `budget_diff` takes lines off the file (fixing the 3299 budget) *and* out of `AdversarialReviewer.review` (fixing the 301). One move, both tests green, and the walker becomes directly unit-testable where it lands.

Everything else in my review stands as written, including the one non-blocking ask to soften "without inspecting" to "without referencing" in the failure message.

Sorry for the extra round caused by `main` moving underneath you — that is the cost of my landing cadence, not of your patch.

eyalgolan:
I've pushed the budget fix to your branch rather than send you round again — the second red test was `main` moving under you, so it was mine to clear. Two commits: the `main` merge (`fd52c54d`) and the fix (`ba36530d`). Your work is untouched.

**What I changed.** The walker and the rejection it produces are now `InspectionTracker` in `review/diff_coverage.py`. That is where they belong rather than where they fit: `_COVERAGE_NOTE` in that same module is what names the cut paths and asks for them to be read, and this is what decides whether the ask was honoured — and it is directly unit-testable there, which it was not inline.

`reviewer.py` keeps only wiring, which took +27 down to +14:

```
                                        before   after
review/reviewer.py                        3299    3286   (frozen 3272)
AdversarialReviewer.review                 301     300   (cap 300, un-frozen)
```

`cut_paths` now initialises on the existing `full_files, omitted_files = "", []` line instead of a new one, which is what puts `review` back under its cap and off `FROZEN_FUNCTION_LINES` entirely. The remaining +14 is re-anchored in `FROZEN_FILE_LINES` with the measurement and the reason, in the form the neighbouring entries use — it is wiring that cannot live anywhere else, bought for a reviewer that can no longer reach a verdict on a diff whose test files were silently cut.

**Your non-blocking ask is done too.** The message now says "without referencing" rather than "without inspecting", for the reason I gave: the evidence is a tool call's input, so it establishes the path came up, never that its contents were read.

**Five tests on the traversal**, each mutation-checked — dropping the `tool_use` guard, flattening the nested walk, and restoring the overclaiming message each turn one of them red. One is worth calling out: the guard test uses a `denied` event, because `agent/codex_backend.py` really emits one carrying a `tool_input`. Without the guard, a reviewer whose read was **blocked by the guard** would satisfy the coverage check having seen nothing.

```
13380 passed, 38 skipped, 2 xfailed in 595.39s
```

Clean — `main` itself has one xdist-ordering flake in that run (#510) and this branch didn't even hit it.

Thank you for this one. The failure mode you found in #437 is subtle and real: git orders `src/` before `tests/`, so blind prefix truncation drops exactly the files a reviewer needs to judge whether a change is covered. That is a good first contribution.

I'm not approving my own commits, so this needs an independent pass before it lands — CI first, then a reviewer who didn't write the fix.

eyalgolan:
Green on `e3a6ab6f` (your branch with `main` merged in, which now carries #500). Every CI job passed on the first attempt:

```
success  CLA ledger      success  File inventory    success  Python
success  Web board       success  Wheel carries the board
success  ASCII locale    success  Container images build
success  Which areas a PR touches
```

I re-verified locally on that same head rather than trusting the earlier run, because the merge touched `RELEASE_MANIFEST.txt` — the one file where a clean text merge can still be semantically wrong:

```
OK: 1673 file(s) match RELEASE_MANIFEST.txt        (and 0 tracked files unpinned)
$ pytest tests/test_structural_budget.py tests/test_diff_coverage.py -q
29 passed
1 failed, 13388 passed, 38 skipped, 2 xfailed
```

That one failure is #510's xdist-ordering flake; `main` fails the identical test in the identical run, so it is not from this branch.

Worth knowing what it cost to get here, since none of it was your doing: the first green run took **three** attempts of an unchanged commit, each failing on a different wall-clock assertion elsewhere in the suite (a scheduler tick bound, then a quadratic-scan ratio), both passing locally in 1.75s. I've filed that separately — a performance property measured in absolute seconds on a shared runner is a coin flip, and external contributors pay for it.

I've dismissed my earlier `CHANGES_REQUESTED`, since the budget blocker it named is fixed and the "without inspecting" wording ask is done. **That is housekeeping, not an approval** — I wrote the fix commits on this branch, so the approving review has to come from someone who didn't. Over to the maintainer.

Thanks again for #437. The diff-ordering failure mode you found is the kind that hides indefinitely because nothing about it looks wrong.

eyalgolan:
Ran the project's own fresh-context gate over this branch rather than sign off on my own commits, and it earned its keep — it found a false negative in the check, in the part I wrote, not yours.

**The coverage check accepted the wrong file.** `path in value` marked a cut path covered whenever it appeared as a *substring* of anything in a tool input. Measured over this repository's own tracked files, 92 pairs are substrings of each other:

```
Dockerfile               inside  Dockerfile.mcp, Dockerfile.action
.gitignore               inside  web/.gitignore
README.md                inside  docs/README.md, contributors/README.md
src/no_human/api/app.py  inside  a fixture corpus's verbatim copy of it
```

A reviewer that read any of the longer ones was recorded as having covered the shorter one — in a guard whose whole job is to refuse that verdict. The input is now split into path-shaped tokens and each compared as a path: a **relative** token must *be* the required path, only an **absolute** one may carry it as a suffix (that case is why containment was there — the reviewer works in a throwaway clone). All 92 pairs fail the new rule; exact, `./`-prefixed, absolute, in-prose and `:14`-suffixed spellings of a genuine read all still pass.

**And the ledger claimed cuts that never happened.** A large prefix — commit message plus stat block — can push the raw diff over the cap while every patch still fits. `cut_paths` is empty, but the note was appended unconditionally, so the reviewer read *"these changed-file patches were cut… Inspect every listed path"* followed by nothing. Reproduced with a 900-line prefix and two small patches at a 4,000-char cap: 14,538 raw, `cut_paths == []`, header present. It is now emitted only when something really was cut.

Both fixes are mutation-checked — restoring containment turns the four wrong-file cases red, restoring the unconditional note turns the new one red.

CI also caught one of my own: the docstring I wrote named `eval/reviewer_recall` as an example, which `tests/test_reviewer_recall_guard.py` forbids any `src/` module from doing. The guard is right; the example was expendable.

**One gate finding I did not fix, deliberately.** `diff_override` still blind-truncates at `_DIFF_CAP` — the same #437 failure on the other entry point, and the one `nh gate --pr` itself uses. It is pre-existing on `main`, and it needs a decision rather than a patch: that path runs single-turn with no tools, so a ledger asking the reviewer to inspect paths is unfollowable there and the guard would reject every verdict. Filed separately with the options rather than widened into your PR.

The gate also flagged `contributors/mhmdwaelanwr.md` as an unrelated file riding along — that one it got wrong. It is your CLA ledger entry and the `CLA ledger` job requires it.

Tamper guard clean throughout: tests 13469→13483, assertions 39528→39559, no reductions. Local full suite on the final head, and CI, both reported below once green. Still not approving my own commits — that review has to come from someone who didn't write them.

eyalgolan:
**Gate verdict: PASS** on `81fd5f14`, from the project's own fresh-context reviewer (`claude-opus-4-8`), tamper guard clean — tests 13469→13484, assertions 39528→39561, no reductions.

```
$ pytest -q -n 4 -m "not slow and not nightly"     (local, same head)
1 failed, 13403 passed, 33 skipped, 2 xfailed
```

That one is #510's xdist-ordering flake; `main` fails the identical test in the identical run. CI on the same head: every job green.

It still left three notes, and one of them was about a sentence I wrote, so I've pushed `2c00e141` for it. My docstring said *"All 92 pairs fail this"*. That is true of the **relative** spellings I measured; an **absolute** token is only checked by suffix, so an absolute path to a fixture copy of a cut file still satisfies it:

```
_names_path("/root/testdata/corpus/x/base/src/no_human/api/app.py", "src/no_human/api/app.py")  -> True
_names_path("testdata/corpus/x/base/src/no_human/api/app.py",       "src/no_human/api/app.py")  -> False
```

Closing that needs the repo root to anchor against and this layer isn't given one, so the docstring now names the residual instead of implying it away, and a test pins all three cases so the sentence can't drift from the behaviour. Docstring and test only — no behaviour change from the PASSed head.

The other two are filed rather than folded in here:

- **non-ASCII changed-file paths mangle in the ledger** — `_unquote_path` runs `ast.literal_eval` over git's octal-escaped bytes, so `\303\251` becomes two code points instead of one UTF-8 character. A cut file with a non-ASCII name is then unreferenceable and the review ends `ReviewerUnavailable`. Fails closed, so it's a liveness bug, not a soundness one. It's your original code, unchanged by anything I pushed, and it deserves its own test surface rather than a seventh round here.
- **the ledger reserves worst-case space for every changed path**, not just the cut ones, so a very wide diff hits `DiffCoverageError` earlier than it needs to. The gate itself calls this "the safe direction" and it's a documented choice — left alone.

`diff_override` was raised again and correctly marked as the already-filed follow-up.

That's the review done: independent verdict PASS, CI green, tamper guard clean. I've still not approved it, because four of the commits on this branch are mine — that signature has to come from someone who didn't write them.

eyalgolan:
Landed on `main` as 36d7bf93 — squashed locally with you as author and me as committer, which is how external contributions land here (never `gh pr merge`, so this PR doesn't auto-close). Gate PASS on 2c00e141, every required check green, tamper guard clean. Thank you — #437 was a real hole and it's closed.

PR #478, opened 2026-09-16
Subprocess output decodes with the host codepage, losing non-ASCII

eyalgolan:
> **Review passed** (3 rounds) · **Tests passed** (13163 passed, 0 failed) · **Merge policy** ready

## Evidence
| Check | Result |
|---|---|
| Independent review | ✅ **PASSED** — 3 rounds · [proof](https://github.com/no-human-ai/no_human/blob/1a802e3c843fc98d91c8e6ccfa280988121ccbbf/.nh-evidence/64eab62b0ede4096b3ac72a359ca8525/review.md) |
| Verifiers | ✅ 2 of 2 satisfied · [proof](https://github.com/no-human-ai/no_human/blob/1a802e3c843fc98d91c8e6ccfa280988121ccbbf/.nh-evidence/64eab62b0ede4096b3ac72a359ca8525/verifiers.md) |
| Tests | ✅ PASS — 13163 passed, 0 failed, 0 errors · [proof](https://github.com/no-human-ai/no_human/blob/1a802e3c843fc98d91c8e6ccfa280988121ccbbf/.nh-evidence/64eab62b0ede4096b3ac72a359ca8525/tests.md) |
| CI | success |
| Merge policy | ✅ ready — 6 of 6 rules satisfied · [proof](https://github.com/no-human-ai/no_human/blob/1a802e3c843fc98d91c8e6ccfa280988121ccbbf/.nh-evidence/64eab62b0ede4096b3ac72a359ca8525/merge-policy.md) |

<details><summary>1 finding raised and addressed across review rounds</summary>

- Missed decode site: walks_provision.install_walks — walks_provision.py:52 runs `runner(step, capture_output=True, text=True, timeout=timeout)` where `runner` de

</details>
<details><summary>Verifiers (2)</summary>

- ✅ tests-assert-something — 3 files
- ✅ no-unvalidated-status-write — 6 files

</details>
<details><summary>Merge-ready policy (6 rules, source: default)</summary>

- ✅ review_passed — review PASSED on head
- ✅ tests_ran_and_passed — tests: 0 failed of 13163 run
- ✅ tamper_guard_clear — tamper guard did not fire
- ✅ repro_gate — repro gate pass
- ✅ verifiers_all_satisfied — 2 verifiers, none failed
- ✅ ci — ci: success

</details>

## Acceptance criteria
[...]

PR #503, opened 2026-09-16
GitRepo's reads decode as UTF-8, so a non-Latin title stops killing the run

eyalgolan:
**P0 on Windows.** A task titled in Hebrew commits fine and then takes the whole run down. Measured against the shipped **0.2.3** desktop build on a he-IL (cp1255) host:

```
File "subprocess.py", line 1599, in _readerthread
File "encodings\cp1255.py", line 23, in decode
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9c in position 91
...
File "no_human/vcs/git.py", line 262, in _run
AttributeError: 'NoneType' object has no attribute 'strip'
```

`git commit` echoes the subject back through `_run`; the Hebrew lamed is UTF-8 `D7 9C`, and `0x9C` is undefined in cp1255. With `text=True` alone the decode uses the **locale** codec, so it raises. On Windows `Popen._readerthread` reads each pipe in a *thread*, so the exception kills the thread rather than reaching the caller — `_communicate` joins an already-dead thread without raising, the buffer stays empty, and its tail (`stdout[0] if stdout else None`) yields `None`. `_run` then calls `.strip()` on `None`. **The user's work is left on its branch with no PR and the task never reaches approval.**

This is [468f1f12]'s fix applied to the other half of the same defect. That one fixed the **land** path (`approve_merge._sh`); this is the **attempt** path, which runs first — so on this locale the attempt died before the already-fixed gate was ever reached.

### Scope: 15 sites, all text-mode, none binary
`vcs/git.py` has 15 `subprocess.run` calls and no `Popen`/`check_output`; all 15 already passed `text=True`, and this adds `encoding="utf-8", errors="replace"` to each. Verified by AST, not by eye. Its git verbs are `merge-base`, `fetch`, `ls-remote`, `cat-file -e`, and the dynamic argv of `_run`/`_run_porcelain_lines` — `cat-file -e` is an existence probe that prints nothing, so **there is no blob or archive read in this file to corrupt**.

The kwargs go inline at each call rather than through a shared helper, deliberately: this module's own comment records that the egress-allowlist scanner resolves the git subcommand from the literal `["git", ...]` at the `subprocess.run` site, and a helper parameter turns every call into an unclassifiable `exec:git <dynamic>`. `tests/test_egress_allowlist.py` passes.

### Tests — both fail against the unfixed module
1. **Round-trip under a forced non-UTF-8 locale.** Forces cp1255 via `locale.getencoding` rather than trusting the host's locale — I verified the monkeypatch really does drive subprocess's decode, so this is **load-bearing on a UTF-8 runner too**, where the bug is otherwise invisible. It asserts the subject comes back *intact*, not merely that nothing raised: `errors="replace"` alone would satisfy a no-crash assertion while handing the caller mojibake, and a commit subject is read by humans.
2. **Per-call-site guard.** The sites are what rots — a new `subprocess.run(..., text=True)` added later would reintroduce the crash somewhere no round-trip test covers. It also asserts byte-mode calls stay byte-mode, since an `encoding` on a blob read corrupts it.

### Measured on the failing locale
Against this subset (egress-allowlist + the five git suites) on he-IL/cp1255, the change **fixes three tests that `main` fails here** — `test_changed_files_returns_raw_leading_space_and_non_ascii_paths` and two `test_git_quoted_paths` cases — and **breaks none**. Remaining failures in the subset are pre-existing on this host and unrelated: `test_git_config_exec`, `..`-through-symlink semantics, and a 100 ms `ls-remote` latency budget Windows process spawn misses (156–266 ms on `main`, 109 ms here — measured 5 runs each way, so it is not caused by this change).

### Why nothing caught it
The full Python suite runs on ubuntu, where the locale codec is UTF-8 and every one of these calls decodes correctly. The windows CI job runs only `tests/test_windows_portability.py` and `tests/test_windows_credential_file.py`, so no existing test executes this code where the codepage can raise.

### Relationship to #478
This is deliberately **not** #478. That one is a ~50-file repo-wide decode pass landing late in the 0.2.4 cut with real blast radius across `orchestrator.py` and `manifest_repair.py`; holding it was the right call. This is the surgical `vcs/git.py`-only subset that removes the crash, in the shape already proven by 468f1f12. #478's non-`git.py` remainder still stands for 0.2.5.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016ogStwaUkMhFfDeV86Qmat


eyalgolan:
Landed on main as 1221bc7b — squashed from head 96449aba8 (tree byte-identical), identity-normalized to eyalgolan on both author and committer, and Co-Authored-By/Claude trailers stripped per the repo convention. The review gate passed on all checklist items and CI run 35155198365 was green on 96449aba8. Closing.

Issue #510, opened 2026-09-16
A constraint-level reviewer-backend test silently short-circuits under pytest -n 4 (already red on main)


eyalgolan:
## Problem

`tests/test_gate_oneshot.py::test_check_credential_consults_the_reviewers_role_backend` fails under `pytest -n 4` and passes when run alone. It is reproducible, not a random flake, and it is **already failing on `main`** — I hit it while gating a landing batch and only cleared the batch by proving main fails it too.

```
main (24acd2d6), pytest -q -p no:randomly -n 4:
  1 failed, 13403 passed, 232 skipped, 2 xfailed
  FAILED tests/test_gate_oneshot.py::test_check_credential_consults_the_reviewers_role_backend

same test alone, same tree:
  1 passed in 16.53s
```

## The failure

The test asserts the gate refuses with a message naming `codex` when the reviewer role is pinned to the codex backend. Under `-n 4` it refuses earlier, for a different reason:

```
>   with pytest.raises(GateUnavailable, match="codex") as exc_info:
E   AssertionError: Regex pattern did not match.
E     Expected regex: 'codex'
E     Actual message: 'the current branch has no commits beyond origin/main'
```

So the gate is short-circuiting on a repo-state check before it ever reaches the credential check the test is about. The test builds its own repo through `_make_repo_with_origin(tmp_path)`, and under xdist that fixture repo is apparently not in the state the test assumes — the branch ends up with no commits beyond `origin/main`.

## Why it matters beyond the noise

This test exists to pin constraint §6d: a reviewer explicitly pinned to a non-default backend must be checked through that backend and must refuse rather than silently fall through to the claude-only path. Right now, in the configuration CI and every local full-suite run actually use, **that guarantee is not being tested** — the assertion never reaches the code it is meant to protect. A regression in `_check_credential`'s backend dispatch would not be caught by this test under `-n 4`.

That makes it a gap in a constraint-level guard, not just a flaky line in a log.

## Acceptance criteria

- Identify why the fixture repo has no commits beyond `origin/main` under xdist but does when run alone — a shared temp path, a worker-shared git config or template repo, or an ordering dependency on another test's cwd. Name the cause; do not paper over it with a retry or a `-p no:xdist` marker.
- The test asserts the codex refusal under `-n 4` as well as alone. It fails on main before the fix.
- Add the missing coverage the short-circuit hid: the gate reaching `_check_credential` with a codex-pinned reviewer, asserted on the message, independent of repo state.
[...]

Issue #430, opened 2026-09-15
GitHub review summaries never reach the coder, and some bot line comments count as human feedback


EllaSharakanski:
## Problem
- `fetch_github_pr_comments` (`vcs/pr_watcher.py`) reads the inline-comment and conversation-comment endpoints, but never `/pulls/{n}/reviews`. A review's summary body exists only on the review object. So a "Request changes" review whose feedback is only in the summary never wakes the task. In a sample of 52 change-request reviews on human-authored PRs across three open-source repos, about 15% carried their feedback only in the summary body. In 54%, the summary would be lost even though the line comments arrived.
- `_is_bot_author` (`blockers/wake.py`) treats only a `[bot]` login suffix or an entry in `ignore_comment_authors` as a bot. Line comments from GitHub's built-in AI reviewer come from a login without that suffix but with `user.type: "Bot"`, so they are treated as human feedback and can burn a revision round.

## Acceptance criteria (recorded API fixtures)
1. A CHANGES_REQUESTED review whose body is the only feedback wakes the task with that body.
2. A COMMENTED review with a non-empty body wakes the task.
3. An APPROVED review body does not.
4. A comment or review with `user.type == "Bot"` is not treated as human feedback by default. A new documented config setting lets a user opt a named bot back in.
5. A review body carrying the agent's own marker is filtered.
6. A review whose `submitted_at` is at or before the cursor is not fresh (reviews have no `created_at`).
7. A review body plus two line comments from the same review produce one resume and one revision-round increment.

eyalgolan:
Accepted — the first half is confirmed exactly as you describe.

`vcs/pr_watcher.py` reads `repos/{repo}/pulls/{n}/comments` (`:172`) and `repos/{repo}/issues/{n}/comments` (`:194`). Nothing anywhere in `src/` fetches `/pulls/{n}/reviews` — the single occurrence of that path is prose in a docstring (`ci_action/github.py:9`), not a call. I checked with a control, since an empty grep on its own proves nothing: the same search shape finds the comment endpoints above.

So a "Request changes" review whose feedback lives only in the summary body genuinely never reaches the task. Your ~15%-of-52 figure is the part that makes this worth fixing rather than noting — a reviewer typing their objection into the summary box is normal behaviour, not an edge case, and right now that feedback is silently dropped.

The second half, bot line comments counting as human feedback, I have not re-measured and am taking from your report.

[...]

PR #513, opened 2026-09-17
Fix acceptance criteria intake from issue sections

### LINKED ISSUE #511: nh task add <issue-url> produces an ungradable task: 0 of 56 open issues yield any acceptance criteria
## Problem

`nh task add <github-issue-url>` yields `acceptance_criteria = []` for **every issue in this repository**, because the parser accepts only GitHub task-list checkboxes and nobody here writes them.

`src/no_human/intake/github_issues.py:56,68-70`:

```python
task.acceptance_criteria = _checklist_items(body)

def _checklist_items(body: str) -> list[str]:
    """Pull GitHub task-list checkboxes (`- [ ] ...`) as acceptance criteria."""
    return [m.strip() for m in re.findall(r"^\s*[-*]\s*\[[ xX]\]\s*(.+)$", body, re.M)]
```

Only `- [ ]` / `- [x]` matches. A plain `- bullet` under a `## Acceptance criteria` heading does not. `jira.py:66`, `linear.py` and `gitlab_issues.py:57` share the same shape, so this is not GitHub-specific.

## Measured across every open issue

```
open issues scanned:                     56
  ... containing any checkbox item:       0
  ... with an "Acceptance criteria"
      heading and zero checkboxes:       25
```

Zero. Not "most" — none. Every issue this project has, written by contributors and by me, ingests with no acceptance criteria at all. The body still lands in `description`, so the criteria survive as prose and are invisible to everything that grades: the task is ungradable by construction, and nothing warns.

I filed four issues myself tonight (#502, #508, #509, #510) and every one of them is in the ungradable 25. This is not a contributor-formatting problem; it is the house convention meeting a parser that documents a different one.

## Why this is worse than it looks

The acceptance criteria are what the reviewer grades against and what `verifiers` check. A task with `[]` does not fail loudly — it proceeds, the coder works from prose in the description, and the gate has nothing specific to hold the work to. The signal most likely to catch a half-done task is simply absent, silently.

## Proposal

Either is acceptable; the second is the smaller change and closes the hole.

1. **Parse the section, not just the syntax.** Take bullets under a heading matching `#+ *acceptance criteria` when no checkbox items are present. Keeps the checkbox path exactly as it is.
2. **Refuse to create an ungradable task quietly.** When a body contains an acceptance-criteria heading and the parser finds zero items, warn loudly at ingest, naming the mismatch and the `--criteria` flag. Better a noisy ingest than a silent ungradable task.

I lean toward doing both: (2) is the safety net that works for every future shape, (1) removes the friction for the shape everyone already writes.

## Acceptance criteria

- [ ] An issue body with a `## Acceptance criteria` heading followed by plain `-` bullets ingests with those bullets as `acceptance_criteria`. Test fails on main.
- [ ] An issue body using `- [ ]` checkboxes ingests exactly as it does today, with no change to the extracted list.
- [ ] An issue with an acceptance-criteria heading from which zero items are extracted produces a visible warning naming the issue and suggesting `--criteria`, rather than a silent empty list.
- [ ] A body with neither a heading nor checkboxes still ingests with `[]` and no warning, as today.
- [ ] The same behaviour for `jira.py`, `linear.py` and `gitlab_issues.py`, or an explicit note in the PR saying why those are deliberately left alone.

## Note

Filed after another session hit this ingesting #428 and correctly cancelled the resulting defective task. The reproduction is theirs; the measurement above is mine. Until this lands, file with explicit `--criteria` flags rather than by URL — and note that any task already created from a URL in this repo has empty criteria.

Ayaan-20-11:
Fixes #511.

Issue intake previously extracted acceptance criteria only from Markdown checkboxes. This caused issues using the repository’s normal ## Acceptance criteria section with plain bullets to create tasks with empty criteria.

This change:

Preserves existing checkbox extraction exactly as the first-priority path.

Falls back to plain - or * bullets beneath an Acceptance criteria heading when no checkboxes are present.

Warns clearly when that heading exists but no criteria can be extracted, naming the issue and suggesting --criteria.

Applies the behavior to GitHub, GitLab, Jira, and Linear intake.

Preserves Jira ADF headings, list items, and task-list checkboxes during flattening so the behavior works with real Jira descriptions.

Leaves Monday unchanged because #511 explicitly scopes the change to the four adapters above.

Verification

UV_CACHE_DIR=/private/tmp/no-human-uv-cache uv run pytest -q tests/test_intake.py tests/test_jira_intake.py tests/test_intake_linear.py

146 passed in 1.74s

Syntax compilation passed.

Type check for the new extractor passed.

Manifest validation passed:

OK: 1670 file(s) match RELEASE_MANIFEST.txt

Note: the broader adapter type check still reports an existing, unrelated annotation mismatch in src/no_human/intake/jira.py:112 for httpx.get(params=...); no changed line is flagged.

AI assistance

AI assistance was used to implement the change and tests. I reviewed the behavior, ran the verification above, and can explain the implementation.


eyalgolan (dismissed):
Welcome, and thank you for this — you picked the harder of the two options in #511 and the shape you chose is the right one: one shared extractor, checkbox path untouched, section fallback behind it. I ran it against the thing the issue was actually measured on rather than against your test fixtures, and the headline result is good.

## What I verified first

I replayed every issue this repository has through both the old regex and your `extract_acceptance_criteria`, on your head, which already contains today's `main` (`a00adde9`, Release 0.2.4) — `git merge-base --is-ancestor origin/main 80e6ca44` is true, so no merge was needed:

```
open   58 issues   identical 34 (33 both-empty)   newly extracted 24   regressions 0
closed 66 issues   identical 66 (66 both-empty)   newly extracted  0   regressions 0
```

Twenty-four issues that ingested ungradable now ingest with real criteria, and **not one existing extraction changed**. That is the claim #511 makes, demonstrated on the real corpus. The checkbox path is genuinely untouched.

The warning also reaches a human, which I did not take on trust — there is no `logging.basicConfig` anywhere in `src`, so I checked that `lastResort` actually carries it after the CLI is imported:

Importing `no_human.cli.commands` first (so any CLI-time logging setup applies), then normalizing a body whose heading has prose under it:

```
stdout: criteria: []
stderr: GitHub issue o/r#1 has an Acceptance criteria heading but no extractable criteria; pass --criteria explicitly.
```

No new dependency, no credential surface, and no ReDoS in the new patterns — a 128,000-character pathological line costs 0.28 ms.

Four things to fix. Two are mechanical, two are behaviour.

## 1. CI will fail: the new module is not in `RELEASE_MANIFEST.txt`

```
$ python scripts/check_release_manifest.py --strict ; echo "EXIT=$?"
FAIL: 1 tracked file(s) not listed in RELEASE_MANIFEST.txt:
  src/no_human/intake/criteria.py: tracked but not listed
EXIT=1
```

Your PR body quotes `OK: 1670 file(s) match RELEASE_MANIFEST.txt` and that line is real — but it is the **default** mode, which downgrades an unlisted file to a warning and prints the OK anyway. CI runs `--strict` (`.github/workflows/ci.yml:216`, job "File inventory"), where the same state is exit 1. Easy trap, and partly the tool's fault for printing OK above a WARN.

```
```

## 2. The CLA ledger entry is missing

`contributors/` has no `ayaan-20-11.md`, so the `CLA ledger` job blocks the merge regardless of the code. It is a file in git, added in a commit of your own — `CLA.md` and `contributors/README.md` have the format. Read `CLA.md` first; agreeing is the point of it.

## 3. A Jira "Definition of done" list now silently replaces the acceptance criteria

This one is a real degradation, and it is caused by the ADF change in combination with the body-wide checkbox precedence. A Jira description with an Acceptance-criteria bullet section *and* a separate task list — an extremely ordinary shape:

```
## Acceptance criteria
- retries three times
- logs the failure
## Definition of done
[x] PR raised   [ ] release notes updated
```

```
main:   acceptance_criteria = []
yours:  acceptance_criteria = ['PR raised', 'release notes updated']
```

`main` produced nothing here because ADF task items flattened without markers. Yours teaches `_adf_text` to emit `- [x] ` / `- [ ] `, which is correct in itself — but `criteria.py:26-29` then takes checkboxes **from anywhere in the body** before it ever looks for the section, so the DoD list wins and the actual criteria are dropped. Nothing warns, because the checkbox branch returned a non-empty list.

Empty criteria are a known failure the reviewer can be told about. *Wrong* criteria are graded as if they were real, and the two under the heading are the ones the work will be judged against. This is worse than what it replaces.

The same precedence bug is reachable on the Markdown adapters, where it is a missed opportunity rather than a regression (`main` was equally wrong):

```
Intro
- [ ] unrelated todo
## Acceptance criteria
- the real one
-> ['unrelated todo']
```

The fix is precedence, not the ADF work: look for the acceptance-criteria section **first**, and inside it prefer checkbox items, falling back to plain bullets; only when there is no section at all should you sweep the whole body for checkboxes. That keeps #511's second criterion (a checkbox-only body ingests exactly as today) and stops a list under another heading from speaking for the section.

## 4. The safety net misses the commonest heading shape — on a real issue in this repo

`_HEADING` (`criteria.py:12-13`) anchors on `$`, so it only matches a heading whose text is exactly "acceptance criteria". Issue **#430**, open in this repository right now, is headed:

```
## Acceptance criteria (recorded API fixtures)
```

That heading never matches, the loop at `criteria.py:31` never enters, and the function falls off the end at line 49 returning `[]` **with no warning at all**. Measured across the corpus: 26 open issues carry an acceptance-criteria heading, 24 extract, 1 has checkboxes — and #430 ingests silently ungradable.

That is precisely the outcome #511's third criterion exists to prevent ("rather than a silent empty list"). Right now the warning only fires for bodies that already matched the strict heading, which is the case least in need of a safety net.

Two changes close it: match the heading on a prefix (`#+ *acceptance criteria\b`, trailing text allowed — `:` and `(...)` are both common), and warn whenever *any* heading line mentions acceptance criteria but nothing was extracted, not only when the strict pattern matched.

A behaviour battery I ran, so you can see the edges in one place — the first column is what happens today:

```
      plain                    -> ['a', 'b']
      trailing text            -> []            <- #430, silent
      colon                    -> []            <- silent
      bold, no hash            -> []            <- silent
WARN  numbered                 -> []
      nested bullets           -> ['a', 'sub of a', 'b']
      code fence in section    -> ['a', 'not a criterion']
      next heading stops       -> ['a']
      no-space heading stop    -> ['a', 'noise']
      hr in section (* * *)    -> ['a', '* *']
      checkbox elsewhere       -> ['unrelated todo']
WARN  two AC headings          -> []            <- second heading HAS bullets
      CRLF body                -> ['a', 'b']
      none at all              -> []
```

## Smaller things, not blockers

- **The second heading is never consulted.** `return []` at `criteria.py:48` fires inside the loop, so a template's empty "Acceptance criteria" stub makes the filled-in one below it unreachable. Move the `return` after the loop and keep scanning.
- **Fenced code blocks are read as criteria.** Issue bodies here routinely contain ```` ``` ```` blocks; a `- ` line inside one becomes a criterion, and a `#` line inside one closes the section early. A three-line fence toggle in the inner loop handles both.
- **Nested sub-bullets become sibling criteria.** `- a` / `  - sub of a` yields two. Sub-bullets almost always elaborate their parent; skipping indented items would be truer to intent.
- **`_HEADING` allows `##Notes`-style headings but `_ANY_HEADING` (line 14) requires a space**, so `##Notes` fails to close the section and its bullets are swept in. Make the two consistent.
- **Numbered lists.** `1. …` warns rather than extracting, which satisfies the issue — but #430 and several others here use numbers, so `^\s*(?:[-*]|\d+[.)])\s+` would cover the house style for two characters. Your call; if you skip it, the warning is the honest fallback and that is fine.
- **`monday.py:248` still has its own `_checklist_items`.** Your PR body explains it as out of scope for #511 and I accept that — but leaving one of five adapters on a private copy of the behaviour is how these drift apart again. Happy for it to be a follow-up issue rather than this PR; say which you prefer.

## Suite

Full fast suite on your head, the same command CI runs:

```
1 failed, 13380 passed, 38 skipped, 2 xfailed in 494.60s
FAILED tests/test_gate_oneshot.py::test_check_credential_consults_the_reviewers_role_backend
  Expected regex: 'codex'
  Actual message: 'the current branch has no commits beyond origin/main'
```

That failure is not yours — it is the xdist-ordering flake filed as #510, whose symptom is word-for-word "the fixture repo has no commits beyond `origin/main` under xdist but does when run alone". It passes alone on your branch (`1 passed`) and I am running the same full suite on `main` as a control; I will post the result here either way.

Your own tests are good — they pin the boundary (`stops at next heading`) rather than only the happy path, they cover all four adapters, and they add without removing anything. Fix the manifest, add the CLA file, and rework the precedence so the section wins over stray checkboxes; the heading shape is the one I care most about after that, because it is the silent case the issue was filed over.

eyalgolan:
Control result, as promised: **the suite failure is not yours.** I ran the identical command on `origin/main` alone (`a00adde9`, nothing from this PR):

```
main:        1 failed, 13374 passed, 32 skipped, 2 xfailed in 468.47s
your branch: 1 failed, 13380 passed, 38 skipped, 2 xfailed in 494.60s
FAILED tests/test_gate_oneshot.py::test_check_credential_consults_the_reviewers_role_backend   (both)
```

Same single failure, same test, on a tree that contains none of your changes — the xdist-ordering flake filed as #510. Your six extra passing tests are the ones this PR adds.

So the suite is a clean signal for this branch. The four items in my review still stand; nothing about test breakage is among them.

eyalgolan:
I've pushed the three code fixes to your branch (`88d3448f`) rather than send you round again. Your commit is untouched underneath it. **One thing is still yours and only yours: the CLA ledger file.** More on that at the end.

## What I changed, and why

**The section now outranks a stray checklist.** This was the one I'd have blocked on. Criteria were taken from any checkbox in the body *before* the heading was ever looked for. On Jira that made your `_adf_text` work into a regression rather than a gap — emitting `- [x]` markers made a "Definition of done" task list visible to the sweep:

```
## Acceptance criteria        <- plain bullets
## Definition of done         <- a task list

main:   []
yours:  ['PR raised', 'release notes updated']     <- the wrong list, silently
now:    ['retries three times', 'logs the failure']
```

Empty criteria are a known failure the reviewer can be told about. Wrong criteria are graded as if they were real. The order is now: the acceptance-criteria section, then the body's checkboxes, then nothing — so a checkbox-only issue and an unfilled template stub both still work, and the sweep loses only to a section that actually has items.

**The heading no longer has to be exact.** `_HEADING` anchored the phrase to end-of-line, so issue **#430** — open in this repo, headed `## Acceptance criteria (recorded API fixtures)` — matched nothing, fell off the end of the function, and returned `[]` **with no warning at all**: precisely the silently ungradable task #511 was filed over. The heading text now only has to *start* with the phrase, and the safety net is deliberately looser than the parser, so a bold `**Acceptance criteria**` that yields nothing warns too.

**And the section is read as markdown.** Fenced blocks are skipped (a `- ` line in a pasted diff is not a criterion; a `#` line in one no longer ends the section early), numbered items count — which is how #430 and several others here write theirs — and every acceptance-criteria heading is consulted, not just the first, so a template stub cannot hide the filled-in one below it.

## Measured, not asserted

Old regex vs. new extractor over every issue this repository has:

```
open    58 issues   gained 25   lost 0   changed 0   spurious warnings 0
closed  66 issues   gained  0   lost 0   changed 0   spurious warnings 0
104 criteria extracted in total; none of them junk
```

25, not your 24 — #430 is the new one, and it now yields all seven of its criteria.

```
1 failed, 13398 passed, 38 skipped, 2 xfailed in 645.65s
```

That one failure is #510's xdist-ordering flake, which `main` fails identically in the same run. Not yours.

`tests/test_intake_criteria.py` pins each shape against a real body from this repo. Ten of its eighteen cases are no-regression guards that hold either way; the other eight fail against the previous extractor — I ran them against your version to check they were pinning something rather than decorating it.

`RELEASE_MANIFEST.txt` now lists `criteria.py` and the new test file. Worth knowing for next time: the `OK: 1670 file(s) match` line you quoted is the **default** mode, which downgrades an unlisted file to a warning and prints OK above it. CI runs `--strict`, where the same state is exit 1.

## Two things I deliberately left alone

- Nested sub-bullets still become sibling criteria. Dropping indented items risks losing real ones, and no issue here is affected.
- `monday.py:248` still has its own `_checklist_items`. Your PR body's reasoning is right and I'm not widening scope here — I'll file it as a follow-up.

## What is still needed

`contributors/ayaan-20-11.md`. The `CLA ledger` job blocks the merge until every commit author has one, and that file is your agreement to `CLA.md` — I can fix your code, but I cannot agree to a licence on your behalf, and I won't. Read `CLA.md`, copy the template in `contributors/README.md`, commit it to this branch, and the PR is done.

Thank you for this. You picked the harder of the two options in #511 and the shape you chose — one shared extractor, checkbox path preserved, section fallback behind it — is the right one. Everything above is refinement on a correct design.

eyalgolan:
CI has now run on `88d3448f` and every job is green except one:

```
success  File inventory        <- the manifest blocker is cleared
success  Python                <- full suite
success  ASCII locale (collection)
success  Web board
success  Wheel carries the board
success  Container images build; the MCP one answers a handshake
success  Which areas a PR touches
failure  CLA ledger            <- the only thing left, and it is yours
```

So the code is done. `contributors/ayaan-20-11.md` is the single remaining step: copy the template in `contributors/README.md`, read `CLA.md` first, commit it to this branch, and the PR is mergeable.

eyalgolan:
Ran the project's own fresh-context gate over this branch, and it found a real one — in my change, not yours. Pushed as `8e21ca8c`.

**A named criteria section that yielded nothing was borrowing someone else's checklist.** I had the body-wide checkbox sweep run whenever a section produced zero items, rather than only when the body names no section at all. So:

```
## Acceptance criteria
See the linked design doc.
## Definition of done
- [x] PR raised
```

returned the Definition-of-done list as the criteria — silently, because the sweep returns before the warning fires. That is exactly the wrong-data outcome this PR exists to close, reopened through the fallback I added, and my own docstring already claimed the fallback ran "only when there is no section". The code didn't.

The sweep is now gated on no section being named. A checkbox-only issue — the case it exists for — is untouched.

```
open   58 issues   gained 25   lost 0   changed 0   spurious warnings 0
closed 66 issues   gained  0   lost 0   changed 0   spurious warnings 0
```

Same numbers as before the fix: no issue in this repository has the shape whose handling changed.

**One of my tests was pinning the defect.** `test_an_empty_heading_falls_back_to_the_body_checkboxes` asserted the borrowing behaviour, so it was protecting it. It is replaced by a parametrized pair — an unfilled stub and a prose-only section — each asserting `[]` plus the warning, and the mutation that ungates the sweep turns both red.

**And the gate asked for a Jira test I should have written.** The ADF flattener emits `## ` and `- ` as *prefixes*, so the line-based parser only works if each block already starts a line — and a heading that is the document's first node cannot demonstrate that. There is now one that follows a paragraph. It passes, so the flattener does insert the separator; that is now established rather than assumed.

The gate's third note — a section with one stray checkbox drops co-located plain bullets — it marked as acceptable and documented, and I agree; that is the checkbox-precedence rule #511 asked to preserve.

```
tests/test_intake_criteria.py tests/test_intake.py tests/test_jira_intake.py
tests/test_intake_linear.py      166 passed
```

Tamper guard clean across the branch: tests 13469→13492, assertions 39528→39564, no reductions.

Everything still green except the one thing that is yours: **`contributors/ayaan-20-11.md`**. Read `CLA.md`, copy the template in `contributors/README.md`, commit it here, and this is done.

eyalgolan:
Thanks for adding the ledger file — the content is right. One thing blocks it: `aba582c0` is authored from `ayaankapoor@Ayaans-MacBook-Pro-2.local`, which isn't attached to your GitHub account, so GitHub resolves that commit to no user at all. The ledger exists to show, later and to a stranger, *who* agreed — an unattributable commit can't do that, and no ledger file can fix it.

Your other commit uses `ayaankapoor2011@gmail.com`, which GitHub does resolve to `@Ayaan-20-11`. Re-author just that one commit with it:

```
git -c user.email=ayaankapoor2011@gmail.com commit --amend --reset-author --no-edit
git push --force-with-lease
```

I'm deliberately not doing this for you: it's your signature on a licence agreement, and a maintainer-rewritten one is the first thing an auditor would question.

Worth knowing — **your PR found a bug in the gate itself.** `CLA ledger` was going to pass you anyway, because `.[].author.login // "UNLINKED-EMAIL"` applies jq's `//` to the whole stream rather than per element, so the sentinel only appeared when *every* author was unresolvable. On your PR:

```
before   Ayaan-20-11 eyalgolan
after    Ayaan-20-11 eyalgolan UNLINKED-EMAIL
```

Fixed on `main` in `f12dedaf`. So after you amend, pull `main` in — the check is real now.

Everything else on this branch is green: full suite `13410 passed, 0 failed`, every other CI job passing, and the extractor measured over all 124 issues in this repo (25 gain criteria, 0 lose any).

eyalgolan:
Landed on `main` as 1d7edf47 — squashed locally with you as author and me as committer, which is how external contributions land here (never `gh pr merge`, so this PR doesn't auto-close).

Gate PASS on f1455bca, every required check green, tamper guard clean (tests 13489→13513, assertions 39575→39614). I pinned your CLA file in `RELEASE_MANIFEST.txt` in f1455bca — `File inventory` runs `--strict`, where a tracked-but-unlisted file is exit 1.

Thank you. #511 was measured at 0 of 56 issues yielding any criteria; it's now 25 of 58, with nothing lost. And your unattributable first attempt at the CLA commit is what exposed that the `CLA ledger` gate itself was broken — that's fixed on main in f12dedaf.

PR #529, opened 2026-09-17
Review a PR without trusting the branch its workflow came from

eyalgolan:
> **Review passed** (2 rounds) · **Tests failed** (13476 passed, 1 failed) · **Merge policy** not ready

## Evidence
| Check | Result |
|---|---|
| Independent review | ✅ **PASSED** — 2 rounds · [proof](https://github.com/no-human-ai/no_human/blob/1acc5cede5f626a05656c5537c91d385198a3ade/.nh-evidence/f5ee04d20aa14ed6a06b3c3e819474c2/review.md) |
| Verifiers | ✅ 1 of 1 satisfied · [proof](https://github.com/no-human-ai/no_human/blob/1acc5cede5f626a05656c5537c91d385198a3ade/.nh-evidence/f5ee04d20aa14ed6a06b3c3e819474c2/verifiers.md) |
| Tests | ❌ FAIL — 13476 passed, 1 failed, 0 errors · [proof](https://github.com/no-human-ai/no_human/blob/1acc5cede5f626a05656c5537c91d385198a3ade/.nh-evidence/f5ee04d20aa14ed6a06b3c3e819474c2/tests.md) |
| CI | success |
| Merge policy | ❌ not ready — 1 of 6 rules failed: tests_ran_and_passed · [proof](https://github.com/no-human-ai/no_human/blob/1acc5cede5f626a05656c5537c91d385198a3ade/.nh-evidence/f5ee04d20aa14ed6a06b3c3e819474c2/merge-policy.md) |

<details><summary>Verifiers (1)</summary>

- ✅ tests-assert-something — 2 files

</details>
<details><summary>1 failing test</summary>

- `tests/test_gate_oneshot.py::test_check_credential_consults_the_reviewers_role_backend`

</details>
<details><summary>Merge-ready policy (6 rules, source: default)</summary>

- ✅ review_passed — review PASSED on head
- ❌ tests_ran_and_passed — tests: 1 failed of 13477 run
- ✅ tamper_guard_clear — tamper guard did not fire
- ✅ repro_gate — repro gate pass
- ✅ verifiers_all_satisfied — 1 verifiers, none failed
- ✅ ci — ci: success

</details>

## Acceptance criteria
- A design document under docs/ answers all four design points (prompt injection through the reviewed diff, the tamper guard with no checked-out tree, the trigger, and the cost bound), each with a citation to a file and line in the current tree that a reader can open and check.
[...]

PR #531, opened 2026-09-17
ci: run this repository's own review gate on its pull requests

eyalgolan:
The review-gate Action shipped in 0.2.4 and nothing ran it: the four workflows
on main are this project's CI, and the product that sells a pull-request review
gate was not running one. `.github/workflows/review-gate.yml` runs it on every
`pull_request` against this repository.

`uses: ./` rather than a pinned ref, so a pull request that changes `action.yml`,
`Dockerfile.action` or `src/no_human/ci_action/` is reviewed by the revision it
ships. The two `actions/checkout` arguments are correctness rather than style:
without `ref: github.event.pull_request.head.sha` the ephemeral merge commit is
what lands on disk and `ci_action/run.py` refuses to cite line numbers against
it, and without `fetch-depth: 0` it cannot resolve the merge base from the
default depth-1 checkout; both are its exit 2. `credential_mode` and
`fail_on_findings` are left unset so `action.yml`'s declared defaults apply,
and a test asserts what those defaults are rather than trusting them to stay
put. `concurrency` with `cancel-in-progress` bounds the cost: a task branch is
pushed repeatedly while an attempt runs, and every run is an image build plus a
model call.

The credential is read from a secret named `CLAUDE_CODE_OAUTH_TOKEN`, and at
the time of writing that secret is NOT set on this repository: every
same-repository run is the Action's own fail-closed exit 2, every fork run is a
green skip, and the gate is therefore wired rather than running. docs/security.md
says so in those words rather than implying a live credential sits here. A fork pull request is not: `run.py` returns
its fork skip before it reads the `credential` input, so a fork is a green,
comment-free skip rather than a red missing-secret run. docs/security.md
records both halves of what such a secret means, including the half that is not
reassuring — push access to this repository is not limited to its maintainer,
and a same-repository pull request runs its own branch's workflow file with the
secret and no approval step.

The test compares the whole parsed workflow against an expected structure
rather than asserting the values that look load-bearing. Its first version did
the latter, and nine single-change edits were then found that break the gate for
real while leaving every assertion green — deleting `github_token` (no default
in action.yml, so run.py exits 2), misspelling the secret's name rather than the
input's, `runs-on: windows-latest` for a Docker action, referencing `./` before
the checkout that materialises it, a job-level `permissions` block that
overrides the workflow-level one at runtime, `if: false`,
`continue-on-error: true`, `dry_run: "true"`, and dropping the `${{ }}` around
the concurrency group so one constant group makes `cancel-in-progress` cancel
other pull requests. All nine turn it red.

Equality's blind spot is its oracle: `safe_load` is not GitHub's parser, and
every construct the two read differently is invisible to a comparison of what
safe_load returned. That gap was patched per construct four times — anchors,
then duplicate keys, then merge keys, then YAML 1.1 boolean spellings — and
each patch was written after an edit had already slipped past the previous one.
The fourth still let `timeout-minutes: 036` through: a YAML 1.1 octal that
PyYAML reads as exactly 30, so the structure matched, with an `int` tag the
boolean check never looked at. Numeric coercion was simply the next class, and
a guard taught one class at a time is always one class behind.

So the walk is inverted. Every scalar must be a plain string, or its exact
(path, tag, raw text) triple must be listed — five of them, each admitted at
one path and nowhere else. The path is load-bearing and an earlier draft left
it out: with the pair alone, `on:` -> `true:` destroyed the workflow's trigger
key and left all seven tests green, because both spellings were allowed
anywhere and both resolve to Python True, which a whole-document equality
cannot tell apart. The reciprocal, `cancel-in-progress: on`, was green too.
Both are red now. Default-deny cannot be one COERCION class behind, because the coercion
nobody thought of is not on the list: `0x1e` and `0.0` are both rejected here
without having been named anywhere, and `0.0` is the one that matters, because
Python's `0.0 == 0` walks it straight through dict equality. Three of the five
entries are constructs that had to be admitted by name, which is the mechanism
working rather than a concession: GitHub's trigger key is spelled `on`, which
YAML 1.1 reads as the boolean True; `pull_request:` carries an empty value; and
`cancel-in-progress` is a real boolean. `On`, `off`, `yes` and `036` are all
still rejected. An event-stream pass covers anchors, which the node graph
cannot see.

It is not a complete guard and the code says so: `!!int 30` composes to the
same (tag, raw text) pair as `30`, so an explicit tag is invisible to it, as is
a `%YAML` directive. Whether GitHub rejects either is unverified, so both are
recorded as blind spots rather than guessed at.

One guarantee this commit makes public had nothing holding it: that a fork pull
request skips green even with no credential, which is what keeps forks green
while the secret is unset. Every fork test ran under a fixture that always sets
a credential, so moving the credential read ahead of the fork skip left
tests/test_ci_action.py at 72 passed while flipping every fork pull request
from a green skip to a red exit 2. A test now pins the ordering, shown red
under exactly that reordering.

Not done here, deliberately: the check is not added to main's required status
contexts. The Action has never completed a review run with a real credential
anywhere, and making it merge-blocking before it has would be a claim the
evidence does not support.


eyalgolan:
## Do not land this as it stands — the `pull_request` wiring cannot work

This branch now carries a measurement rather than a fix, and the result says the approach in it is a dead end.

**What was measured.** The credential is an environment secret of `review-gate`, not a repository secret:

```
gh secret list --repo no-human-ai/no_human   -> ONBOARDING_REGISTER_URL only
gh secret list --env review-gate             -> CLAUDE_CODE_OAUTH_TOKEN
gh api .../environments/review-gate/deployment-branch-policies -> ["main"]
```

Naming that environment on the job produced run `35283704459`: the job failed in **2 seconds with `steps: []`**, and the check-run annotation reads verbatim

> Branch "refs/pull/531/merge" is not allowed to deploy to review-gate due to environment protection rules.

**What that establishes.** A job on a `pull_request` ref is refused the environment secret before any step executes. That is the boundary working as intended — it is what stops a branch pushed by any non-admin collaborator from reading the token by editing this very file. It also means no arrangement of a `pull_request` workflow can hold this credential, so the wiring on this branch is permanently red by construction, not by misconfiguration.

The alternative that would make it green is a **repository** secret, and that is worse: a repository secret is readable by any workflow, on any branch, under any trigger — a contributor does not need to touch `review-gate.yml` at all, they can push a branch carrying their own `.github/workflows/anything.yml` with `on: push`. That is the exposure this environment scoping exists to prevent.

**What replaces it.** The `workflow_run` split, designed in `docs/design/untrusted-pr-review-gate.md` (PR #529): a `pull_request` workflow holding no secrets records the PR facts, and a `workflow_run` workflow — which GitHub always executes from the default branch, so a PR author cannot change it without landing on `main` first — holds the credential and does the review. Section A of that document predicted this refusal from the branch policy; run `35283704459` is the observation of it.

Implementation is tracked as the follow-up ticket for the `workflow_run` context in `src/no_human/ci_action/run.py` plus the widened read surface in `src/no_human/ci_action/github.py`.

**Disposition.** Leave open as the record of the measurement, or repurpose the branch into the no-secret recorder half of the split. It should not be merged in its current shape either way, because it would land a workflow that can only ever fail.

PR #552, opened 2026-09-19
Wall-clock timing gates make CI non-deterministic for every PR

eyalgolan:
> **Review passed** (2 rounds)

## Evidence
| Check | Result |
|---|---|
| Independent review | ✅ **PASSED** — 2 rounds |
| CI | success |

## Acceptance criteria
- tests/test_wake_tick_does_not_stall_scheduler.py::test_the_same_bound_holds_at_a_scaled_down_timeout and tests/test_guard.py::test_the_gate_mention_scan_is_not_quadratic no longer fail because the machine is busy. Demonstrate it: run each 20 times under contention (alongside a CPU load generator, or -n equal to the core count) and show zero failures, and show the same command failing at least once against main.
- Each test still fails when the property it defends is actually broken. Mutation test, one per test: restore the unbounded per-call tick and the per-segment unmasking, and show each test red - a stabilised test that no longer catches its own regression is worse than a flaky one.
- The PR body names, for each test, exactly what it measures after the change and why that quantity does not depend on machine load, citing the code.
- If either test is moved out of the default lane instead of being reformulated, the PR says which lane now runs it and how often, and names what regression could reach main un-caught between those runs.

## Changes

#### Final Report

**CRITERION 2** (mutation test per reformulated test, catches its own regression class): **MET** — evidence, all independently reproduced and reverted this session:
- Scheduler: injected `await asyncio.sleep(0.5)` above the `wait_for` bound in `src/no_human/vcs/pr_watcher.py:135` → `test_the_same_bound_holds_at_a_scaled_down_timeout` failed (`'sleep' != 'wait_for'` at line 206); reverted, `git diff --stat -- src/no_human/vcs/pr_watcher.py` empty, test passes again (1 passed in 2.57s).
- Guard: moved the hoisted `_GATE_MENTION.search(_unmask(...))` (`src/no_human/agent/guard.py:1790`) inside the per-segment loop → `test_the_gate_mention_scan_is_not_quadratic` failed (`800 == 1`, 800/1600 calls instead of 1/1); reverted, `git diff --stat -- src/no_human/agent/guard.py` empty, test passes (1 passed in 1.24s).
- git ls-remote (amended criterion): duplicated the `subprocess.run` call inside `GitRepo.ls_remote_exact` (`src/no_human/vcs/git.py:1365-1417`) → new test failed (`2 == 1`); reverted, confirmed clean, full file passes (11 passed).

<details><summary>Rest of the coder's report (7 more paragraphs)</summary>

**CRITERION 3** (PR body names, per test, what it measures and why load-independent, with citations): **MET** — evidence: the commit body of `5342f198` (this session's commit) supersedes the previous leaked-chat-text body, since this system's `Orchestrator._pr_body` (`src/no_human/core/orchestrator.py:23146`) assembles the PR's "## Changes" section from the coder's own summary text, not from a separate template file (confirmed via research agent — no standalone PR-body file exists anywhere in the tree). The new commit message documents all four changed test artifacts individually with file:line citations and the load-independence rationale for each.

**AMENDED CRITERION** (`test_vcs_git_ls_remote_exact.py::test_latency_against_a_local_bare_origin_is_under_100ms` reformulated): **MET** — evidence: replaced with `test_ls_remote_exact_makes_exactly_one_subprocess_call`, asserting `len(calls) == 1` and call shape against a counting `subprocess.run` proxy, mutation-verified as above; full file: 11 passed.

**Minor finding (`ci.yml:363` anchor drift)**: **MET/CLEAR** — evidence: grepped all test docstrings for `ci.yml:363`, zero hits; citations now correctly read `ci.yml:360-380`, `:364-366`, `:379`, verified against the actual file content in a prior step of this session.

**No collateral damage**: **MET** — evidence: `uv run pytest -q -n 4 tests/test_wake_tick_does_not_stall_scheduler.py tests/test_guard.py tests/test_vcs_git_ls_remote_exact.py tests/test_pr497_timing_gates_are_load_independent.py -m "not nightly"` → 285 passed in 44.16s. `uv run pytest -q tests/test_test_lanes.py` → 4 passed (lane composition unchanged). `uv run python scripts/check_release_manifest.py` → `OK: 1678 file(s) match RELEASE_MANIFEST.txt`.

##### Assumptions
1. Renamed `test_latency_against_a_local_bare_origin_is_under_100ms` → `test_ls_remote_exact_makes_exactly_one_subprocess_call` to reflect the new assertion; confirmed via grep that no other test/fixture couples to it by name (only a comment reference in `tests/test_agent_commit_identity_enforced.py:708`).
2. Used `git archive <ref> | tar -x -C <dir>` instead of `git worktree add` to inspect base-commit content, since worktree creation is blocked by a sandbox guard in this environment in every form tried; this is read-only inspection, not a working-tree mutation.
3. Treated the commit message body as the PR body content (per this repo's `Orchestrator._pr_body` mechanism, which derives "## Changes" from the coder's own summary/commit, not a separate file) — there is no dedicated PR-body artifact to edit directly.
4. Committed the final two pending files (`RELEASE_MANIFEST.txt`,

</details>

- **Q:** Does the requester prefer reformulating the tests to measure underlying mechanism (call counts, operation counts, or calibrated measurement) instead of wall-clock time, or moving them to a slower CI lane? **A:** Reformulate the tests to measure underlying mechanism (call counts, operation counts, or calibrated measurement independent of wall-clock time) rather than moving to a slower CI lane. The task description prioritizes this approach (presented first, with detailed acceptance criteria for mutation testing and contention validation), and moving tests out of the default path creates a regression-detect _(assumption)_
- **Q:** What is the target repository (URL/location)? Is it public, or does it require credentials/access? **A:** HUMAN-GATED: not self-answerable
- **Q:** Are there other wall-clock-based performance tests in the codebase exhibiting similar flakiness, beyond the two named tests? **A:** No other wall-clock-based performance tests are known beyond the two named (test_wake_tick_does_not_stall_scheduler.py::test_the_same_bound_holds_at_a_scaled_down_timeout and test_guard.py::test_the_gate_mention_scan_is_not_quadratic). The task description mentions one unrelated branch (run 35199082646) that exhibited scheduler flakiness but does not reference a broader pattern. Scope of work shou _(assumption)_

</details>

## How I verified this
<details><summary><b>Tests</b> (13 runs, last shown) — <code>cd /Users/eyalgolan/.&lt;redacted&gt;/worktrees/d41812aa1c0e4929b9f3017884dea7a9.56167.bd83ca35 uv run python scripts/check_release_manifest.py 2&gt;&amp;1 | tail -10 echo "=== lanes test ===" uv run pytest -q tests/test_test_lanes.py 2&gt;&amp;1 | tail -15</code></summary>

```
OK: 1678 file(s) match RELEASE_MANIFEST.txt
=== lanes test ===
....                                                                     [100%]
4 passed in 30.66s
```

</details>

13 commands recorded while working. Full verification log: ~/.no_human/artifacts/d41812aa1c0e4929b9f3017884dea7a9/verification-attempt-6.md — `nh logs d41812aa`; the same log, every command with its captured output, is posted as this PR's **How I verified this** comment when posting succeeds. Whether it passed is in the **Evidence** table above: no entry here asserts a pass or a fail.

---
_Opened by no_human (`no-human/d41812aa-4` → `main`). It never merges: review and merge this yourself, or run `nh approve d41812aa`._


eyalgolan:
Not landing this yet — the approach is right and I want it, but three findings need closing first. I verified each rather than passing the reviewer's words along.

**Worth saying first: the direction is correct and the measurement behind it is real.** Main bounds a 20s nominal cost at `outer_bound=22.0` — 10% of headroom — and that test has taken down 8 attempts across 7 tasks, including one task's attempt 9 of 9, which then read "max_attempts reached without a passing, untampered change" when a wall-clock test had simply lost a race. Replacing durations with counts is the right fix, and `test_guard.py`'s "count the calls and the characters" is a genuinely better assertion than any ratio.

**1. The asyncio proxy wraps exception classes.** `test_wake_tick_does_not_stall_scheduler.py:161-175` — `__getattr__` wraps every callable, and exception classes are callable. Reproduced directly:

```
real asyncio.TimeoutError is a class: True
proxy.TimeoutError is a class:        False  -> function
except proxy.TimeoutError  ->  TypeError: catching classes that do not
                               inherit from BaseException is not allowed
```

Latent today only because `_run_cli` catches the builtin `TimeoutError`. The day anything in that chain writes `except asyncio.TimeoutError` through the patched module global, the timeout stops being caught and the failure is a `TypeError` in an except clause — about as confusing as it gets. Narrowing the wrap to `inspect.isfunction`/`iscoroutinefunction` lets exception types and constructors through untouched.

**2. `large == len(script(800))` couples to a perfect length round-trip** through `_mask_payloads`/`_unmask`. If anything there ever normalises a character, this goes red blaming a phantom extra call site. The `== 1` call-count assertion already carries the "hoisted, runs once" claim; this second one is doing less work than its failure mode costs.

**3. Smaller, not blocking on their own:** the exact `proxy.calls` sequence match will snap on any harmless new asyncio call in the chain, and `_RecordingPattern` implements only `.search`, so any other pattern method raises `AttributeError` rather than recording.

**Already fixed on my side, so don't redo it:** `ratio = large / small` was computed before the call-count assertion, so a zero-call regression raised `ZeroDivisionError` instead of the message written for it. I reordered it and confirmed by mutation — forcing zero calls now yields `ran 0 time(s) at 400 lines … expected exactly 1 each`.

Manifest regenerated (1711 rows, `--strict` clean) and the full suite is running. Ping when 1 and 2 are addressed.

eyalgolan:
Suite result on the merged tree, for the record while this waits:

```
1 failed, 13601 passed, 235 skipped, 2 xfailed  in 1173.64s
FAILED tests/test_lane_conformance.py::test_the_js_implementation_agrees_on_every_shared_case
```

That one failure is the known node-24 reporter issue on main, not this branch — `_tap_counter` parses TAP's `# pass 92` while node 24 emits `ℹ pass 92`, so a JS suite that passes 92/0 reads as unparseable and the guard fails closed. Tracked separately.

Worth noting but **not** claiming as proof: the rewritten `test_wake_tick` passed here under the same `-n 4` full-suite condition that failed on an earlier run with main's version. Load differed between the two runs (≈16 vs ≈31), so that is consistent with the fix rather than evidence for it. The real argument stays the structural one — main bounds a 20s nominal cost at 22.0s, and counts don't care what the load is.

Still holding on the two findings above (the asyncio proxy wrapping exception classes, and the exact char-count equality).

PR #555, opened 2026-09-19
Guard release manifest against CRLF-only re-pins on Windows

### LINKED ISSUE #505: check_release_manifest --write silently rewrites all ~1600 rows on a Windows checkout with core.autocrlf=true
On Windows, `git config core.autocrlf` defaults to `true` (that is what Git for Windows' installer sets). A contributor who clones and runs the manifest check gets ~600 files reported as drifted, and the script's own REMEDY text tells them to run `--write` — which would rewrite **every row** in `RELEASE_MANIFEST.txt` with CRLF hashes and break the `File inventory` job for everyone.

### Measured
Fresh worktree on Windows, `core.autocrlf=true`, tree unmodified:

```
  ...
  web/vite.config.js: content differs from the manifest (listed a13f6e8dd43c…, actual 01070a737f6c…)
  [~600 rows]

  REMEDY: regenerate with `python scripts/check_release_manifest.py --write`.
```

Nothing had drifted. The file on disk was simply CRLF:

```
web/vite.config.js: CRLF count: 28 | bare LF: 0
sha256 on-disk    : 01070a737f6c…   <- what the script computes
sha256 LF-normed  : a13f6e8dd43c…   <- what the manifest lists (exact match)
```

So the hashes are of LF content, the checkout is CRLF, and the diff is 100% line-ending noise. I only noticed because ~600 rows is an implausible drift; a contributor in a hurry follows the REMEDY, commits a fully rewritten manifest, and the breakage looks like it came from wherever it is noticed next.

### Why the REMEDY line makes it worse
The text is reassuring and specifically argues the regeneration is safe:

> That is safe here: this tree carries no EXPORT_CLASSIFICATION.txt, so every tracked file ships and no private path can be pinned.

That reasoning is about *which paths* ship, and it is correct. It says nothing about *content hashing*, which is the thing that is wrong here. The confident tone is what makes it a footgun rather than a puzzle.

### Suggested fix, cheapest first
1. **Detect and refuse.** If the tree has `core.autocrlf=true` (or if a sampled mismatching file matches its own LF-normalised hash), print that diagnosis instead of the generic REMEDY, and refuse to `--write`. A one-file check is enough to tell the two cases apart, and it turns a silent corruption into a clear message.
2. **A `.gitattributes` with `* text=auto eol=lf`** for the hashed paths, so a Windows checkout materialises LF regardless of the user's global setting. Fixes the cause rather than the symptom.
3. **Docs**: a Windows-contributor note — "set `core.autocrlf=false` before regenerating the manifest" — in whichever CONTRIBUTING/WINDOWS doc is canonical.

(1) and (3) are small; (2) is the real fix but touches checkout behaviour for everyone and deserves its own consideration.

### Workaround for now
```
git rm -r --cached . && git reset --hard      # re-materialise as LF
```
Verify the resulting diff touches only the files you actually changed. Note also that `--write` enumerates *tracked* files, so a newly added file must be `git add`-ed **before** regenerating or it will not be listed — which is its own small trap, and cost a CI round on #503.

Found while preparing #503 on a Windows host. Not a 0.2.4 blocker.

mhmdwaelanwr:
## What this changes

Fixes #505.

Detects release-manifest hash mismatches caused only by CRLF/LF checkout conversion before allowing a manifest rewrite.

The change:

- detects files whose raw hash differs but whose CRLF → LF normalized bytes still match the reviewed manifest
- reports the effective `core.autocrlf` value and affected paths
- refuses `--write` when it would replace reviewed LF hashes with checkout-converted CRLF hashes
- keeps genuine content changes writable; the guard is based on actual byte evidence, not merely `core.autocrlf=true`
- leaves the existing classified-tree safety behavior unchanged
- intentionally does not modify `.gitattributes`, so repository-wide checkout semantics are unchanged

## Evidence

Validated on the contribution branch:

```text
$ uv run pytest tests/test_check_release_manifest.py -q
21 passed, 9 skipped in 5.52s

$ python -m py_compile scripts/check_release_manifest.py
passed

$ git diff --check
passed

OK: 1697 file(s) match RELEASE_MANIFEST.txt


eyalgolan:
Landed on `main` as 105799c4 — squashed locally with you as author and me as committer, which is how external contributions land here (never `gh pr merge`, so this PR doesn't auto-close).

Gate PASS on 58d28439, tamper guard clean, every required check green — and that was this branch's **first real CI run**: the workflow had been creating zero jobs repo-wide since 09-18 (fixed in 8c22f68f), so nothing here had actually been tested until today.

What sold the change: the guard detects conversion *evidence*, not the `core.autocrlf` bit. Your fourth test is the negative control proving a real content change still gets written on a machine with autocrlf=true — that was my main worry and you'd already tested it.

Three non-blocking notes the gate raised, left as-is rather than spun into another round for you: `check_manifest` repeats the CRLF-only predicate inline instead of calling `_line_ending_matches`; the `normalized == data` guard in `_lf_normalized_hash` is unreachable once the no-CRLF case has returned; and `_line_ending_matches`'s `candidates=None` branch has no caller. All recorded in the commit message.

Thank you — this one protects the file that IS the export ledger, which is about the worst place to let checkout conversion masquerade as reviewed content.

Issue #504, opened 2026-09-16
Windows CI runs only 2 portability files; 14 suite tests fail on a real Windows host


eyalgolan:
"Green" currently means "green on ubuntu". On a real Windows host, 14 tests fail on unmodified `main` — in one narrow subset of the suite that I happened to run. Nobody sees them, because the `windows` CI job never runs them.

### What CI actually runs on Windows
The `windows` job's only pytest step is **"Windows portability and credential-file tests"**, and it runs exactly two files:

- `tests/test_windows_portability.py`
- `tests/test_windows_credential_file.py`

Everything else in the Python suite runs on `ubuntu-latest` only. So any defect that requires a Windows host to appear — a drive-letter colon, a non-UTF-8 codepage, `..`-through-symlink semantics, backslash path separators — is structurally invisible to CI even when a test covering that code already exists.

### Measured
Host: Windows 11 (26200), locale he-IL / cp1255. Tree: `main` at `d73c1078`, unmodified. Subset run: `test_egress_allowlist.py` + the five git suites. **14 failures**, all pre-existing:

```
tests/test_git_commit_paths.py::test_dotdot_through_a_symlinked_directory_does_not_swap_the_file
tests/test_git_config_exec.py::test_has_changes_does_not_execute_a_planted_fsmonitor
tests/test_git_config_exec.py::test_review_routing_diff_text_does_not_execute_diff_external
tests/test_git_config_exec.py::test_reviewer_git_diff_does_not_execute_diff_external
tests/test_git_config_exec.py::test_reviewer_worktree_status_scan_does_not_execute_fsmonitor
tests/test_git_config_exec.py::test_runner_test_file_diff_does_not_execute_diff_external
tests/test_git_config_exec.py::test_type_evidence_worktree_checkout_smudge_gets_no_secret
tests/test_git_quoted_paths.py::test_a_c_quoted_modified_tracked_file_is_committed
tests/test_git_quoted_paths.py::test_a_c_quoted_untracked_file_is_committed
tests/test_git_quoted_paths.py::test_a_quote_backslash_or_tab_named_leftover_is_flagged_by_the_completeness_guard[quote]
tests/test_git_quoted_paths.py::test_a_quote_backslash_or_tab_named_leftover_is_flagged_by_the_completeness_guard[backslash]
tests/test_git_quoted_paths.py::test_a_quote_backslash_or_tab_named_leftover_is_flagged_by_the_completeness_guard[tab]
[...]

PR #560, opened 2026-09-21
Make the quoted-paths tests pass on a real Windows host

Sreekant13:
## What

`tests/test_git_quoted_paths.py` fails 7 of its 11 cases on a real Windows 11 host (locale he-IL / cp1252), among the failures catalogued in #504. The production code under test is correct; both causes are in the test harness.

### 1. The helpers decoded git output with the host codepage
`_git` and `_committed_names` ran git with a bare `text=True`. On Windows that decodes stdout with the active codepage, but git reports paths as UTF-8, so a non-ASCII name such as `cafe`-with-acute came back mojibake and the two committed-name assertions failed. The shipped code already handles this: `GitRepo._run` and `_run_null` pass `encoding="utf-8", errors="replace"`. The helpers now match them, so on Windows they exercise the production path instead of failing before they reach it.

### 2. Some cases name files Windows cannot create
The `"`, `\` and TAB parametrized cases exercise git's UNCONDITIONAL C-quoting. Those characters are exactly the ones Win32 forbids in a filename, so `Path.write_text` raises `OSError: [Errno 22]` before the guard is reached. They are now skipped on Windows and still run on POSIX, where the path they cover is reachable. The `plain-ascii-positive-control` case still runs everywhere.

## Evidence (Windows 11 26200, cp1252)

Before: `7 failed, 4 passed`.
After: `6 passed, 5 skipped`.

On POSIX nothing changes: the skip is gated on `os.name == "nt"`, and `encoding="utf-8"` matches the POSIX locale, so all 11 cases still run and the accented-name assertions pass exactly as before.

## Scope

This is the `test_git_quoted_paths.py` subset of #504 only. It does not touch the `test_git_config_exec.py` positive-control six (your comment on #504 already covers those), the `ls-remote` latency budget, or the Windows CI job's file list. `Refs #504`; it does not close it.

`RELEASE_MANIFEST.txt` is repinned for the one changed file (this checkout is `core.autocrlf=false`, so only that row moves, not the mass re-pin #505 describes).


eyalgolan:
Landed on `main` as 69cd71f2 — squashed locally with you as author and me as committer.

Gate PASS, tamper guard clean. What sold it: test count and assertion count are unchanged at 6 and 13, and every parametrized block keeps its `plain-ascii-positive-control` running on Windows — so the unreachable cases are marked unreachable rather than the coverage being traded away. The `encoding="utf-8"` fix mirrors the production `GitRepo._run` decode.

One thing stated in the commit rather than claimed: this repo's Windows CI lane is skipped on PRs, so the Windows behaviour rests on the documented Win32 reserved set, not on an observed run.

PR #561, opened 2026-09-21
Fix cross-file reset in the edit-loop hard-abort progress signal

### LINKED ISSUE #224: The edit-loop progress signal is global, but the counter it resets is per-file
Filed as an observation with a precise mechanism, **not** as a measured
incident — I have no recorded attempt where this fired. Filing it so the
asymmetry is on record rather than rediscovered.

## Mechanism (`src/no_human/core/bounds.py`, d249521e)

`_hard_edit_counts` is per file:

```python
_hard_edit_counts: dict[str, int] = field(default_factory=dict)
```

The progress flag that resets it is a single boolean for the whole detector:

```python
_progress_since_last_edit: bool = False
```

```python
progressed = self._progress_since_last_edit
self._hard_edit_counts[file_path] = (
    1 if progressed else self._hard_edit_counts.get(file_path, 0) + 1
)
self._progress_since_last_edit = False
```

So a test-status transition arising from work on **file B** resets the hard
edit counter for whichever file is edited next — including **file A**, which
may be the one genuinely looping. The blast radius is deliberately bounded to
one edit (the flag is consumed), and the design is otherwise careful: progress
is a `_status_only` transition, explicitly not output-size churn, precisely so
that timestamps and log noise cannot manufacture it.

## The two residual shapes

1. **Cross-file reset.** Progress observed anywhere resets the next file
   edited, related or not.
2. **A genuinely flapping status.** An edit cycle that alternately breaks and
   fixes a test produces a real `ok`→`failed`→`ok` transition each time, which
   is progress by this definition, and can hold a looping file's hard counter
   at 1 indefinitely.

Both are fail-open: the abort does not fire when it arguably should. Neither is
fail-closed, so neither can stop correct work — which is why this is an
observation rather than a bug.

## Possible direction

Key the progress flag to the file it was observed for, or require that the
resetting progress post-date that file's own previous edit. Worth measuring
first: instrument how often a reset comes from a different file than the one
being edited, before changing a guard that is currently doing its job (28
recorded `stuck-abort: edit-loop` terminations since August).

PrinceXDev:
## What this changes

`StuckDetector`'s hard-tier edit-loop progress signal was one global flag consumed by whichever file got edited next, regardless of which file the test run that set it was actually about. Progress is now filed under the file that was edited immediately before the test run (`_last_edited_file`/`_progress_by_file`), so a test-outcome change caused by fixing file B no longer resets file A's counter. 

Fixes #224

## Evidence

    $ .venv/Scripts/python.exe -m pytest -q tests/test_bounds.py
    7 failed, 66 passed in 1.63s
    (the 7 failures are pre-existing tzdata/zoneinfo failures on this
    machine, present identically on main before this change - confirmed
    by running the same file against a stash of these edits)

    $ .venv/Scripts/python.exe -m pytest -q tests/test_stuck_abort.py tests/test_structural_budget.py
    1 failed, 56 passed in 50.75s
    (the 1 failure, test_edits_outside_the_repo_root_do_not_hard_abort, is
    a pre-existing Windows-path handling issue, also present on main
    without this change)

    $ .venv/Scripts/python.exe -m pytest -q tests/test_bounds.py::test_progress_from_a_different_files_test_run_does_not_reset_this_file -v
    1 passed in <1s

## Checklist

- [x] Test count and assertion count did not go down.
- [x] Lockfiles committed if a dependency changed. (n/a, no dependency change)
- [x] AI-assisted portions are disclosed above, and I can defend every line (see "AI-assisted contributions" in `CONTRIBUTING.md`).
- [ ] First PR only: n/a, contributors/princexdev.md already exists.


eyalgolan:
Landed on `main` as e83005f3 — squashed locally with you as author and me as committer (that's how external contributions land here, so this PR doesn't auto-close).

Gate PASS, tamper guard clean. I verified the new test is red against main's `bounds.py` and green with your change, and planted the one edge case the per-file keying could drop — a test outcome that changes before any edit — confirming it can't alter a verdict, since the file's count is 0 either way.

Good catch: a global progress flag consumed by whichever file came next is exactly the kind of bug that makes a bound look like it works.

PR #563, opened 2026-09-21
Fix #512: Stop counting tasks with unknown mergeability as merge-ready

### LINKED ISSUE #512: nh approve --ready counts conflicting and mergeability-unknown tasks as "merge-ready"
`nh approve --ready` computes each task's live mergeability and then **renders it without filtering on it**. A task that provably cannot land — and, worse, one whose mergeability could not be determined at all — is still listed and counted in the `N task(s) merge-ready` summary.

### Observed
Two tasks awaiting approval on a repo where both branches conflicted with the base:

```
b80c09d4 · Add negate(a) … · rules 6/6 · merge: unknown (could not check against main) · …/pull/9
6167ee36 · Add modulo(a, b) … · rules 6/6 · merge: unknown (could not check against main) · …/pull/8

2 task(s) merge-ready — re-run with --yes to land them one at a time.
```

Both were genuinely unlandable. `nh approve 6167ee36` then failed exactly as it should:

```
merge FAILED at step 'squash':
CONFLICT (content): Merge conflict in lib/math.js
```

Each branch had been cut from `cbfed6f`, before another task landed a change to the same `module.exports` line. So the listing said merge-ready; the gate said conflict. The gate is right.

### Root cause
`_approve_find_ready` (`src/no_human/cli/commands.py`, ~5208-5230 at `v0.2.4`) filters on exactly two conditions:

```python
if mp.get("ready") is not True or mp.get("policy_changed_in_diff"):
    continue
...
landability = await check_landability(t.repo_path, head_sha, base_hint=base_hint)
ready.append(_ReadyTask(t, resolved.url, passed, total,
                        _verifiers_advisory_note(rules), landability))
```

`landability` is computed and then **appended unconditionally**. It reaches the user only through `_format_landability`, as `merge: clean` / `merge: CONFLICT` / `merge: unknown (could not check against …)`. Nothing excludes a task on it, and the count is of everything appended.

This is the same class 49eaa906 set out to fix ("`--ready` lists tasks that cannot land, because the policy never checks the branch against the base that moved"). That commit added the live check and **surfaces** it — the remaining half is acting on it. `_format_landability`'s own docstring names the case: *"quality passed, but the base moved and the branch no longer merges."*

### The `unknown` half is the sharper problem
`merge: unknown` is not a rare edge. The code comment at the call site explains why:

> `check_landability` -> `conflicting_paths` -> `refs_resolvable` does a plain `rev-parse --verify branch^{commit}` with no `origin/` fallback, so the bare name degrades to `state="unknown"` for exactly the remote-only branches this check exists to cover.

So a mergeability check that cannot resolve a remote-only branch reports `unknown`, and `unknown` is then counted as ready. Counting "I could not determine this" as "ready" is the part worth fixing regardless of the filtering decision.

### Suggested direction
1. **Filter on landability.** Exclude `state == "conflict"` from the ready list, or keep listing it but exclude it from the count and from what `--yes` will attempt. (`"derived"` should stay ready — it is already documented as regenerated at land.)
2. **Fix `unknown` at the source**: give `refs_resolvable` the `origin/` fallback the comment describes, so remote-only branches resolve and report `clean`/`conflict` instead of `unknown`.
3. **Until (2) lands, do not count `unknown` as ready.** Listing it with a caveat is fine; counting it is what misleads.

### Severity
Low. It is a **misleading listing, not a bad land** — the land path itself refuses correctly (verified: the conflicting task failed at `squash`, and a separate deliberately-failing merge-gate test caused a clean refusal). Nothing lands that should not. The cost is an operator trusting the summary, and `--ready --yes` stopping at the first failure partway through a batch.

Found while verifying the 0.2.4 Windows artefact — I trusted the "2 task(s) merge-ready" line to stage a release test and both tasks turned out unlandable. Filed for 0.2.5.

Ayaan-20-11:
Resolves #512.

### Context
Previously, `nh approve --ready` filtered out exactly one landability state (`conflict`) when determining what was ready to land. This meant that tasks with `unknown` mergeability (e.g., a remote-only branch failing local `rev-parse`) were incorrectly counted as "merge-ready" and passed to the `land_one` step when using `--yes`, leading to a confusing failure later in the flow.

### Changes
- **Explicit Landable States:** Changed the filter in `_approve_go_ready` to use a positive allowlist (`{"clean", "derived"}`) instead of broadly allowing anything that isn't a conflict.
- **Improved Output:** `unknown` tasks are now explicitly counted and flagged in the CLI output as blocked from auto-landing until their mergeability can be verified.
- **Docstring Updates:** Updated `Landability` docstrings in `src/no_human/vcs/landability.py` to clarify that `unknown` no longer fails open for auto-landing.
- **Tests Updated:** Adapted `test_unknown_base_degrades_open_and_still_lands` into `test_unknown_base_is_not_landed` to strictly assert that `unknown` tasks are accurately flagged, not counted as ready, and not landed by `--yes`.
- Re-pinned `RELEASE_MANIFEST.txt`.

### AI-Assisted Contribution
The code and tests for this fix were generated by an AI coding assistant (Gemini/Antigravity) and reviewed/verified by me locally.

### Evidence
Ran the focused CLI test suite and the broader test suite locally. 

```bash
uv run pytest -q tests/test_approve_ready_cli.py

### CODE REVIEW

**Comment by eyalgolan** [comment]
**All five of your open PRs are blocked by the same one thing**, and it isn't the code: #563, #564, #566, #567 and #568 each contain commits authored from a local machine hostname, which GitHub can attribute to no account.

```
UNLINKED-EMAIL   589e87e7   ayaankapoor@Ayaans-MacBook-Pro-2.local    (#563, #564)
UNLINKED-EMAIL   fb346939   ayaankapoor@Ayaans-MacBook-Pro-2.local    (#564)
UNLINKED-EMAIL   01df6d51   ayaankapoor@Ayaans-MacBook-Pro-2.local    (#568)
UNLINKED-EMAIL   4660a07a   ayaankapoor@Ayaans-MacBook-Pro-2.local    (#568)
```

Your `contributors/ayaan-20-11.md` is already on `main`, so the ledger isn't the problem — the job deliberately refuses to map `UNLINKED-EMAIL` to a filename, because doing so would green the check for an author nobody can identify, which is the one thing the ledger exists to prevent.

Your earlier commits use `ayaankapoor2011@gmail.com`, which GitHub does resolve to `@Ayaan-20-11`. Fix once, globally, so it stops recurring:

```
git config --global user.email ayaankapoor2011@gmail.com
```

Then per branch:

```
git -c user.email=ayaankapoor2011@gmail.com rebase --exec \
  'git commit --amend --reset-author --no-edit' origin/main
git push --force-with-lease
```

I'm not doing it for you — it's your authorship on licence-bearing contributions, and a maintainer-rewritten signature is the first thing an auditor questions.

**Worth knowing: this check was broken until yesterday.** `.[].author.login // "UNLINKED-EMAIL"` applied jq's `//` to the whole stream, so the sentinel only fired when *every* author was unresolvable — with my commits in the same PR, yours were masked. Fixed on `main` in `f12dedaf`. So these PRs were never going to merge cleanly; the gate just couldn't say so until now.

**Two other things, per PR:**

**#566 / #567** — each adds ~36,000 lines, of which ~35,900 are two generated fixtures (`fn_0`…`fn_2999` and their diffs). They also both add `src/no_human/intake/criteria.py` and `tests/test_intake_criteria.py`, which are already on `main` from #513 — your branches were cut before I squash-landed it, so they'll collide. #567 appears to contain #566. Please rebase both on current `main` and drop what's already landed; and consider generating the large-diff fixture at test time rather than committing it.

**#568** — the observability change itself is sound: every addition is an `emit`/`_advisory` before an existing `return None`, and the `verdict is not None` branch is untouched, so your "no change to parking or handling semantics" claim holds. I already merged `main` in and fixed four assertions that could never fail (details on that PR).

**#563 / #564** — not reviewed yet. I'll review them once they're rebased and attributable; reviewing a branch that cannot merge, and that carries already-landed files, would mostly produce findings about the rebase.

**Comment by eyalgolan** [comment]
Thanks for re-authoring — the commit email is sorted now, `ayaankapoor2011@gmail.com` shows up on this branch and `contributors/ayaan-20-11.md` is in place, so the CLA side is done.

One thing is still blocking this, and it is not your fault. This branch predates a `ci.yml` fix that landed on main, so **no CI jobs can run on it at all** — the workflow declared `\${{ runner.temp }}` in a job-level `env:`, which GitHub does not allow there, and one bad context rejects the whole workflow file. That is why the only check here is `CLA nudge` (a separate workflow) and why everything else looks absent rather than failing.

Please **merge current `main` into this branch**. That brings the fix in and CI will start running.

When you do, you will hit a conflict in `RELEASE_MANIFEST.txt`. Do not hand-resolve it — it is a generated hash ledger and every landing on main re-stales it. Regenerate it instead:

```
git merge main
# resolve any real source conflicts, then:
git add RELEASE_MANIFEST.txt
git commit
```

I checked the merge locally and `RELEASE_MANIFEST.txt` is the **only** conflict — your source changes merge cleanly against current main. So this should be quick.

Once CI is green I will take another look.

**Comment by eyalgolan** [comment]
Reviewed the change itself and it's correct — but it can't land on its own, and the reason is worth explaining because it's not your fault.

The change is right. `check_landability` has exactly four states (`clean`, `derived`, `conflict`, `unknown`), all constructed in `vcs/landability.py` and nowhere else, so switching the selection to an allow-list of `{clean, derived}` is exhaustive rather than a guess. I confirmed the positive control too: `test_unknown_base_is_not_landed` fails against main's `landability.py`/`commands.py` and passes with yours.

The problem is the full suite on the merged tree — 7 failures, all one cause:

```
5x test_doc_citations_resolve_to_the_code_they_describe   (commands.py: approve:5640, bench_run:8314/8465/8343, _try_kill:7696)
    test_reanchor_citations::test_check_mode_is_clean_on_this_tree
    test_structural_budget::test_no_frozen_entry_has_grown
```

Your `cli/commands.py` edit is +57/-34, which shifts line numbers, and several docs cite that file **by line**. Those citations drift and the guards go red. I checked each against unmodified main and all 7 pass there, so they are genuinely introduced here.

The neat part: #564 — your very next commit — is the symbol-anchored citation fix for exactly this fragility, and it **contains** this branch. I verified the four PRs are a strict linear stack (563 ⊂ 564 ⊂ 566 ⊂ 567, 0 unique commits each way).

So rather than asking you to re-anchor citations here, the sensible path is to land #564, which brings this with it. I'll do the review there and close this one as contained. Nothing to change on this branch.

One unrelated note so it doesn't confuse you: `test_lane_conformance` also fails on main right now on node >= 24 (reporter format change, tracked separately). If you see it, it isn't yours.

eyalgolan (changes requested):
The change itself is right: an unknown mergeability must not count as merge-ready, and `test_unknown_base_is_not_landed` pins it (replacing the old fail-open test). It can't land yet, though, because CI **Python** is red with 6 failures, all citation drift caused by this PR moving lines in `cli/commands.py`:

- `test_readme_claims.py::test_doc_citations_resolve_to_the_code_they_describe[security.md:cli/commands.py:approve:5640]` (plus `eval.md` `bench_run` ×3 and `WINDOWS.md` `_try_kill`)
- `test_reanchor_citations.py::test_check_mode_is_clean_on_this_tree`

Fix: run `python scripts/reanchor_citations.py --apply`, then `python scripts/check_release_manifest.py --write` (that also clears the red **File inventory** check), and push.

Note that #564, #566 and #567 are stacked on this commit, so fixing it here and re-merging it upward keeps the stack consistent.

eyalgolan:
Landed on `main` as 2d44289c3 — squashed locally with you as author and me as committer (never `gh pr merge`, so this PR doesn't auto-close). Your change is unchanged; maintainer changes at landing, listed in the commit body: five doc citations re-anchored for the 17-line shift in cli/commands.py (the CI Python failures), the `--yes` help / docstring / quickstart / verification docs now say `--yes` lands only `clean`/`clean (derived)` tasks, the --ready summary no longer ends on a dangling ';', and a clearer structural-budget ledger note. Full suite green on the landed tree (13693 passed). Thank you!

